PR #12, opened 2019-07-17
11 refactor routing

### LINKED ISSUE #11: refactor authentication routing
includes making the landing page go to login or straight to app depending on user authentication level

DanielSeehausen:
this pr is regrettably really large. I would have liked to do it in smaller chunks, but one refactor led to another! so it goes

worked on with @vdiezel 

I will write comments in line!


src/App.jsx
```diff
@@ -0,0 +1,29 @@
+import React from  "react"
```
DanielSeehausen: this component has one role:

provide a middle layer for "are we loading or not"

this enabled `Routes` to be functional and single purpose!

src/components/devicecontrol/DeviceControl.js
```diff
@@ -15,7 +15,6 @@ import {callLoadDevice} from "../../redux/thunks/thunks";
 
 const mapStateToProps = state => {
     return {
-        authenticated: state.authenticated,
```
DanielSeehausen: unused
authentication has been delegated to our routing, instead of at the component level

src/components/devicecontrol/DeviceControl.js
```diff
@@ -102,8 +101,6 @@ class DeviceControl extends React.Component {
 
         const { initializing } = this.state;
 
-        if (!this.props.authenticated) return <Redirect to="/login"/>;
```
DanielSeehausen: same as above

src/components/devicecontrol/manageaccess/ManageAccess.js
```diff
@@ -84,4 +83,4 @@ class ManageAccess extends React.Component {
     }
 }
 
-export default connect(mapStateToProps, mapDispatchToProps)(ManageAccess);
```
DanielSeehausen: this is an auto editor change because git + others generally wants all files to end with a newline

src/components/devicecontrol/setup/DeviceSetup.js
```diff
@@ -36,7 +35,7 @@ const mapDispatchToProps = dispatch => {
 };
 
 const DeviceSetup = props =>  {
-
```
DanielSeehausen: again auto editor change removing unused whitespace. You can take out viewing whitespace changes in the diff view options

src/components/login/Login.js
```diff
@@ -116,5 +116,5 @@ class Login extends React.Component {
         )
     }
 }
-export default connect(mapStateToProps, mapDispatchToProps)(withRouter(Login));
```
DanielSeehausen: this was nefarious and absolutely rocked my world for longer than I would like to admit :(
`withRouter` on the outside!

src/components/main/Main.jsx
```diff
@@ -49,20 +56,9 @@ class Main extends React.Component {
     }
 
     getCurrentTabFromUrl(){
```
DanielSeehausen: this was slimmed way down because, going forward, as our url paths grow (and we put more url parameters in them), we will still want to route to the correct tab if it exists instead of deferring to one of them (in this case, `devices`). Additionally, if someone puts in an aberrant url directly, we may want to be blowing up with a 404 instead of silently deciding intended behavior. This gives us a chance to fix something quickly if its our issue, and is explicit to the user when they are mistaken on some url manipulation (which I doubt our clients will be doing much). 

TL;DR: If a url is screwed up, its most likely our fault, not a user's. We should blow up so we can fix it.

src/components/main/Main.jsx
```diff
@@ -120,9 +117,25 @@ class Main extends React.Component {
                     deleteEphemeralNotificationsHandler={this.props.callDeleteEphemeralNotifications}
                     ephemeralNotifications={getEphemeralNotifications(this.props.user)}
                 />
-            </Sidebar>
+              </Sidebar>
```
DanielSeehausen: routing moved to the sidebar!
will prolly abstract a sidebar component out of this soon, but bigger 🐟 to fry for us atm!

src/components/routing/PrivateRoute.jsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
```
DanielSeehausen: this is some #dank stuff @vdiezel and I put in together. Works well :)

src/components/routing/Routes.js
```diff
@@ -1,78 +1,22 @@
 import React from  "react";
-import { HashRouter, Route } from "react-router-dom";
+import { BrowserRouter, Route } from "react-router-dom";
```
DanielSeehausen: this gets the `#` out of our url!

src/components/routing/Routes.js
```diff
@@ -1,78 +1,22 @@
...
-                </div>
-                }
-        </HashRouter>)
-    }
-}
+import PrivateRoute from "./PrivateRoute";
 
-export default connect(mapStateToProps, mapDispatchToProps)(Routes);
 
+const Routes = () => (
```
DanielSeehausen: functional components are beautiful components! If you can get away with no state in a component, definitely go for it.

src/index.js
```diff
@@ -1,13 +1,14 @@
...
 import { Provider } from "react-redux";
-import Routes from './components/routing/Routes';
+
+import './styles/semanticoverrides.css';
+import App from './App';
 
 ReactDOM.render(
   <Provider store={store}>
-    <Routes/>
+    <App />
```
DanielSeehausen: top level entry point, later, we can use this component to determine login screen vs. signup (if no browser cookies preent) vs. straight to the sidebar routed stuff!

src/redux/initialState.js
```diff
@@ -2,7 +2,7 @@ export const initialState = {
     maxTimeBetweenHeartbeats: 3000,
     user: { notifications: [] },
     activeMenuTab: "devices",
-    authenticated: false,
+    authenticated: undefined,
```
DanielSeehausen: this gives us 3 states for authentication:
`undefined` === not yet resolved (so we know when the loading screen is present
`false` === resolved: not authed
`true` === resolved: authed

src/redux/reducers/reducer.js
```diff
@@ -13,11 +13,12 @@ import {
 
 function rootReducer(state = initialState, action) {
     if (action.type === ACTIONS.SET_SESSION){
-        const user = createUserObject(action.payload.userData);
+        const userData = action.payload.userData
+        const user = userData && createUserObject(action.payload.userData);
```
DanielSeehausen: to fix a bug where `createUserObject` was being run on `null`

petros751:
It seems that the roots have improved a lot. I love them

PR #13, opened 2019-07-17
1 refac restful

vdiezel:
Changed the thunks to fit the new sdk calls and added missing parameters in the components when they are called.


src/components/devicecontrol/manageaccess/ManageAccess.js
```diff
@@ -18,9 +18,9 @@ const mapStateToProps = state => {
...
 const mapDispatchToProps = dispatch => {
     return {
-        callAddDeviceUser: (deviceId, email, groupName, hostName) => dispatch(callAddDeviceUser(deviceId, email, groupName, hostName)),
+        callAddDeviceUser: (deviceId, email, groupName, hostName) => dispatch(callGrantDeviceAccess(deviceId, email, groupName, hostName)),
         callRemoveDeviceUser: (deviceId, userId, groupName, hostName) =>
-            dispatch(callRemoveDeviceUser(deviceId, userId, groupName, hostName)),
+            dispatch(callRevokeDeviceAccess(deviceId, userId, groupName, hostName)),
         callModifyDeviceAccess: (deviceId, userId, newGroup, oldGroup) =>
             dispatch(callModifyDeviceAccess(deviceId, userId, newGroup, oldGroup))
     };
```
vdiezel: renamed "addUser" and "removeUser" to grantAccess and revokeAccess

src/components/devicecontrol/manageaccess/UserTable.js
```diff
@@ -43,7 +43,7 @@ class UserTable extends React.Component {
                             user={user}
                             deviceAccess={deviceAccess}
                             deviceId={activeDevice.id}
-                            itsMe={me.email == user.email}
+                            itsMe={me.email === user.email}
                             me={me}
                             isOwner = {user.id === activeDevice.owner}
```
DanielSeehausen: what do you think about making the `activeDevice.owner` -> `activeDevice.ownerId`?
if its a lot of work then not worth it, of course, but that makes it more explicit we are dealing with an id, instead of a user object, which i'd assume `owner` is

src/components/devicecontrol/manageaccess/User.js
```diff
@@ -35,6 +35,8 @@ class User extends React.Component {
 
         const accessGroup  = this.props.access.find(group => group.users.includes(this.props.user.id));
 
+        debugger;
```
DanielSeehausen: this guy snuck in
image

src/components/devicecontrol/processes/Processes.js
```diff
@@ -11,18 +11,18 @@ import { connect } from "react-redux";
...
-    callRemoveProcessUser, callUpdateProcess
+    callRevokeProcessAccess, callUpdateProcess
 } from "../../../redux/thunks/thunks";
 
 const mapDispatchToProps = dispatch => {
     return {
         callGetProcesses: (limit, skip, deviceId) => dispatch(callGetProcesses(limit, skip, deviceId)),
-        callAddProcessUser: (processId, email, groupName) => dispatch(callAddProcessUser(processId, email, groupName)),
-        callRemoveProcessUser: (processId, userId, groupName) => dispatch(callRemoveProcessUser(processId, userId, groupName)),
+        callAddProcessUser: (processId, email, groupName) => dispatch(callGrantProcessAccess(processId, email, groupName)),
```
DanielSeehausen: have a leftover `callAddProcessUser` here.
I don't think its breaking, but on the off chance you wanted to change all refs to it and it just snuck by...
and it is then used by the same name way down at the end of the render method. (line 282)

src/components/devicecontrol/processes/Processes.js
```diff
@@ -11,18 +11,18 @@ import { connect } from "react-redux";
...
+        callRemoveProcessUser: (processId, userId, groupName) => dispatch(callRevokeProcessAccess(processId, userId, groupName)),
```
DanielSeehausen: also here with `callRemoveProcessUser`
and down on line 283

src/components/main/Main.jsx
```diff
@@ -115,6 +115,7 @@ class Main extends React.Component {
                     Logout
                 </Menu.Item>
                 <ConciergeModal
+                    user={this.props.user}
                     show={this.state.showConciergeModal}
                     closeHandler={this.toggleModal.bind(this, "showConciergeModal")}
                     deleteEphemeralNotificationsHandler={this.props.callDeleteEphemeralNotifications}
```
DanielSeehausen: name swaps here `deleteEphemeralNotificationsHandler` => `callDeleteEphemeralNotificationsHandler`

DanielSeehausen: name swaps here `deleteEphemeralNotificationsHandler` => `callDeleteEphemeralNotificationsHandler`

DanielSeehausen: here as well with a rename: `callDeleteEphemeralNotificationsHandler` => `deleteEphemeralNotificationsHandler`

we may be passing down a function as a prop here that we pulled out of redux (`callDeleteEphemeralNotificationsHandler`), to the `ConciergeModal`.

would we be able to grab this function directly out of redux in the `ConciergeModal` and avoid passing it down from `Main` instead? I believe it is considered an anti-pattern, and we lose some of redux's glory when we don't take advantage of the global state nature of it.

src/components/devicecontrol/manageaccess/ManageAccess.js
```diff
@@ -18,9 +18,9 @@ const mapStateToProps = state => {
 
 const mapDispatchToProps = dispatch => {
     return {
-        callAddDeviceUser: (deviceId, email, groupName, hostName) => dispatch(callAddDeviceUser(deviceId, email, groupName, hostName)),
+        callAddDeviceUser: (deviceId, email, groupName, hostName) => dispatch(callGrantDeviceAccess(deviceId, email, groupName, hostName)),
         callRemoveDeviceUser: (deviceId, userId, groupName, hostName) =>
```
DanielSeehausen: name change didn't carry over here from `callRemoveDeviceUser` -> `callRevokeDeviceAccess`

src/components/devicecontrol/manageaccess/UserTable.js
vdiezel: I think that's a good idea! I changed it. This affects hambda and REACTor.

src/components/devicecontrol/manageaccess/User.js
vdiezel: EMBERASSING! Should run a global search before commiting - could be a good idea for a script

src/components/main/Main.jsx
vdiezel: yes, but thats a general pattern, whenever I passed down a "callXY" thunk down to a compoenent I dropped the call part.

src/components/devicecontrol/processes/Processes.js
vdiezel: fixed.

src/components/main/Main.jsx
vdiezel: Both is possible. I followed here the old pattern of "containers" but after reading https://github.com/elastic/kibana/issues/14547 I tend to be more on the side of connecting child components on demand - but this means refactoring most of the "container" components we have which would be 2-3 hours of work I guess. I'd like to postpone this a little bit.

src/components/devicecontrol/manageaccess/ManageAccess.js
vdiezel: fixed

PR #16, opened 2019-07-18
Fix proptype error

DanielSeehausen:
fixes a big ugly error that was throwing:

image

- bunches together the authorization components into `src/components/authorization/`
- renames components to match routes
- fixes duplicate usage of `Login` component name
- very minor code golfing - forgive me and the vim macros
- fixes bug introduced re: url parsing when the hashrouter was removed
- fixes redirect bug in `About`

note: there are actually very few changes here. All of 95% the import diff lines are auto semicolon removals


src/components/authentication/signUp/SignUp.js
```diff
@@ -174,5 +173,6 @@ class SignUp extends React.Component {
         )
     }
 }
-export default connect(null, mapDispatchToProps)(withRouter(SignUp));
+
+export default withRouter(connect(null, mapDispatchToProps)(SignUp))
```
DanielSeehausen: this `withRouter` addition/move is the majority of this PR!

src/components/about/About.js
```diff
@@ -1,90 +1,85 @@
-import React from 'react';
-import {Divider, Container, Segment, Header, Grid, List} from "semantic-ui-react";
-import { Redirect } from "react-router-dom";
+import React from 'react'
+import {Divider, Container, Segment, Header, Grid, List} from "semantic-ui-react"
+import { withRouter } from "react-router-dom"
 
-const About = props => {
-
-        if (!props.authenticated) return <Redirect to="/login"/>;
```
DanielSeehausen: this should have been removed in the last pr 😬

petros751:
All good

PR #20, opened 2019-07-18
WebPack config file

petros751:
I read the documentation and through that I created the webpack configuration file.
Probably we will use the **resolve.alias** instead the **resolve**


webpack.config.js
```diff
@@ -0,0 +1,6 @@
+var path = require('path');
```
DanielSeehausen: good rule of thumb to follow is: if it can be a `const`, make it a `const`! After that, preference `let` over `var`, because `var` can get a programmer into some nasty trouble!

[const](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const)
[let](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let)
[var](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/var)

DanielSeehausen: good rule of thumb to follow is: if it can be a `const`, make it a `const`! After that, preference `let` over `var`, because `var` can get a programmer into some nasty trouble!

[const](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const)
[let](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let)
[var](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/var)

webpack.config.js
```diff
@@ -0,0 +1,6 @@
...
+
+resolve: {
```
DanielSeehausen: this is looking like a good start, but I think you may find when you go to test it this won't work!

put yourself in the perspective of the computer:

"I go to execute webpack"
"I got to check the `webpack.config.js` file in the project root directory by default
"I execute this file and **import** the contents it exported so I can use them"

webpack.config.js
```diff
@@ -0,0 +1,6 @@
...
+    root: path.resolve('./src'),
+    extensions: ['', '.js']
+  }
```
DanielSeehausen: this red arrow here is telling us something important:

_The message is displayed because otherwise there is no way to tell the difference between a file where there is a newline at the end and one where is not. Diff has to output a newline anyway, or the result would be harder to read or process automatically._

_Note that it is a good style to always put the newline as a last character if it is allowed by the file format. Furthermore, for example, for C and C++ header files it is required by the language standard._
[for more information](https://stackoverflow.com/questions/5813311/no-newline-at-end-of-file)

DanielSeehausen:
heya @petros751 we have since removed the `dev` branch for a more streamlined deploy process -- would you be able to reset this to the `master` branch? Sorry for asking you to undo what we previously asked you to do haha :(. All part of the process of refining our deploy strategy, I suppose...

PR #33, opened 2019-07-22
added deployment step to s3 to circleCI

### LINKED ISSUE #4: implement CD GitHub
An important step towards turning our process into a machine!

- [x] discuss deploy process with the team
- [ ] implement CD pipeline for deploy to production when master is merged in to

vdiezel:
added a deployment step to our circleCI workflow. In case of a merge into master, it will run the following steps after build:

- installs the AWS CLI 
- ships the content to the s3 bucket (--delete removes files that are not being updated which gets rid of the cached files in the build)
- invalidate the content in the CDN (otherwise we need up to 1 hour until updated content is beeing served)

closes #4


vdiezel:
I don't think the filters apply to the build step at all, every job is indicated by a dash in the workflow

PR #52, opened 2019-07-24
added subscribing to the process topics when reloading browser

vdiezel:
fixed subscribing and refactored the subscribing workflow for process:

We subscribe in 2 places now:

1. The normal way  of subscribing happens in the process viewer: When the ProcessViewer is loaded, we determine in which state the current process is and subscribe accordingly. 
2. For the ProcessBuilder we now subscribe just to phases which tell us, when the process has finally started. Afterwards the process will be passed down to the ProcessViewer and we subscribe to the topics we want. We are overwriting the phases-subscription because we need another callback!


src/components/device/processes/Processes.jsx
```diff
@@ -13,9 +13,11 @@ import {
 import {
   callGrantProcessAccess, callCloneProcess, callCreateProcess,
   callDeleteProcesses,
-  callGetProcesses, callLoadProcessViewer,
+  callGetProcesses,
+  callSubscribeToProcessTopics,
```
vdiezel: some renaming to be more on point here

src/components/device/processes/Processes.jsx
```diff
@@ -97,8 +98,6 @@ class Processes extends React.Component {
 
     if (process.state === "executable"){
       this.props.loadProcessBuilder(process)
-    } else if(process.state === "running" || process.state === "finished"){
```
vdiezel: delegated this to the ProcessViewers componentDidMount

src/components/processbuilder/ProcessBuilder.jsx
```diff
@@ -17,7 +17,7 @@ import {
 } from "../../redux/actions/actions"
 import { getPhasesArray, getPhasesKeys, getProcessDownStreams, amendDescriptionMismatch } from "../../utils/process"
 import { connect } from "react-redux"
-import { callLoadProcessViewer, callStartProcess, callUpdateProcess } from "../../redux/thunks/thunks"
+import { callSubscribeToProcessStart, callStartProcess, callUpdateProcess } from "../../redux/thunks/thunks"
```
vdiezel: New thunk to subscribe to a topic after starting a process

PR #55, opened 2019-07-25
removed stored state from the process, instead derive it from the pro…

### LINKED ISSUE #26: Pause/Resume Feature
As a user, I can pause and resume my process and receive meaningful visual feedback regarding what things actually paused and what did not.

- [ ] determine exactly which processes actually 'pause', and which continue running
- [ ] clarify what the payload contract with firmware is to have an action like this work successfully
- [ ] write the front-end required for a user to see a button, click a button, receive visual feedback that things are indeed paused, and be able to unpause. Take into consideration that the user may try to rapidly click pause/unpause, and any latency between the web app and the device may cause unexpected behavior (ideally, receive confirmation from the device that it paused before giving the user the option to unpause).

vdiezel:
…gression array and the device state. Finalized all missing pieces for the resume and pause feature.

This commit turned out bigger than expected as I discovered a flaw in the architecture: The process state is not necessary at all, as it can be derived from the progression of a process and the device state of the device it is running on. All device states that are related to a process, currently only "pause" and "running", also save the processId of the currently running process. 

This allowed me to simplify the subscribing process.

Unfortunately it turned out on firmside to be a bit more effort to finalize the pause-and-resume feature. However I tested the client and backend side manually and it works. Therefore I consider this ticket closed. When we finally do E2E-tests, we will do bugfixes if something pops up on our side.

closes #26


src/components/device/processes/ProcessScreen.jsx
```diff
@@ -4,20 +4,18 @@ import { connect } from "react-redux"
...
 
-  switch (activeProcess.state) {
-    case "executable":
+  switch (true) {
+    case (activeProcess.state === "executable"):
       return <ProcessBuilder activeProcess={activeProcess} activeDevice={activeDevice}/>
-    case "running":
-      return <ProcessViewer activeProcess={activeProcess} activeDevice={activeDevice}/>
-    case "finished":
+    case (PROCESSVIEWER_STATES.includes(activeProcess.state)):
```
vdiezel: avoids 3 repetitions of the ProcessVIewer call

src/components/device/processes/ProcessTile.jsx
```diff
@@ -35,6 +35,11 @@ const ProcessTile = props => {
...
       statusColor = "status-blue"
       iconType = "circle"
+    } else if (state === "paused") {
+      image = processRunning
+      displayedStatus = "paused"
+      statusColor = "status-orange"
+      iconType = "circle"
     } else if (state === "executable") {
       image = processExecutable
       displayedStatus = "executable"
```
vdiezel: new display props for process and device Widgets

src/components/processbuilder/ProcessBuilder.jsx
```diff
@@ -132,7 +132,7 @@ class ProcessBuilder extends React.Component {
               activeDevice={this.state.activeDevice}
               setProcessBuilderValidationErrorsHandler={this.props.setProcessBuilderValidationErrors}
               startProcessHandler={this.props.callStartProcess}
-              subscribeToProcessStartHandler={this.props.callSubscribeToProcessStart}
+              subscribeToProcessStartHandler={this.props.callPrefirePhaseSubscription}
```
vdiezel: a better description

src/components/processviewer/ProcessViewer.jsx
```diff
@@ -30,9 +31,10 @@ const mapDispatchToProps = dispatch => {
       dispatch(callSetStreamCalibration(deviceId, streamId, initialize, slope, offset)),
     callNextProcessPhase: (processId, deviceId) => dispatch(callNextProcessPhase(processId, deviceId)),
     callStopProcess: (processId, deviceId) => dispatch(callStopProcess(processId, deviceId)),
-    callPauseProcess: (processId, deviceId) => dispatch(callPauseProcess(processId, deviceId)),
+    callPauseProcess: (processId, deviceId, downStreamIds) => dispatch(callPauseProcess(processId, deviceId, downStreamIds)),
```
vdiezel: downstream which should be stopped on pause are passed down the device together with the command

src/redux/actions/actions.js
```diff
@@ -26,6 +26,7 @@ export const ACTIONS = {
   MODIFY_PROCESS_ACCESS: "MODIFY_PROCESS_ACCESS",
   UPDATE_PROCESS: "UPDATE_PROCESS",
   CREATE_PROCESS: "CREATE_PROCESS",
+  FINISH_PROCESS: "FINISH_PROCESS",
```
vdiezel: added a method to set running processes to finish when a device goe into the idle state - because we removed the publishing a "fake" last phase with phaseId -1 to indicate a process end. When now simply check, if there was a process running when a device went back to idle

PR #88, opened 2019-08-14
added tests for calibration, renamed phaseKey to phaseId for consistency

### LINKED ISSUE #85: Calibration - divide by zero issue in graphical mode

vdiezel:
fixes #85 

 - Added tests for Calibraitonmodal
-  renamed phaseKey to phaseId for consistency


vdiezel:
I'll

src/components/processviewer/PhasesData.jsx
```diff
@@ -27,7 +27,6 @@ class PhasesData extends React.Component {
             <Streams
               processUpStreams={this.props.processUpStreams}
               activeProcess={this.props.activeProcess}
-              activeProcessData={this.props.activeProcessData}
```
DanielSeehausen: 🕵 🔥

src/components/processviewer/modals/CalibrationModal.jsx
```diff
@@ -324,7 +325,10 @@ class CalibrationModal extends React.Component {
       const {offset} = this.state
```
DanielSeehausen: not a blocking request, but there is a lot of DRYing up and function abstraction opportunities in this method, which is pretty hard to read meaningfully at its current length.

src/redux/reducers/reducer.js
```diff
@@ -135,9 +135,9 @@ function rootReducer(state = initialState, action) {
 
     const updatedProcesses = processes.map(process => {
       if (process.id === action.payload.processId) {
-        for (const currentPhaseKey of Object.keys(process.description)){
-          if (action.payload.phaseId === currentPhaseKey) {
-            process.description[currentPhaseKey].duration = action.payload.duration
+        for (const currentPhaseId of Object.keys(process.description)){
```
DanielSeehausen: same here re: DRY - we are repeating this same "find phase by id" lookup several places in the reducer. Not something worth changing now - just noting for the time being.

src/utils/process.js
```diff
@@ -19,18 +19,18 @@ export function getPhasesArray(process){
...
+// returns the phase ids in the right order
 
-export function getPhasesKeys(process){
-  return getPhasesSequence(process).map(phase => phase.key)
+export function getPhaseIds(process){
+  return getPhasesSequence(process).map(phase => phase.id)
 }
 
-// returns the phases as [[key1, phaseDataObject1],...] in the right order
+// returns the phases as [[id1, phaseDataObject1],...] in the right order
```
DanielSeehausen: is this comment block meant to stay in?

test/components/CalibrationModal.test.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react'
```
DanielSeehausen: this new test file is actually 🔥 

the tests (going from what we looked at on your computer locally that is not yet up to date here) are short, super readable, and provide really great user experience assertions without being so granular they are not worth our time to write. 

awesome!

src/redux/reducers/reducer.js
vdiezel: I'm also uneccessarily looping though keys and do a comparison instead of simply fetching the phase by the phaseId from the process.description

src/components/processviewer/modals/CalibrationModal.jsx
vdiezel: the calibration needs a refac for sure, it's probably the most complicated code in the app so far

vdiezel:
Added more test coverage - however I couldn't test certain behaviour because it would require either clicking on the graph (which is probably a tough job to simulate) OR setting the state of component manually which is an anti pattern in RTL as we assert behaviour out of the user perspecttive

Removed a bit of repetition from the reducer

PR #120, opened 2019-08-19
WIP: Merge builder into builder

### LINKED ISSUE #107: as a user I should be able to hide controller streams in the process config in process viewer

### LINKED ISSUE #118: Integrate the process builder into the process viewer

vdiezel:
This one is almost done:

- Moved the process builder into the process viewer
- on the fly connected most child components to the store instead of passing down state and handlers down as props
- a bit of code reformatting

re: Tests -> testing this one is a huge effort, because it requires a lot of state mocking, but I'll give it a shot before I merge this one in

closes #118 
closes #107


src/components/device/manageaccess/User.jsx
```diff
@@ -1,20 +1,32 @@
...
+    errorMessage: '',
   }
 
   handleOpen = () => this.setState({isOpen: true})
 
   handleClose = () => this.setState({isOpen: false})
 
   deleteAndClose = async (userId, groupName) => {
-    await this.props.revokeDeviceAccessHandler(this.props.deviceId, userId, groupName, this.props.me.userName)
+    await this.props.callRevokeDeviceAccess(this.props.deviceId, userId, groupName, this.props.me.userName)
```
DanielSeehausen: in the future, we may have more than just users who have access to devices
i.e. groups/organizations

so at some point we might want to think about incorporating those into our access naming schemes
i.e. `callRevokeUserDeviceAccess`, (nothing blocking now, just something to think of)

src/components/device/manageaccess/UserTable.jsx
```diff
@@ -1,17 +1,30 @@
...
+const mapDispatchToProps = dispatch => ({
+  callModifyDeviceAccess: (deviceId, userId, newGroup, oldGroup) =>
+    dispatch(callModifyDeviceAccess(deviceId, userId, newGroup, oldGroup))
+})
 
 class UserTable extends React.Component {
 
   async modifyAccess(userId, newGroup) {
 
     const oldGroup = this.props.activeDevice.access.find(group => group.users.includes(userId)).name
```
DanielSeehausen: we will be writing methods like this look up again (we can see lookups similar to this sprinkled all over our code base now).

while this isn't something to change for this PR, it may be worth our time to provide methods on the models to abstract these things whenever we see them. this will save us big time in the long run. When/if we want to change these lookup methods.(if this app grows as intended, that will be a lot in the future, and this is especially true for our permissioning stuff). 

i.e.:

```javascript
const oldGroup = this.props.activeDevice.access.find(group => group.users.includes(userId)).name

// becomes this if we are actually using instances
const oldGroupName = activeDevice.getUsersAccessGroup(user).name

// or this if we are staying functional
const oldGroupName = Device.getUsersAccessGroup(device.id, user.userId).name
```

src/components/device/processes/ProcessScreen.jsx
```diff
@@ -2,24 +2,16 @@ import React from "react"
...
 
-  switch (true) {
-    case (activeProcess.state === "executable"):
-      return <ProcessBuilder activeProcess={activeProcess} activeDevice={activeDevice}/>
-    case (PROCESSVIEWER_STATES.includes(activeProcess.state)):
-      return <ProcessViewer activeProcess={activeProcess} activeDevice={activeDevice}/>
-    default:
-      return null
-  }
+  return <ProcessViewer activeProcess={activeProcess} activeDevice={activeDevice}/>
```
DanielSeehausen: ++ ✝️ refactor easy -6

src/components/processes/Processes.jsx
```diff
@@ -52,21 +52,21 @@ class Processes extends React.Component {
...
+    if (event.key === 'Enter') {
       const {queryString, filterProperty} = this.state
 
       const filtered = []
 
-      if (queryString.trim() !== "") {
+      if (queryString.trim() !== '') {
 
-        const re = new RegExp(queryString, "i")
+        const re = new RegExp(queryString, 'i')
```
DanielSeehausen: may be an opportunity to simplify this with `filter`, while keeping our nesting low:

```javascript
filterProcesses(event) {
  const { queryString, filterProperty } = this.state
  const query = queryString.trim()

  if (event.key !== 'Enter' || query === '') return

  const processIds = this.props.processes.filter(process => process.name.includes(query))

  this.setState({filteredProcessIds: processIds}, () => {
    this.updateActivePage(1)
  })
}
```

src/components/processviewer/body/builder/PhaseTabContent.jsx
```diff
@@ -0,0 +1,156 @@
...
+                  onClick={this.editName.bind(this)}
+                  floated="right"
+
+                >
+                  <Icon name='edit'/>
+                </Button>}
+              </Form>
+              <Divider/>
+              <Accordion fluid styled>
+                {Object.entries(groups).map((group, i) => {
```
DanielSeehausen: iteration can be tough to follow in render statements, which are often best kept as short as possible (the render statements, that is)

not worth doing now, but worthwhile habit is abstracting all work out of render statements in the future.

I am not 100% on the group object structure, but If I am understanding it, we may be able to use Object.entries a little more cleanly:
```javascript
Object.entries(groups).map(( [groupName, groupStreams], idx) => {
  // groupName and groupStreams should now be in scope without unpacking them from `group`
}
```

src/components/processviewer/body/dataviewer/DataGraphs.jsx
```diff
@@ -0,0 +1,65 @@
...
+import { Container, Grid, Segment } from 'semantic-ui-react'
+import{ getGroups } from 'utils/process'
+
+class DataGraphs extends React.Component {
+
+  state = {
+    activeIndices: []
+  }
+
+  handleClick = (e, titleProps) => {
```
DanielSeehausen: this toggle be duplicated [here](https://github.com/ospin-web-dev/REACTor/pull/120/files#diff-634ce616fb3ed096717c9f6040453c23R51)

maybe worth making a toggle for ourselves:
```javascript
function addOrRemove(array, value) {
    var index = array.indexOf(value);

    if (index === -1) {
        array.push(value);
    } else {
        array.splice(index, 1);
    }
}
```

src/components/processes/Processes.jsx
vdiezel: Thanks! We definetely want to simplify this - but I propose in a slightly different way:
  
    1. I'd like to use a regex here because I want to check case insensitive. Because of that I also have to check if the queryString is different from an emty string otherwise the regex doesn't work
    2. Instead of reduce I'd like to use filter and map, because this makes more clear what is actually going on

src/components/processviewer/body/builder/PhaseTabContent.jsx
vdiezel: Agree!

src/components/processviewer/body/dataviewer/DataGraphs.jsx
vdiezel: I'll leave this one untouched for now because we may want to changed the behaviour of the application here: This takes care of opening up the right entries in the accordion - as Jan wanted to open all entries across all phases we will probably refactor this anyway --> I'd like to spare the duplicate refactoring. But I'll keep that solution in my mind!

src/components/device/manageaccess/User.jsx
vdiezel: Yes, that's a good point there. We also might write the function in a way that it can take any kind of "identity" (Group OR user) and apply the access rights modififcation

PR #133, opened 2019-08-28
WIP: Implement process description snapshots

vdiezel:
A big one including 
important things commented

- creating process snapshots when a process is opened
- displaying diffs when a running process is changed
- commiting the changes to a running process, possibly applying a fancy phase insertion when a stream target in a running phase has been changed
- a couple of tests

Last step missing: dispatching a device command to the backend


src/components/authentication/Login.jsx
```diff
@@ -97,7 +97,7 @@ class Login extends React.Component {
                     <div
                       className="link-style"
                       onClick={this.goToSignup.bind(this)}
-                    >No accout yet? Sign up!</div>
+                    >No account yet? Sign up!</div>
```
vdiezel: fixed type

src/components/device/modals/CreateProcessModal.jsx
```diff
@@ -66,7 +67,7 @@ class CreateProcessModal extends React.Component {
     if (res.success){
       const newProcess = this.props.deviceProcesses[0]
 
-      this.props.loadProcessBuilder(newProcess)
+      this.props.loadProcessViewer(newProcess)
```
vdiezel: processBuilder doesn't exist independently anymore

src/components/processviewer/ProcessViewer.jsx
```diff
@@ -127,7 +131,7 @@ class ProcessViewer extends React.Component {
             {processBuilderValidationErrors.length > 0 && (
               <Message error>
                 <Message.Header>
-                  Process could not start.
+                  Process could not be dispatched.
```
vdiezel: renamed so it also fits when commiting changes to a running process fails the validation

src/components/processviewer/body/builder/ChangesFooter.jsx
```diff
@@ -0,0 +1,67 @@
...
+const mapStateToProps = state => ({
+  processDescriptionSnapshots: state.processDescriptionSnapshots
+})
+
+const mapDispatchToProps = dispatch => ({
+    callUpdateProcess: (processId, params) => dispatch(callUpdateProcess(processId, params)),
+    setProcessBuilderValidationErrors: errors => dispatch(setProcessBuilderValidationErrors({errors}))
+})
+
+async function commitChanges(props, diffs) {
```
vdiezel: 1. check if modifications pass validation (like stream target limits, "undefined" stream targets etc..)
2. apply modifications to the process (see comment at the function)
3. save the updated process (this also updates the snapshot)
4. if the changes need to be dispatch to the device (currently only phase renaming DOESN't need to be dispatched) we dispatch a request to update the process description on the device

src/components/processviewer/body/builder/ChangesTable.jsx
```diff
@@ -0,0 +1,66 @@
...
+  )
+}
+
+const renderUnknownChange = () => null
+
+const ROW_RENDER_DISPATCH = {
+  phaseCountChange: renderPhaseCountChangeRow,
+  phasePropChange: renderPhasePropChangeRow,
+  streamTargetChange: renderStreamTargetChangeRow,
+  unknownChange: renderUnknownChange
```
vdiezel: added `unkownChange` this to prevent crashes when unexpected changes happen

src/components/processviewer/body/builder/PhaseTabContent.jsx
```diff
@@ -2,8 +2,9 @@ import React from 'react'
 import { connect } from 'react-redux'
 import Group from './Group'
 import TransitionFooter from './TransitionFooter'
+import ChangesFooter  from './ChangesFooter'
 import { Divider, Accordion, Icon, Form, Header, Button } from "semantic-ui-react"
-import { getGroups } from "utils/process"
+import { getGroups, phaseIsModifiable } from "utils/process"
```
vdiezel: `phaseIsModifable` return false for finished phases, otherwise true and is used e.g. to disable UI elements

src/redux/helper/objectMapper.js
```diff
@@ -60,7 +60,12 @@ export const createDeviceObject = deviceData => ({
   calibration: deviceData.calibration,
   streamDefaults: deviceData.streamDefaults,
   streamNameAliases: deviceData.streamNameAliases,
-  readyForCommand: true
+  readyForCommand: true,
+  heartBeatSubscription: null,
```
vdiezel: this prevents double subscriptions per device for every topic

src/redux/middleware/deviceLogging.js
```diff
@@ -0,0 +1,68 @@
+import { ACTIONS } from '../actions/actions'
```
vdiezel: Ignore this file...shouldn't be here at all

src/redux/reducers/reducer.js
```diff
@@ -515,9 +540,9 @@ function rootReducer(state = initialState, action) {
     }
 
     // applying states for running processes derived from their devices (current options: "running" and "paused")
-    const runningProcesses = state.devices.map(device => (
-      {id: device.runningProcessId, state: device.state})
-    )
+    const runningProcesses = state.devices
```
vdiezel: we are only interested in running devices here

src/redux/thunks/thunks.js
```diff
@@ -156,69 +157,29 @@ export function callNextProcessPhase(processId, deviceId) {
   }
 }
 
-export function callPrefirePhaseSubscription(process, device){
```
vdiezel: We don't need the prefire subscription anymore

src/utils/PDVC.js
```diff
@@ -0,0 +1,162 @@
...
+
+  static isPhaseNameChange(diff) {
+    return diff.type === 'phasePropChange' && diff.path[diff.depth - 1]
+  }
+
+  static runtimeRelevantChanges(diffs){
+    return !diffs.every(this.isPhaseNameChange)
+  }
+
+  static applyRunningPhaseModifications(process, snapshot, insertPhase){
```
vdiezel: This function is called whenever we apply changes to a running phase. We pass in a flag `insertPhase` which indicates whether we have to insert a phase ourselves. Currently there is only one case when this is done: When a stream target is changed. Renaming a running phase is not of interest for the device (wont even be notified on this) and duration and transition changes do not required a phase insertion.

If we do have to insert a phase, we the following:
  1. add it to the process end 
  2. move it to right position (after the running phase)
  3. overwrite its properties with the props from the running phase
  4. overwrite the properties of the running phase with the saved props from the snapshot, however replace the duration with the elapsed time

we return the updated process. the elapsed time and the entry phase ID

src/utils/PDVC.js
```diff
@@ -0,0 +1,162 @@
...
+        name: runningPhaseSnapshotData.name,
+        duration: elapsedTime,
+        transition: runningPhaseSnapshotData.transition,
+      })
+    }
+
+    return { process: updatedProcess, elapsedTime, entryPhaseId: insertedPhaseId || runningPhaseId }
+  }
+
+  static applyDescriptionModification(process, snapshots,  diffs){
```
vdiezel: 1. Initialize default values
2. check if running phase was modified, if yes, check if we need to insert a phase `hasStreamTargetChange`, and update the process description 
3. if runningPhase was not modified, we still need to check if the running phase is a time based phase in order to send the right elapsed time to the device (for now `elapsedTime` for manual transition will be always 0, we can discuss this behaviour)
4. return to the component

src/utils/process.js
```diff
@@ -116,33 +131,30 @@ export function groupByControllerName(controllerStreams){
 }
 
 export function getStreamTarget(process, targetPhaseId, streamId){
-  for (const phaseId of Object.keys(process.description)){
-    if (targetPhaseId === phaseId) return process.description[phaseId].downstreams[streamId].value
-  }
+    return process.description[targetPhaseId].downstreams[streamId].value
```
vdiezel: *facepalm*

src/utils/timeutils.js
```diff
@@ -10,7 +10,7 @@ export function parseDuration(time) {
   duration.days = days
   duration.hours = hours
   duration.minutes = minutes
-  duration.seconds = seconds
+  duration.seconds = Math.round(seconds)
```
vdiezel: we are only interested in full seconds here

src/components/processviewer/body/builder/PhaseTabContent.jsx
```diff
@@ -75,6 +76,8 @@ class PhaseTabContent extends React.Component {
     const phaseId = this.props.phaseId
 
     const groups = getGroups(processDownStreams)
+    const phaseCanBeModified = phaseIsModifiable(this.props.activeProcess, phaseId)
+    const phaseNeverRan = phaseNeverStarted(this.props.activeProcess, phaseId)
```
vdiezel: Only phases which never started can be deleted

PR #149, opened 2019-09-03
Remove phase id reassignment

### LINKED ISSUE #130: Remove re-assignment of phase keys when a phase is deleted or moved to starting phase
Currently, phase keys get re-assigned for legacy reasons re: compatibility with the hardware. Since nina's work on the firmware, this has been updated. 

The starting phase key should no longer be re-assigned to 0. 

This can be considered part of the ad-hoc changes

@vdiezel knows whats going on here the best :))

vdiezel:
closes #130 , #134


src/components/device/modals/CreateProcessModal.jsx
```diff
@@ -47,7 +47,8 @@ class CreateProcessModal extends React.Component {
       deviceName: this.props.activeDevice.name,
       name: processName,
       comment: processComment,
-      description: {}
+      description: {},
+      entryPhaseId: 0
```
vdiezel: Adding a default entrryPhaseId when creating a process

src/components/processviewer/body/builder/ChangesFooter.jsx
```diff
@@ -30,14 +31,14 @@ async function commitChanges(props, diffs) {
...
+    description: updatedProcess.description,
+    entryPhaseId: parseInt(entryPhaseId)
   }
 
   await props.callUpdateProcess(updatedProcess.id, updateParams)
 
   if (PDVC.runtimeRelevantChanges(diffs)){
-    // dispatch update command
-    console.log(updateParams, elapsedTime, entryPhaseId)
+    await this.props.callStartProcess(activeProcess, elapsedTime, true)
```
vdiezel: resuing the startProcess command here

src/redux/helper/objectMapper.js
```diff
@@ -32,6 +32,7 @@ export const createProcessObject = processData =>  {
     createdBy: processData.createdBy,
     cloneHistory: processData.cloneHistory,
     description: processData.description,
+    entryPhaseId: processData.entryPhaseId || 0,
```
vdiezel: legacy support

src/components/processviewer/phasesdisplaybar/PhasesNavbarWrapper.jsx
```diff
@@ -81,7 +81,7 @@ class PhasesNavbarWrapper extends React.Component {
                     <PhaseTabDraggable
                       phase={phase}
                       key={2 * i + 1}
```
DanielSeehausen: just curious what the `key` calculation here is used for

src/redux/helper/objectMapper.js
DanielSeehausen: good thinking with the default there

src/components/processviewer/phasesdisplaybar/PhasesNavbarWrapper.jsx
vdiezel: Because I'm creating two elements which can not be nested in a parent (for drag-and-drop reasons) I hab to provide unique keys for the Divider and the PhaseTab so I split them up into even and uneven numbers!

DanielSeehausen: clever! i love it

PR #152, opened 2019-09-03
trash icon on process tile now showing only on hover

### LINKED ISSUE #144: device process tiles should only display their trash icon when an icon is being hovered over
image

Current behavior:
- process icons always display their trash icon

Desired behavior:
- [x] on hover, process icons have their trash icons quickly fade in
- [x] when mouse exits hover, process icons trash icons quickly fade out
- [x] by default, when not hovered, process icons do not display their trash icons

vdiezel:
closes #144


src/components/device/processes/ProcessTile.jsx
```diff
@@ -29,6 +29,8 @@ const ProcessTile = props => {
 
   function hoverDummy() {}
 
+  const [showTrashIcon, toggleTrashIcon ] = useState(false)
```
DanielSeehausen: love new react!

PR #156, opened 2019-09-05
replaced loading streams individually with a call to the new load pro…

### LINKED ISSUE #119: As a user, my process data in the process viewer should load quicker
Currently we are firing a GET request for very single strea, e.g. for 20 streams 20 GET requests are fired in the data viewer. We should have a new endpoint for the initial process data loading that loads all stream data in a single call!

vdiezel:
…cess summary function. 

Loads data much faster on intial process opening.

closes #119


src/components/device/setup/ConnectedStreamsTable.jsx
```diff
@@ -41,6 +41,9 @@ const CalibrationTable = props => {
           <Table.HeaderCell>
             Description
           </Table.HeaderCell>
+          <Table.HeaderCell>
+            ID
```
vdiezel: Added streamId because it's handy for debugging

src/components/processviewer/ProcessViewer.jsx
```diff
@@ -38,21 +39,30 @@ class ProcessViewer extends React.Component {
...
+    if (activeProcess.state === 'executable'){
+      const missMatchStreams = getMismatchingStreams(activeProcess, activeDevice)
 
       if (missMatchStreams.missingInDevice.length || missMatchStreams.missingInProcess.length) {
         this.setState({showDescriptionMismatchModal: true, missMatches: missMatchStreams})
-      }}
+      }
+    }
+    else {
+      this.setState({ loadinData: true })
```
vdiezel: A local state variable to indicate that data is loaded - will be used soon ( when we refactor the loading spinner to be more consistent)

src/redux/reducers/reducer.js
```diff
@@ -238,7 +238,7 @@ function rootReducer(state = initialState, action) {
     const processes = state.processes
 
     const updatedProcesses = processes.map(process => {
-      if (process.id === processId) {
+      if (process.id === processId && process.data.some(entry => entry.id === streamId)) {
```
vdiezel: firmware sometimes publishes data that is not part of a current upstream (it's a bug I believe) - to prevent crashes we double check here if the upstream really exists

src/components/device/setup/ConnectedStreamsTable.jsx
DanielSeehausen: HOLLA

src/components/processviewer/ProcessViewer.jsx
```diff
@@ -43,16 +44,22 @@ class ProcessViewer extends React.Component {
...
+
+    if (activeProcess.state === 'executable'){
+      const missMatchStreams = getMismatchingStreams(activeProcess, activeDevice)
 
       if (missMatchStreams.missingInDevice.length || missMatchStreams.missingInProcess.length) {
         this.setState({showDescriptionMismatchModal: true, missMatches: missMatchStreams})
-      }}
+      }
+    }
+    else {
```
DanielSeehausen: if there is a more exact state we could match on (similar to how we are asking `activeProcess.state === 'executable'` above), it would be good defensive programming practice to ask it in this `else` clause as well. i.e. if there is something like `activeProcess.state = 'complete'`

...but I don't know what exists, so thought I would ask

src/components/processviewer/body/dataviewer/Stream.jsx
```diff
@@ -16,368 +16,354 @@ import {
...
+     * basic idea:
+     *  determine how many datapoints are visible - if not enough,
+     *  load x datapoints in the given range,plus additional datapoints to the left and right for panning;
+     *  because panning triggers rangeChange-event multiple times, its limited on once every 2 seconds
+     *  Override 'reset'-event to move back to current selected view (otherwise no render is triggered)
+     */
 
-        if (event.trigger === 'pan'){
+    let reloadOnPan = false,
+      newPan = null
```
DanielSeehausen: `newPan` not used outside of the if block - can remove it and assign `const newPan = Date.now()` on line 128 below

src/redux/reducers/reducer.js
DanielSeehausen: is `entry` the same thing as `stream`?
aka is `data` a collection of streams?

src/redux/thunks/thunks.js
```diff
@@ -244,6 +244,23 @@ export function callDownloadStreamData(streamId, streamName, processId) {
...
+export function callLoadProcessDataSummary(processId) {
+  return async dispatch => {
+    try {
+      const res = await nexus.datapoints.getDataSummary({ processId })
+      res.data.forEach(streamData => {
+        dispatch(addHistoricStreamData({ streamId: streamData.streamId, data: streamData, processId }))
+      })
+      return {success: true, message: "OK."}
+    } catch (e) {
+      let error = e
```
DanielSeehausen: both `error` and `e` are used here, but they reference the same object. Maybe easiest to just get rid of `error`

src/redux/thunks/thunks.js
```diff
@@ -244,6 +244,23 @@ export function callDownloadStreamData(streamId, streamName, processId) {
...
+      if (e.response) error = e.response.data
+      console.log(error)
```
DanielSeehausen: if we are to keep logging in the code base like this (hopefully something we abstract later when we have time), could we use `console.error` instead of log?

src/redux/reducers/reducer.js
DanielSeehausen: 👍 nice defensive programming

vdiezel: data is collection of objects with a certain format that is easily fed into the Graph API

src/redux/thunks/thunks.js
vdiezel: I'll create a ticket for this one

PR #157, opened 2019-09-05
Hotfix click on running device

petros751:
I used the `.some` method to find the process that it has same `process.id` and `runnningProcessId`.
And then according the result from the `.some` I redirect the user in the pages.


src/components/devices/DeviceWidget.jsx
```diff
@@ -35,6 +35,7 @@ const Device = props => {
     image = bioreactorOfflineImage
   }
 
+
```
DanielSeehausen: whoops! make sure to remove changes like these (which aren't doing anything) from PRs to keep the history clean!
I know this seems minor, but we do it to make sure that when other devs are reading through the code history, or reviewing a PR, they are kept focused.

src/components/devices/Devices.jsx
```diff
@@ -25,7 +25,8 @@ class Devices extends React.Component {
 
   state = {
     initializing: true,
-    showCreateDeviceModal: false
+    showCreateDeviceModal: false,
+    processId: ''
```
DanielSeehausen: this value (`processId`), is never used, and can be removed.

src/components/devices/Devices.jsx
```diff
@@ -39,13 +40,14 @@ class Devices extends React.Component {
     //await this.props.callCreateDevice("test device", true, "bioreactor")
   }
 
-  handleDeviceClick(deviceId, deviceState, runningProcessId) {
+  handleDeviceClick(deviceId, runningProcessId) {
     this.props.resetActiveDevice()
-    if (deviceId) {
-      this.props.history.push(`/device/${deviceId}/processes`)
-    } else {
+    const processExists = this.props.processes.some(process => process.id === runningProcessId)
```
DanielSeehausen: This is looking good. 

What is this variable really storing though? I think what we are asking is whether or not a user has access to a specific process. 

We know a process exists when a `runningProcessId` is present.
What we are concerned about here is whether or not a user has access rights to a process.

Currently, we are looking in the `processes` in the store to understand whether or not a user has access to the process, but this is a roundabout way of asking our real question: "Does the current user have access to this specific process". 

We would really like to query the processes' access rights with a given user id, but that is a lot of work to change right now, so perhaps we just change the variable to imply exactly what we are doing:

`processExists` => `userHasReadAccessToProcess`

that way, when Ankita and Ali join us, they don't need to understand that all processes a user has read access to automatically populates in `this.props.processes`!

In the future, we want stuff like this to be abstracted to util/model functions, because we will be asking this question a lot in our codebase.

src/components/processviewer/ProcessViewer.jsx
```diff
@@ -44,6 +44,7 @@ class ProcessViewer extends React.Component {
   async componentDidMount(){
 
     this.props.loadProcessViewer(this.props.activeProcess)
+
```
DanielSeehausen: whoops! make sure to remove changes like these (which aren't doing anything) from PRs to keep the history clean!
I know this seems minor, but we do it to make sure that when other devs are reading through the code history, or reviewing a PR, they are kept focused.

src/components/devices/Devices.jsx
DanielSeehausen: (think @vdiezel added this too, but his comment died somewhere)

src/components/devices/Devices.jsx
DanielSeehausen: One more thing on this:

currently, we are passing up the `runningProcessId` from the child component to this callback. 
but in the child component, we are deriving `runningProcessId` from a `device` object that we pass down to the child, which is present in this parent component.

We can reduce the mental overhead here but just passing up the `deviceId`, then deriving the active process in the same place we use it (in this handle click method)

DanielSeehausen:
Now that is a clean PR!

PR #169, opened 2019-09-11
140 upgrade device display

DanielSeehausen:
will annotate with notes in the diff


package.json
```diff
@@ -28,7 +28,8 @@
     "babel-plugin-module-resolver": "^3.2.0",
     "jest-canvas-mock": "^2.1.0",
     "react-router": "^5.0.1",
-    "redux-mock-store": "^1.5.3"
+    "redux-mock-store": "^1.5.3",
+    "uuid": "^3.2.1"
```
DanielSeehausen: used for the device seed - react was complaining about duplicate keys when mapping these in render

src/components/device/DeviceWidget.jsx
```diff
@@ -0,0 +1,64 @@
+import React from 'react'
```
DanielSeehausen: this file was just moved (not actually new). It received some minimal changes, but its mostly the same

src/components/device/spinners/LEDCircle.jsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react'
+
+const animation = (status) => (
+  status === 'Running' ? 'lds-spin' : 'lds-frozen'
+)
+
+const LEDCircle = ({ status }) => (
+
+  <div className="lds-css ng-scope">
```
DanielSeehausen: div crazy town is from the package (I assume every circle has actually two div sets)

src/components/devices/DeviceList.jsx
```diff
@@ -1,7 +1,16 @@
 import React from 'react'
 import { Grid, Card } from "semantic-ui-react"
 
-import DeviceWidget from './DeviceWidget'
+import DeviceWidget from 'components/device/DeviceWidget'
+
+
```
DanielSeehausen: this method below was abstracted out of the render. The `index` key was removed because i didn't see it used anywhere. The `device.id` was updated to `device.deviceId` (<-- I think I did that wrong before)

src/index.js
```diff
@@ -4,6 +4,7 @@ import store from "./redux/store"
 import { Provider } from "react-redux"
 
 import './styles/semanticoverrides.css'
+import './styles/device-spinner.css'
```
DanielSeehausen: perhaps this would be better imported in the component it's being used in? any thoughts here @vdiezel?

src/styles/device-spinner.css
```diff
@@ -0,0 +1,293 @@
+@keyframes lds-spin {
```
DanielSeehausen: this is all from the code we found online
only difference is we sloppily duplicated all classes so we can control whether or not the graphic is spinning or not

test/mocks/deviceDataMock.js
```diff
@@ -62,4 +62,22 @@ export const getMockedDeviceData = () => ({
   calibration: [],
   streamDefaults: [],
   streamNameAliases: [],
+  ...data
```
DanielSeehausen: this was added even though we aren't using it. I like the pattern below a lot (its similar to the trait pattern I used in Hambda, but its explicit, and they can be chained pretty easily)

src/index.js
vdiezel: For now this is okay  - let's postpone those questions when we actually are going to set up a pattern for the styling of the web app

src/styles/device-spinner.css
```diff
@@ -0,0 +1,293 @@
...
+  transform: rotate(360deg);
+  -webkit-transform-origin: 154px 100px;
+  transform-origin: 154px 100px;
+}
+.lds-frozen {
+  width: 200px !important;
+  height: 200px !important;
+  -webkit-transform: translate(-100px, -100px) scale(1) translate(100px, 100px);
+  transform: translate(-100px, -100px) scale(1) translate(100px, 100px);
+}
```
vdiezel: missing new line here

test/mocks/deviceDataMock.js
vdiezel: I like it.

src/styles/device-spinner.css
petros751: I fix it

PR #178, opened 2019-09-15
replaced dropdown in process tile with buttons like the trash icon

### LINKED ISSUE #153: The dropdown menu on the process tiles should be more accessible
# Current Behaviour

The dropdown shows up when clicking on the caret in the tile which is a quite small area 

# Desired Behaviour 

Find a solution that is easier to hit

vdiezel:
closes #153 

Dispalying buttons for *clone*, *edit* and *access* as buttons just like the `delete` button

For Mobile we gotta make sure to provide a different solution because there is no hover effect


vdiezel:
Bildschirmfoto von 2019-09-15 08-57-36

PR #179, opened 2019-09-15
Small code improvements

### LINKED ISSUE #165: abstract error handling in the thunks `catch` blocks
the current implementation violates the DRY pinciple. The error handling as is looks weird because some error might be caused by frontend code, then there will be no `response` props on the caught error object. If it is a backend issue then there will be a reponse object aand want to logout the exact error with is nested in `e.response.data`

vdiezel:
Small code cleanups and bugfixes I found while doing it

- abstracted error logging eand extracting in thunks
- deleted the top level tab processes completely - it wasn't reusing the process table component from device and I think nobody really uses this part of the application for now. We will do better later
- fixed a bug in device access modifcation

# closes #165


src/components/about/About.jsx
```diff
@@ -1,7 +1,53 @@
 import React from 'react'
-import { Divider, Container, Segment, Header, Grid, List } from "semantic-ui-react"
+import { Container, Segment, Header, Grid, List } from "semantic-ui-react"
 import { withRouter } from "react-router-dom"
 
+const added = [
```
vdiezel: made About compoenent a bit DRYer

src/components/device/log/LogList.jsx
```diff
@@ -87,7 +87,7 @@ class LogList extends React.Component {
       }
       this.setState({filteredLogIds : filteredLogIds}, () => {
         this.updateTotalPages()
-        this.updateActivePage(null, { activePage: 2})
+        this.updateActivePage(null, { activePage: 1 })
```
vdiezel: bug

src/components/device/manageaccess/ManageAccess.jsx
```diff
@@ -9,55 +9,42 @@ const mapStateToProps = state => ({
...
-          <AccessDeviceModal
-            open={showNewUserModal}
-            activeDevice={this.props.activeDevice}
-            closeHandler={this.closeNewUserModal.bind(this)}
-          />
-        </Grid.Row>
-      </div>
-    )
-  }
+const ManageAccess = ({ deviceAccess, activeDevice }) => {
```
vdiezel: made component functional

src/components/device/manageaccess/User.jsx
```diff
@@ -10,85 +10,78 @@ const mapStateToProps = state => ({
 
 const mapDispatchToProps = dispatch => ({
   callRevokeDeviceAccess: (deviceId, userId, groupName, hostName) =>
-    dispatch(callRevokeDeviceAccess(deviceId, userId, groupName, hostName))
+  dispatch(callRevokeDeviceAccess(deviceId, userId, groupName, hostName))
 })
 
-class User extends React.Component {
+const User = props => {
```
vdiezel: functional component now

src/components/device/manageaccess/UserTable.jsx
```diff
@@ -12,58 +12,55 @@ const mapStateToProps = state => ({
 
 const mapDispatchToProps = dispatch => ({
   callModifyDeviceAccess: (deviceId, userId, newGroup, oldGroup) =>
-    dispatch(callModifyDeviceAccess(deviceId, userId, newGroup, oldGroup))
+  dispatch(callModifyDeviceAccess(deviceId, userId, newGroup, oldGroup))
 })
 
-class UserTable extends React.Component {
+const UserTable = props => {
```
vdiezel: functional now

src/components/device/manageaccess/UserTable.jsx
```diff
@@ -12,58 +12,55 @@ const mapStateToProps = state => ({
...
+        <Table.Row>
+          <Table.HeaderCell>Username</Table.HeaderCell>
+          <Table.HeaderCell>Email</Table.HeaderCell>
+          <Table.HeaderCell>Group</Table.HeaderCell>
+          <Table.HeaderCell>Actions</Table.HeaderCell>
+        </Table.Row>
+      </Table.Header>
+      <Table.Body>
+        {deviceUsers.map((user) =>
+          <User
```
vdiezel: only reformatting a bit

src/components/device/modals/AccessDeviceModal.jsx
```diff
@@ -10,126 +10,126 @@ const mapStateToProps = state => ({
 
 const mapDispatchToProps = dispatch => ({
   callGrantDeviceAccess: (deviceId, email, groupName, hostName) =>
-    dispatch(callGrantDeviceAccess(deviceId, email, groupName, hostName))
+  dispatch(callGrantDeviceAccess(deviceId, email, groupName, hostName))
 })
 
-class AccessDeviceModal extends React.Component {
+const AccessDeviceModal = props => {
```
vdiezel: made functional - when you have multiple local state variables, you can either call `useState` on each of those or use a state object in `useState` instead (you can also mix those two pttern if you want to). The latter has been chosen - state management looks then quite similar to `class` based approach, however the `setValues` function that is used instead of `setState` needs always to return the full state, similar to a `reducer`

src/components/main/Main.jsx
```diff
@@ -70,18 +69,6 @@ class Main extends React.Component {
               <Icon name='hdd'/>
               Devices
             </Menu.Item>
-            <Menu.Item active={activeTab === 'processes'}
```
vdiezel: Process are gone from the top level now

src/components/notifications/Invitation.jsx
```diff
@@ -0,0 +1,50 @@
...
+const mapStateToProps = state => ({
+    user: state.user,
+})
+
+const mapDispatchToProps = dispatch => ({
+    callAcceptDeviceInvitation: (userId, invitationId) => dispatch(callAcceptDeviceInvitation(userId, invitationId)),
+    callDeclineDeviceInvitation: (userId, invitationId) => dispatch(callDeleteUserNotification(userId, invitationId))
+})
+
+const Invitation = props => {
```
vdiezel: abstracted the invitation component

src/redux/thunks/thunks.js
```diff
@@ -35,6 +35,12 @@ import {
   resetDeviceInitialisation
 } from "../actions/actions"
 
+const logAndExtractError = error => {
```
vdiezel: function checks if thee error is an API  error  - it will have the `response` key in the error object because `axios` is used by our sdk, or if it is a simple error in our code. It logs the error and returns it so it can be passed back to the component

src/redux/thunks/thunks.js
```diff
@@ -65,25 +68,19 @@ export function callDeleteEphemeralNotifications(userId, notifications){
...
     }
   }
 }
 
-export function callDeleteUserNotification(notificationId) {
+export function callDeleteUserNotification(userId, notificationId) {
   return async dispatch => {
     try {
-      await nexus.user.deleteNotification(notificationId)
+      await nexus.user.deleteNotification(userId, notificationId)
```
vdiezel: bug!

src/components/about/About.jsx
DanielSeehausen: very nice - love these changes

src/components/device/manageaccess/ManageAccess.jsx
```diff
@@ -9,55 +9,42 @@ const mapStateToProps = state => ({
...
+
+  const [ showNewUserModal, toggleNewUserModal ] = useState(false)
```
DanielSeehausen: actually #dank

src/components/device/manageaccess/UserTable.jsx
```diff
@@ -12,58 +12,55 @@ const mapStateToProps = state => ({
...
+            access={activeDevice.access}
+            key={user.id}
+            user={user}
+            deviceAccess={deviceAccess}
+            deviceId={activeDevice.id}
+            itsMe={me.email === user.email}
```
DanielSeehausen: will this error if [line 30](https://github.com/ospin-web-dev/REACTor/pull/179/files#diff-67adf38c8fe9edec71ce3077fa08785aR30) above does not trigger true and assign `me`?

separately, if we have `id` available, we should preference to match on that

src/components/device/manageaccess/UserTable.jsx
```diff
@@ -12,58 +12,55 @@ const mapStateToProps = state => ({
...
-              me={me}
-              isOwner = {user.id === activeDevice.ownerId}
-              modifyAccessHandler = {this.modifyAccess.bind(this)}
-            />)
-          })}
-        </Table.Body>
-      </Table>
-    )
+  let me
+  if (deviceUsers.length){
```
DanielSeehausen: we may want to make sure `deviceUsers` is defaulted to [], as it shouldn't ever have a value outside of a populated or empty array. If we want to be lazy and just be defensive within the component, we can use [defaultProps](https://reactjs.org/docs/react-component.html#defaultprops)

then we won't need our :sandwich: code with `let`

src/components/device/menu/DeviceMenu.jsx
```diff
@@ -32,7 +32,7 @@ const DeviceMenu = props => {
   } else if (props.activeDevice.connectionState === "offline") {
     statusColor = "status-grey"
     iconType = "circle outline"
-  }
```
DanielSeehausen: epic 🕵️‍♀️

src/components/device/modals/AccessDeviceModal.jsx
```diff
@@ -10,126 +10,126 @@ const mapStateToProps = state => ({
...
-  close(){
-    this.props.closeHandler()
+  const close = () => {
+    props.closeHandler()
   }
 
-  handleUserUpdate(event, data) {
-    this.setState({newUserEmail: data.value, isError: false, errorMessage: '', isSuccess: false})
+  const handleUserUpdate = (event, data) =>  {
+    setValues({ ...modalState, newUserEmail: data.value, isError: false, errorMessage: '', isSuccess: false })
```
DanielSeehausen: i'm I right to assume `isSuccess` isn't true, even though `isError` is false, is because this is just an update and not a successful post?

DanielSeehausen: nvm i see that below -- nice!

src/components/notifications/Invitation.jsx
DanielSeehausen: very nice!

both destructuring props, as well as calling from props directly happens here. Consider preferencing one or the other

I like the ` const { inv } = props ` a lot, because I immediately know what you are going to use. Seeing that either directly in the parameter ()s, or in the body, is super nice, i'd just include the other props you access in the JSX when you do so.

src/components/notifications/Notifications.jsx
```diff
@@ -1,19 +1,15 @@
...
 const mapStateToProps = state => ({
     user: state.user,
-    authenticated: state.authenticated,
 })
 
-const mapDispatchToProps = dispatch => ({
-    callAcceptDeviceInvitation: (userId, invitationId) => dispatch(callAcceptDeviceInvitation(userId, invitationId)),
-    callDeclineDeviceInvitation: (userId, invitationId) => dispatch(callDeleteUserNotification(userId, invitationId))
-})
+const renderInvitation = (inv, i) => <Invitation inv={inv} key={i}/>
```
DanielSeehausen: [DANK](https://github.com/cjwilburn/dank-memes/blob/master/index.js)

src/components/notifications/Notifications.jsx
```diff
@@ -25,41 +21,14 @@ const Notifications = props => (
...
-                          onClick={() => props.callAcceptDeviceInvitation(props.user.id, inv.id)}
-                        >
-                          Accept
-                        </Button>
-                      </Item.Extra>
-                    </Item.Content>
-                  </Item>
-                )
-            })}
+              .map((inv, i) => renderInvitation(inv, i)
```
DanielSeehausen: again, very [DANK](https://a.wattpad.com/cover/85857926-288-k105280.jpg)
you can also pass `renderInvitation` directly, and it will automatically pass the element and the index
`.map(renderInvitation)`

src/styles/semanticoverrides.css
```diff
@@ -263,7 +263,7 @@ body {
   top: 50%;
   transform: translate(-50%, -50%);
   left: 50%;
-}
```
DanielSeehausen: 🔫

src/components/device/manageaccess/UserTable.jsx
vdiezel: This component is on my refac radar - it's quite messy what is going on there. This fella needs some sense knocked into

vdiezel: And yes, it would error

src/components/notifications/Notifications.jsx
vdiezel: Ah! I never knew it also passes the index - thought it will complain about missing keys in child components!

src/components/device/manageaccess/UserTable.jsx
DanielSeehausen: viktor stop working

Issue #211, opened 2019-09-23
As a developer, I have a component I can invoke that provides a flash banner across the top of the application


DanielSeehausen:
let's add an ability to flash messages for the user. 

Perhaps we have 3 levels (green yellow red)
Red is for errors, i.e. when someone clicks on a process and we can't actually find it (currently a problem we have, that we shouldn't see), we can invoke something like:
```js
// bad bad error
<Banner type={'error'} duration={2000} msg={`process ${process.id} not found!`} />
// validation error
<Banner type={'error'} duration={2000} msg={`provide a target value for sensor ${sensor.name} between ${min}-${max}`} />
// user feedback
<Banner type={'good'} duration={2000} msg={`User ${removedUser.is} removed from device`} />
```

we probably need to push something into redux state for this, actually, so it can be rendered at the top level, and remove itself from redux after its duration runs its course (or something -- just find a clean implementation where it can be called from anywhere in the application)

check our their flash message at 3:35
except ours will be #dank and fade out

once implemented, fix the implementation [here](https://github.com/ospin-web-dev/REACTor/pull/212/files#diff-e7ec9b436129ede64b890d76b7eb1a0dR67) to make use of it please

PR #212, opened 2019-09-23
fixes url parsing of process from url; will not open an unfound proce…

### LINKED ISSUE #204: error when loading processes
see screenshot!

image

I believe this happens when some processes no longer exist on the backend or are somehow otherwise 'un-fetchable'

I know this is a problem @vdiezel has encountered before, so how can we put our heads together and find a good solution to make it so this bad state is either not gotten into, or a way that when it does happen we reasonably handle it for the user

### LINKED ISSUE #206: Get filter working for devices
Currently, the filter does not work

Please get it working to filter by name (MVP)
If more filters are needed later, we will add them as they are required.

image

DanielSeehausen:
…ss and break

this actually closes #204  - name of branch is wrong.


src/components/device/Device.jsx
```diff
@@ -57,9 +58,18 @@ class Device extends React.Component {
     <DeviceSetup activeDevice={getActiveDevice(this.props.devices, this.props.match.params.deviceId)}/>
   )
 
-  renderProcessScreen = () => (
-    <ProcessScreen activeDevice={getActiveDevice(this.props.devices, this.props.match.params.deviceId)}/>
-  )
+  renderProcessScreen = () => {
+    const activeProcess = getActiveProcess(this.props.processes, this.props.match.params.processId)
+    const activeDevice = getActiveDevice(this.props.devices, this.props.match.params.deviceId)
+
```
DanielSeehausen: this is not ideal - and I created a ticket to implement a better solution #211

src/components/devices/Devices.jsx
```diff
@@ -4,21 +4,22 @@ import { connect } from "react-redux"
 import { withRouter } from "react-router-dom"
 import DeviceList from './DeviceList'
 import { resetActiveDevice } from "../../redux/actions/actions"
-import { callCreateDevice, callGetDevices, callGetProcesses } from "../../redux/thunks/thunks"
+import { callStopProcess, callCreateDevice, callGetDevices, callGetProcesses } from "../../redux/thunks/thunks"
```
DanielSeehausen: this is being used to fix an ongoing error in a block of code not present in this PR :/
I will take out soon

src/components/main/Main.jsx
```diff
@@ -108,6 +108,7 @@ class Main extends React.Component {
             <Route exact path='/about' component={About}/>
             <Route exact path='/profile' component={Profile}/>
             <Route exact path='/notifications' component={Notifications}/>
+            <Route path='/device/:deviceId/process/:processId' component={Device}/>
```
DanielSeehausen: this sneaky devil was matching on line 112 below, and then never actually populating the matched :processId param, making it inaccessible

PR #214, opened 2019-09-26
142 edit button for device title description Front End part

petros751:
I have use the state hook for the device input name. I have disabled the `clickHandler` from the Card just for testing purpose.


src/components/device/DeviceWidget.jsx
```diff
@@ -25,21 +25,54 @@ const renderVirtual = () => (
   <Label color='blue' basic className="float-right" size="small">
     Virtual
   </Label>
-)
+) 
 
 const DeviceWidget = ({ device: { connectionState, isVirtual, name, state }, clickHandler }) => {
```
DanielSeehausen: spacing here got messed up I think!

src/components/device/DeviceWidget.jsx
```diff
@@ -25,21 +25,54 @@ const renderVirtual = () => (
...
 
-  const { image, statusText, textColor } = displayMap[connectionState][state] || DEFAULT_DISPLAY
+const { image, statusText, textColor } = displayMap[connectionState][state] || DEFAULT_DISPLAY
+
+const [ descriptionEditable, setDescriptionEditable ] = useState(false)
+
+const [descriptionsValues, setDescriptionsValues] = useState({newDeviceName: ''})
```
DanielSeehausen: perhaps `deviceDescription` and `setDeviceDescription` are a little more accurate!

I don't think we need an object here in `useState`. We should just be able to use the string that comes with the device description as the first default value.

src/components/device/DeviceWidget.jsx
```diff
@@ -25,21 +25,54 @@ const renderVirtual = () => (
...
+}
 
   return (
-    <Card onClick={clickHandler}>
+    <Card>
       <Card.Content>
         <Image floated="left" size='tiny' src={image}/>
-        <Card.Header>{name}</Card.Header>
+        <div>
+          {descriptionEditable ?
```
DanielSeehausen: I like the pattern here how, when they click the edit button, the fields transform. It seems cleaner than a modal. Nice work!

src/components/device/DeviceWidget.jsx
```diff
@@ -25,21 +25,54 @@ const renderVirtual = () => (
...
+            <Card.Header> <input type="text" onChange={handleInputChange} onBlur={submitNewDeviceDescriptions} defaultValue={name} autoFocus /> </Card.Header>
```
DanielSeehausen: this code block should be formatted in normal html format

src/components/device/DeviceWidget.jsx
```diff
@@ -25,21 +25,54 @@ const renderVirtual = () => (
...
+            :
```
DanielSeehausen: if we have a ternary that leads to a multiline code block in a react render method, its usually best practice to make functions that return the jsx instead.

So something like: `{ x ? renderEditableName() : renderName() }`
so other people can understand the code quickly and cleanly!

src/components/device/DeviceWidget.jsx
```diff
@@ -25,21 +25,54 @@ const renderVirtual = () => (
...
-        </Card.Description>
+        <Card.Description> Standard Ospin bioreactor platform </Card.Description>
+        <Button 
+          className='float-right margin-l-4' 
+          size='mini' 
+          basic 
+          circular 
+          icon='edit' 
+          onClick={() => setDescriptionEditable(true)}
+          />
```
DanielSeehausen: this looks great! the `/>` should be un-indented once

src/components/device/DeviceWidget.jsx
```diff
@@ -25,21 +25,54 @@ const renderVirtual = () => (
...
+const [descriptionsValues, setDescriptionsValues] = useState({newDeviceName: ''})
+
+const handleInputChange = (e) => {
+  setDescriptionsValues({...descriptionsValues, newDeviceName: e.target.value})
+}
+
+const submitNewDeviceDescriptions = async() => {
+  const { newDeviceName } = descriptionsValues
+
+  if (newDeviceName === name) {
```
DanielSeehausen: this can be easiest asked at the top of the function:
`if (description === props.device.description) return`

DanielSeehausen: this can be easiest asked at the top of the function:
```
const submitNewDeviceDescriptions = await () => {
  setDescriptionEditable(false) // this always happens on blur
  if (description === props.device.description) return
  // if description is a string of < n characters
  const { statusCode, body } = await updateDevice(device.id, { description })
  // if status code is not 200, display errors
```

src/components/device/DeviceWidget.jsx
```diff
@@ -25,21 +25,54 @@ const renderVirtual = () => (
   <Label color='blue' basic className="float-right" size="small">
     Virtual
   </Label>
-)
+)
```
DanielSeehausen: where is this sneaky guy coming from?

petros751:
OK @DanielSeehausen  I will start right now to fix the changes that you requested and I will add the function from the Nexus

DanielSeehausen:
@petros751 I am going to close this for now as its been open without progress for quite awhile. 
Please feel free to re-open if you have the time to make changes to it (it will need a rebase too).

PR #215, opened 2019-09-26
112 display operating user

### LINKED ISSUE #112: as a user I see who is currently running the device
whoever picks this ticket up, please split it out into another, where the first ticket to be completed is for updating the backend to store information re: who started a process 

then the other ticket is strictly for the UI

[child](https://github.com/orgs/ospin-web-dev/projects/10#card-26851215)

DanielSeehausen:
closes #112 

~will not go in until repeat rendering is fixed~
rendering fixed! testing now!


src/components/device/DeviceWidget.jsx
```diff
@@ -27,23 +29,57 @@ const renderVirtual = () => (
...
+  </Card.Description>
+)
+
+const renderNotRunning = () => (
+  <Card.Description>
+    Standard Ospin bioreactor platform
+  </Card.Description>
+)
+
+const limitString = (str, limit) => ( str.length > 20 ? `${str.substring(0, 20)}...` : str)
```
vdiezel: consider using a default parameter as llimit and use it in the function

DanielSeehausen: ill change - that's a good idea

DanielSeehausen: updated! good 👁 that was a mistake on my part

PR #216, opened 2019-09-26
added a confirmation modal for changes to the running phase and a sho…

### LINKED ISSUE #201: provide user with information about action to change processes
Perhaps we should require confirmation when a user is changing an actively running phase. This also gives us the opportunity to educate them on what is actually happening: "Hey, you are about to change an actively running phase. This will create a new phase immediately there-after and transition it to the active phase (maybe we show them a graphic or something?). I can see how, from a user perspective, the behavior may look like a bug without us explaining it

vdiezel:
…rt explanation for the user

closes #201


src/components/processviewer/body/builder/ChangesFooter.jsx
```diff
@@ -18,32 +20,39 @@ const mapDispatchToProps = dispatch => ({
...
+    props.setProcessBuilderValidationErrors(errors)
 
-const ChangesFooter = props  => {
+    if (errors.length) return
+
+    const runningPhaseId = getRunningPhase(activeProcess).phaseId.toString()
+    const insertPhase = PDVC.hasStreamTargetChange(diffs, runningPhaseId)
 
-  const { activeProcess, processDescriptionSnapshots } = props
+    if (insertPhase) {
```
vdiezel: logi where we split between immediate submit and showing the confirmation modal first

src/components/processviewer/body/builder/PhaseTabContent.jsx
```diff
@@ -121,6 +121,12 @@ class PhaseTabContent extends React.Component {
             </Fragment>
           }
         </Form>
+        {this.props.activeProcess.state === 'running' ?
```
vdiezel: moved Changesfotter to the top of the view

src/utils/PDVC.js
```diff
@@ -87,8 +87,8 @@ class PDVC {
     return diffs.some(diff => diff.phaseId === runningPhaseId)
   }
 
-  static hasStreamTargetChange(diffs){
-    return diffs.some(diff => diff.type === 'streamTargetChange')
+  static hasStreamTargetChange(diffs, phaseId){
```
vdiezel: This was a bug - I wanted to check if a certain phase has streamTargetChanges (for now, we are mainhly interested in the running phase). This caused the insertion of a phase on ad hoc changes when we did modify the runningPhase, but not in a way that needed the insertion of a phase (e.g. chaging the transition method) AND changed the stream target of a future phase.

src/components/processviewer/body/builder/ChangesFooter.jsx
```diff
@@ -18,32 +20,39 @@ const mapDispatchToProps = dispatch => ({
...
+      triggerConfirmationModal(true)
+      return
+    }
+    else await confirmChanges(diffs)
```
DanielSeehausen: no need for else here, as its always

vdiezel: :man_facepalming:

PR #224, opened 2019-09-30
Code cleanup 2

### LINKED ISSUE #223: dry up the reducer

vdiezel:
fixes #223


src/components/device/manageaccess/User.jsx
```diff
@@ -13,7 +13,8 @@ const mapDispatchToProps = dispatch => ({
   dispatch(callRevokeDeviceAccess(deviceId, userId, groupName, hostName))
 })
 
-const User = props => {
+const User = ({ deviceId, user, access, isOwner, me, itsMe, deviceAccess,
```
vdiezel: just a bit of cleanup

src/components/device/processes/ProcessesTable.jsx
```diff
@@ -59,7 +59,7 @@ class ProcessesTable extends React.Component {
         const {selectedIds} = this.state
         if (selectedIds.length > 0) {
           this.setState({ executingAction: true })
-          await this.props.deleteProcessesHandler(selectedIds)
+          await this.props.callDeleteProcesses(selectedIds)
```
vdiezel: bug in the process table view

src/components/processviewer/body/builder/Group.jsx
```diff
@@ -1,94 +1,87 @@
-import React, { Fragment } from 'react'
+import React, { Fragment, useState } from 'react'
 import {Form, Grid, Menu, Accordion, Dropdown} from "semantic-ui-react"
 import Stream from './Stream'
 import ControllerModal from "./ControllerModal"
 
-class Group extends React.Component {
+const Group = ({ activeProcess, activeDevice, phaseId, phase, processDownStreams,
```
vdiezel: made Group component functional
vim indentation messes up when breaking code into multiple lines (line 6)

src/redux/actions/actions.js
```diff
@@ -39,15 +35,15 @@ export const ACTIONS = {
...
   UPDATE_PROCESS_STATE: "UPDATE_PROCESS_STATE",
   SWITCH_PROCESS_PHASE: "SWITCH_PROCESS_PHASE",
   ADD_LIVE_STREAM_DATA: "ADD_LIVE_STREAM_DATA",
-  DELETE_STREAM_CALIBRATION: "DELETE_STREAM_CALIBRATION",
   REMOVE_NOTIFICATIONS: "REMOVE_NOTIFICATIONS",
   CREATE_PROCESS_DESCRIPTION_SNAPSHOT: "CREATE_PROCESS_DESCRIPTION_SNAPSHOT",
   UPDATE_PROCESS_DESCRIPTION_SNAPSHOT: "UPDATE_PROCESS_DESCRIPTION_SNAPSHOT",
-  RESET_DEVICE_INITIALIZATION: "RESET_DEVICE_INITIALIZATION"
+  RESET_DEVICE_INITIALIZATION: "RESET_DEVICE_INITIALIZATION",
+  DELETE_STREAM_CONFIG_PROPERTY: "DELETE_STREAM_CONFIG_PROPERTY",
```
vdiezel: the two new actions replace setting/deleting strea defaults, name aliases and calibrations

src/redux/helper/updateFunctions.js
```diff
@@ -0,0 +1,14 @@
+export const updateEntity = (id, entities, updater, params) => {
```
vdiezel: that is a handler function for the reducer, usage see in reducer

src/utils/device.js
```diff
@@ -24,3 +24,53 @@ export function getRawDataValue(device, streamId, calibratedValue){
   }
   return calibratedValue
 }
+
+export function removeDeviceUser(device, userId){
```
vdiezel: a bunch of helper functions for the device - used in the reducer

src/redux/reducers/reducer.js
```diff
@@ -75,34 +78,23 @@ function rootReducer(state = initialState, action) {
   }
   else if (action.type === ACTIONS.SET_STREAM_TARGET){
 
-    const processes = state.processes
+    const { processes } = state
+    const { processId, phaseId, streamId, value } = action.payload
+    const params = [ phaseId, streamId, value ]
+
+    const updatedProcesses = updateEntity(processId, processes, setStreamTarget, params)
```
vdiezel: `updateEntity` takes an `id` and an array of a resource/entity kind, queries by Id and does the following: 

If an update function is provided, e.g. `setStreamTarget`, it will apply the update function on the entity with the id that is being provided as the first argument to `updateEntity`. The update function always will receive the entity as the first argument, followed by the arguments being provided in params, e.g. in this case it will call `setStreamTarget(process, phaseId, streamId, value)` where process is the one fetched from `processes` using `processId`. 
It is also possible to provide several update functions in an array ( see MODIFY_PROCESS_ACCESS )

If no update function is provided, it will simply create a copy of the entity, and destructuring the provided params into it (see UPDATE_PROCESS).

src/redux/reducers/reducer.js
```diff
@@ -260,233 +238,102 @@ function rootReducer(state = initialState, action) {
...
-    const devices = state.devices.map(device => {
-      if (device.id === deviceId) {
-        addUserToGroup(device, newGroup, userId)
-        removeUserFromGroup(device, oldGroup, userId)
-        return {...device}
-      }
-      return device
-    })
+    const updatersDevice = [ addUserToGroup, removeUserFromGroup ]
+    const paramsDevices = [[ newGroup, userId ], [ oldGroup, userId ]]
```
vdiezel: example for 2 update functions

src/redux/reducers/reducer.js
```diff
@@ -260,233 +238,102 @@ function rootReducer(state = initialState, action) {
...
+    const { processes } = state
+    const { processId, params} = action.payload
 
-    const updatedProcesses = processes.map(process => {
-      if (process.id === action.payload.processId) {
-        return {...process, ...action.payload.params}
-      }
-      return process
-    })
+    const updatedProcesses = updateEntity(processId, processes, null, params)
```
vdiezel: example of simple destructuring of object properties into an entity

src/components/device/manageaccess/User.jsx
```diff
@@ -22,29 +23,29 @@ const User = props => {
   const handleClose = () => toggleDeletePopup(false)
 
   const deleteAndClose = async (userId, groupName) => {
-    await props.callRevokeDeviceAccess(props.deviceId, userId, groupName, props.me.userName)
+    await callRevokeDeviceAccess(deviceId, userId, groupName, me.userName)
   }
```
DanielSeehausen: just testing new multi-line comments feature

src/redux/reducers/reducer.js
```diff
@@ -260,233 +238,102 @@ function rootReducer(state = initialState, action) {
...
 
-    const devices = state.devices.map(device => {
-      if (device.id === deviceId) {
-        addUserToGroup(device, newGroup, userId)
-        removeUserFromGroup(device, oldGroup, userId)
-        return {...device}
-      }
-      return device
-    })
+    const updatersDevice = [ addUserToGroup, removeUserFromGroup ]
```
DanielSeehausen: this name is somewhat misleading
maybe just `updates` or `deviceUpdates` or `accessUpdates`

src/redux/reducers/reducer.js
DanielSeehausen: and perhaps this something along the lines of `updateArgs`
I only pick on this because the helper function is somewhat complex with the optional arrays, so we want to be extra careful for people reading it, and params are those things in a function declaration, and arguments are those things passed in to the params.

src/redux/helper/updateFunctions.js
```diff
@@ -0,0 +1,14 @@
...
+      if (Array.isArray(updater)){
+        updater.forEach((updateFn, i) => updateFn(entity, ...params[i]))
+      }
+      else updater(entity, ...params)
+      return { ...entity }
+    }
+    return entity
+  })
+}
+
```
DanielSeehausen: I like it a lot! good abstraction. I think we could improve the readability a little bit

First the code smell things:
 - if we can get away with nesting blocks less by returning early, we should. Avoid the if block, and instead ask the inverse and return early, because we are essentially saying "this should not run if this condition is not met"
 - if we are wrapping `if` with curly boys, we should maintain the same pattern with the `else`, so you don't have to do visual gymnastics to read the control flow.
 - a function should do one thing and one thing well, if possible. It pays off big time in the long run to try to stay strict to this
 - on line 11, we aren't spreading the object, so this won't shallow clone the array

If I am understanding this, our top concern is "were their update functions passed?", and depending on that, we basically perform one of two things (aka we should split functions here). But then, we would be duplicating the map, which isn't any fun. Luckily, we can get around this as well, by first finding the element to be changed.

```js
// updaters is optional, function should fail if params aren't passed, but not updaters
// defaults to Object.assign (what the spread method of combining objects is syntactic sugar on top of)
export const findAndMultiUpdateEntity = (id, entities, params, updaters = [ Object.Assign ]) => {
  // because this function is increasing in complexing, it deserves some early error handling
  if (params.length !== updaters.length) throw '...whatever'

  const targetIdx = entities.findIndex(...)
  const targetEntity = { ...entities[targetIdx] }

  const mutatedEntity = updaters.forEach(( fn, idx ) => fn(targetEntity, params[idx]))
  return [ ...entities.slice(0, targetIdx), targetEntity, ...entities.slice(targetIdx+1) ]
}


// unfortunately we can't do 'operator overloading' in JS right now, but we can get away with a more functional approach
export const findAndUdateEntity = (id, entities, params, updater = Object.Assign) => (
  findAndMultiUpdateEntity(id, entities, [ params ], [ updater ])
)

// something like this abstracts the implementation from the user, while still providing them a clear API with the functions that don't handle multiple data types for the same paremeters

DanielSeehausen: If I were writing this PR, I would be scared about screwing up somewhere in the implementation of that code, and understanding how widely we will likely be using this function, probably look to provide test coverage for it where I pass in a bunch of different values (if only to make sure I wrote it write, rather than ensure it keeps working over time with refactors). That being said, use your best judgement for whether its worth your time to do that or now.

src/redux/reducers/reducer.js
DanielSeehausen: I love the new function, and this is a great case where its use is super sexy. I think we can provide a more descriptive name for it, and even a separate function name for when its a single function

PR #246, opened 2019-10-16
Update issue templates

vdiezel:


vdiezel:
I have added deviceID and processID to the sheet to be able to track down the problems a bit easier

PR #250, opened 2019-10-23
conditional rerendering of the data graphs to make rendering of data …

### LINKED ISSUE #249: Fix constant rerender of all graphs when new data comes in

vdiezel:
…graph view faster.

closes #249


src/components/processviewer/body/dataviewer/Group.jsx
```diff
@@ -10,6 +10,8 @@ const Group = props => {
                     <Header as="h4">{props.groupName}</Header>
                 </Grid.Row>
                 {props.groupStreams.map((stream, i) => {
+                    const completeStreamData = props.activeProcess.data
```
vdiezel: streamData now passed in as a prop

src/components/processviewer/body/dataviewer/Stream.jsx
```diff
@@ -199,16 +199,21 @@ class Stream extends React.Component {
     this.setState({selectedPhase: data.value, mode: 'phase', zoomParams})
   }
 
+  shouldComponentUpdate(nextProps, nextState){
```
vdiezel: rerenders only when a.) data has changed (load initial data or fetching more data) or b.) state (zooming, panning, etc..)

src/components/processviewer/body/dataviewer/Stream.jsx
```diff
@@ -199,16 +199,21 @@ class Stream extends React.Component {
...
+    return !(this.props.completeStreamData === nextProps.completeStreamData &&
+    this.state === nextState)
+  }
+
   render() {
```
vdiezel: Just a bit cleanup

src/components/processviewer/body/dataviewer/Stream.jsx
DanielSeehausen: Are we ok to be running hard equality on these values? And is it what we want? I can picture potential side effects. 

I believe the code isn't quite doing what you wrote above. I am reading it as:

```
update if:
  the new props.completeStreamData is the same array as previously (not if the data is the same)
or
   the state is the same object
```

src/components/processviewer/body/dataviewer/Group.jsx
vdiezel: `completeStreamData` means that this is the complete stream data for a specific stream (which will be than truncated to what is going to be displayed). `activeProcess.data` contains an array of objects like this `{id: streamId, dataPoints: [{time: 5, value: 37}, ...]}`

src/components/processviewer/body/dataviewer/Stream.jsx
vdiezel: Yes, you are right, I compare references, which works because I'm updating stream data in a functional way. Comparing the data in the array would be much more costly here. Same idea applies to the state of the component. I should have been more concise on the giben explanation.

vdiezel: Which side effects do you mean?

src/components/processviewer/body/dataviewer/Group.jsx
DanielSeehausen: ah makes sense! thanks

src/components/processviewer/body/dataviewer/Stream.jsx
DanielSeehausen: I believe I was confused because of the way the logic is structured around exclusion of possibilities, and that considering state was brought in to the logic. I see now what you were testing before, but it wasn't clear/concise before to me that the issue was props other than completeStreamData updating. 

Re: side effects, imagine the parent component is extended, and starts passing an additional prop. 
And we use this prop in the component, and it changes semi-regularly.

This would lead to immediate unexpected behavior for the developer, because the component is not updating. We have a pretty big component as it is, and it puts the burden on them to either guess correctly that there is some special use of shouldComponentUpdate, or them reading around trying to find out what's going on. I know that wouldn't be my first instinct. 

What value coming through in props is the problem value that is constantly updating? It would be a lot clearer if we instead said "hey, if this value is what changed, don't update", than the inverse, which is "if anything changed that isn't this special problematic prop value, then update". If we can identify that prop, or that's an option, we can make this change without breaking the expected API of a react component. If there is a reason we can't do that, then ignore everything I wrote above this. 

(If we can't actually exclude based on whichever prop is causing the issue) the only request I have is to update the spacing:

```js
return !(
  this.props.completeStreamData === nextProps.completeStreamData &&
  this.state === nextState
)
```

vdiezel: The prop that constantly changed was `activeProcess` because we add data to `activeProcess.data` in  a functional way. Everytime a request returns data for a stream (which is done for every single upstream/graph ) we create a copy of the process - but not a deep one when I think of it right now, because then the references wouldn't match anymore and the first condition in `componentShouldUpdate` would always be `false`. 

Your point with passing down more props from the parent component is valid. The behaviour that we want to have is this: When the `completeStreamData` hasn't changed, don't rerender. Of course this would make the component not listen to any state changes anymore either (when user zooms but data didn't change).  So I had to add the second condition. It can be done a bit better using an `OR` here:

`return this.props.completeStreamData !== nextProps.completeStreamData || this.state !== nextState`

The problem is, that we can't prevent a rerender when the activeProcess has changed (the prop that causes the abundant rerenders) because the graph gets its data from it.

We can try adding data to the process a different way (not copying the process, but only `process.data`) or changing the way data is stored in the application state.

Let me fiddle around

vdiezel: So we have this problem: 

1. When we store all data for a single process in the same object/array we need to create a copy of the object/array otherwise the reducer will not trigger any rendering. If we do a copy however, every graph will be rerendered because the copy is the source of its data and the reference of the object/array has changed.

2. We want the data appear as soon at it is loaded and not wait until all data has been loaded. This means we cant just wait until we got the data from all streams (some streams might take 4-6 seconds to load and then the user just stares at empty graphs for that time, even though 95% of streams returned after 200ms) - if we could wait, we would simply populate all graphs in a single rendering call.

So to provide early rendering and prevent excessive rendering we would need to decouple the containers for the stream data and we would need to put them on top level in application state so the reducer sees when they change. You could do some crazy solutions with object keys that are built from processId + streamId to make them unique and use them in the store to set the data there 

You see any other options?

PR #251, opened 2019-10-29
Admin device ssh tool

DanielSeehausen:
adds ssh request component
viewable by admins


src/components/device/manageaccess/ManageAccess.jsx
```diff
@@ -9,40 +10,48 @@ const mapStateToProps = state => ({
...
-    <AccessDeviceModal
-      open={showNewUserModal}
-      activeDevice={activeDevice}
-      closeHandler={() => toggleNewUserModal(false)}
-    />
-  </Grid.Row>
+
+      { user.roleActions.openSsh ?
+      <Grid.Row>
+        <Divider key={1}/>
```
vdiezel: Because you provide two sibling nodes, don't yo need a key on SSHAccess, too?

DanielSeehausen:
> Looks good! The only thing I feel uncomfortable with is the css inline styling but we gotta wait until we have some pattern for css setup.
> 
> PS: It's viewable by admins and developers, too 😈

removed a whole bunch of it in this one!

PR #252, opened 2019-10-29
Dashboard View for Devices

### LINKED ISSUE #96: As a user I have access to device dashboard  (view mode)
In the device view a user should have access to a device dashboard whicch displays the latest values of all upstreamms published to the client

Level 1: Just displaying all values in table/List
Level 2: Drag'n'Drop to arange the dashboard as desired
Level 3: SVG-Background of the device with the streams aranged to their actual position

vdiezel:
Track the real time state of your device modules in the new dashboard view. Add the modules that mean most to you and benefit from the quick and simple overview of your process state in a single glance. If you want to go more into detail, just use the quick access button that will bring you comfortably to the full process data history of your module.

[Bildschirmfoto von 2019-10-29 21-51-13](https://user-images.githubusercontent.com/37934155/67809247-a7480a00-fa98-11e9-8ca9-e45ad22db2c2.png)

Bildschirmfoto von 2019-10-29 22-06-08


src/components/device/Device.jsx
```diff
@@ -7,10 +7,11 @@ import ManageAccess from './manageaccess/ManageAccess'
 import LogList from './log/LogList'
 import Processes from './processes/Processes'
 import DeviceSetup from './setup/DeviceSetup'
+import Dashboard from './dashboard/Dashboard'
 import {Grid, Loader} from 'semantic-ui-react'
 import ProcessScreen from './processes/ProcessScreen'
 import { getActiveDevice, getDeviceProcesses } from 'utils/device'
-import { getActiveProcess } from 'utils/process'
+import { getProcessById } from 'utils/process'
```
vdiezel: named this more accurate

src/components/device/Device.jsx
```diff
@@ -102,6 +110,7 @@ class Device extends React.Component {
                 <Route exact path='/device/:deviceId' component={this.renderProcesses}/>
                 <Route path='/device/:deviceId/processes' component={this.renderProcesses}/>
                 <Route path='/device/:deviceId/process/:processId' component={this.renderProcessScreen}/>
+                <Route path='/device/:deviceId/dashboard' component={this.renderDashboard}/>
```
vdiezel: dashboard becomes a new top level tab in the device component

src/components/device/dashboard/Dashboard.jsx
```diff
@@ -0,0 +1,103 @@
...
+              deleteHandler={() => removeStreamFromDashboard(activeDevice, stream.id, callUpdateDevice)}
+            />
+          )}
+        </Card.Group>
+      </Segment> : null }
+    </Container>
+  )
+}
+
+export default connect(mapStateToProps, mapDispatchToProps)(Dashboard)
```
vdiezel: So in general I wanted to try a different approach (one that maybe is more REST):

I noticed that we have a lot of backend functions that modify props on the device (streamNameAliases, streamDefaults, streamCalibrations etc) - in a strictly RESTful API this would be done by using `PUT` on the device. We would simply always provide an updated version of e.g. the streamDefaults and overwrite them in the device in the database. Instead we provide some detailed information to a backend function like setStreamDefault and do business logic on the backend to find the property in the device and apply some kind of operation, e.g. setting a new stream default for stream with the `id == =1234`.

Once we receive the OK on the frontend, we actually have to do the exact same logic to apply the changes to the device so the changes become visible in the UI.

Disadvantages of our approach: updateDevice
- data manipulation logic has to run on both ends
- we need a lot of backend functions which is bad because of the test coverage we need, more doors for erros, stack limits for AWS and nexus functions to be written

Advantage: 
- provides possibility for exceptional behavior when certain parameters on the device are set in the backend

I tried a new way here: I apply the data manipulation logic within the component (I started replacing our utils collections with static classes), and then simply call `PUT` on the device (`callUpdateDevice`) and once I receive an OK from the backend, I pass the same update params down to te reducer.

Disadvantage: 
- business logic running within the component. it needs to because e.g. we need the already updated data in the thunk for the backend call and the thunks run before the reducer (where all the business logic should actually be) - so we violate seperation of concerns a bit. But due to abstraction into static helper classes the confusion in the component can be kept minimal
- validation of the `PUT` methods e.g. for `device/update` will become more complex as we need to check which props are supposed to be set and apply validation accordingly

Advantages: 
- much less code
- data manipulation runs only on one side
- better decoupling of frontend and backend 
- more RESTful

src/components/device/dashboard/StreamWidget.jsx
```diff
@@ -0,0 +1,66 @@
...
+              />
+            </div>
+          </div>
+        </Container>
+      </Card.Content>
+    </Card>
+  )
+}
+
+export default StreamWidget
```
vdiezel: a copy of the process tile component to display the streams

src/components/processviewer/ProcessViewer.jsx
```diff
@@ -17,15 +17,14 @@ import {
   PROCESSVIEWER_STATES
 } from 'utils/process'
 import { getMismatchingStreams } from 'utils/validation'
-import { callUpdateProcess, callSubscribeToProcessTopics, callLoadProcessDataSummaryStreamwise } from 'redux/thunks/thunks'
+import { callUpdateProcess, callLoadProcessDataSummaryStreamwise } from 'redux/thunks/thunks'
```
vdiezel: got rid of those - we instead subscribe to all  topics when the device is loaded. I did this in preperation of receiving dataPoints outside of the context of a process.

src/components/processviewer/body/builder/PhaseTabContent.jsx
```diff
@@ -140,7 +140,7 @@ class PhaseTabContent extends React.Component {
         }
         <Divider/>
         <Accordion fluid styled>
-          {Object.entries(groups).map(([groupName, groupStreams], i) => (
+          {groups.map((group, i) => (
```
vdiezel: added helper to make if more readable

src/components/processviewer/modals/DescriptionMismatchModal.jsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react'
 import { Modal, Form, Icon, Button, Message, Divider, Table, List } from "semantic-ui-react"
-import { geStreamNameAlias } from "utils/device"
+import { getStreamNameAlias } from "utils/device"
```
vdiezel: typo

src/redux/helper/objectMapper.js
```diff
@@ -60,6 +60,7 @@ export const createDeviceObject = deviceData => ({
   calibration: deviceData.calibration,
   streamDefaults: deviceData.streamDefaults,
   streamNameAliases: deviceData.streamNameAliases,
+  dashboardStreams: deviceData.dashboardStreams || [],
```
vdiezel: the new property on device for the dashboard : In future I would like to have all stream configuration properties (dashboard, aliases, defaulrs, calibration) maybe in a single object. This would however increase the chance that, using a `PUT` on this object, through the frontend we somehow overwrite everything

src/utils/device.js
```diff
@@ -72,3 +98,40 @@ export function setStreamConfigurationProperty(device, streamId, prop, values){
...
+
+    return Object.entries(groups)
+      .map(([ name, streams ]) => ({ name, streams }))
+      .sort((a, b) => a.name.localeCompare(b.name))
+  }
+
+  static getGroupStreams(device, groupName){
+    return device.streams.filter(stream => stream.groupName === groupName)
+  }
+}
```
vdiezel: added this static class that will contain all helper functions for devices

src/redux/helper/objectMapper.js
DanielSeehausen: that's a good concern.

we can avoid the danger of a PUT overwriting everything by restricting the PUT endpoint to only update certain parameters. Then, if they want to change stream configurations, we can have specific endpoints that do. 

[rails docs really succinctly display how to do this on a high level in their docs here](https://edgeapi.rubyonrails.org/classes/ActionController/StrongParameters.html) Easy to read even if not familiar with ruby!

i.e., the method of data storage does not need to (and rarely does) match the API

src/utils/device.js
DanielSeehausen: love it! very nice

src/components/device/dashboard/Dashboard.jsx
```diff
@@ -0,0 +1,103 @@
...
+import StreamWidget from "components/device/dashboard/StreamWidget"
+import { callUpdateDevice } from 'redux/thunks/thunks'
+
+const mapStateToProps = state => ({
+  processes: state.processes
+})
+
+const mapDispatchToProps = dispatch => ({
+  callUpdateDevice: (deviceId, params) =>
+  dispatch(callUpdateDevice(deviceId, params))
```
DanielSeehausen: this is a little misleading, and inconsistent convention. prolly easiest to bach this up onto the same line (its not that long)

if your heart burns for multiline, consider:
```js
callUpdateDevice: (deviceId, params) => (
  dispatch(callUpdateDevice(deviceId, params))
)
```

src/components/device/dashboard/Dashboard.jsx
```diff
@@ -0,0 +1,103 @@
...
+})
+
+const addStreamToDashboard = async (activeDevice, streamId, updateHandler) => {
+  if (activeDevice.dashboardStreams.some(stream => stream.streamId === streamId)) return
```
DanielSeehausen: opportunity to add `DeviceUtils.hasDashboardStream(device, streamId)`

src/components/device/dashboard/Dashboard.jsx
```diff
@@ -0,0 +1,103 @@
...
+
+  const stream = DeviceUtils.getStreamById(activeDevice, streamId)
+  if (!stream) return
+
+  const updatedDashboardStreams = DeviceUtils.addStreamToDashboard(activeDevice, stream.id)
+  await updateHandler(activeDevice.id, { dashboardStreams: updatedDashboardStreams })
```
DanielSeehausen: is there any sense to pushing this kind of stuff in to the DeviceUtils methods, so that we aren't mutating the local device data and the persisted device data in two different places?

This might make even more sense if the persisting were to fail, then we can meaningfully handle the response and undo/never do the local change, and return only a single result (yes it worked, UI you are good to represent this new state || no something failed, UI don't render as if you succeeded)

src/components/device/dashboard/Dashboard.jsx
```diff
@@ -0,0 +1,103 @@
...
+const createStreamOptions = (activeDevice, groupName) => {
+  return groupName ? DeviceUtils.getGroupStreams(activeDevice, groupName)
+    .filter(stream => stream.type === 'upstream')
+    .map((stream, i) => ({ key: i, text: stream.name, value: stream.id })): []
+}
+
+
+const Dashboard = ({ activeDevice, processes, callUpdateDevice }) => {
+
+  const dashboardStreams = DeviceUtils.populateDashboardStreams(activeDevice)
```
DanielSeehausen: this should be a `getDashboardStreams`. I believe this is generally used when data is being persisted in some manner elsewhere.

src/components/device/dashboard/StreamWidget.jsx
```diff
@@ -0,0 +1,66 @@
...
+
+const redirectToGraph = process => {
+  const newPath = this.props.history.location.pathname.replace('dashboard', 'process')
+  this.props.history.push(`${newPath}/${process.id}`)
+}
+
+const StreamWidget = ({ runningProcess, stream, deleteHandler, activeDevice }) => {
+
+  const value = runningProcess ? getLastStreamDataPoint(runningProcess, stream.id) : ''
+  const unit = stream.unit !== '-' ? stream.unit : ''
```
DanielSeehausen: same request for `null` on the unit

src/components/device/dashboard/StreamWidget.jsx
DanielSeehausen: this component is :sharp: as hell. awesome work

src/components/processviewer/body/dataviewer/Group.jsx
```diff
@@ -1,30 +1,31 @@
...
+                key={i}
+                stream={stream}
+                activeProcess={activeProcess}
+                activeDevice={activeDevice}
+                completeStreamData={completeStreamData}
+              />)
+            })}
+          </Grid>
+        <Divider/>
+      </Container>
```
DanielSeehausen: (only because I see indentation changes in the diff)
this highlighted block above is still double indented

src/utils/device.js
```diff
@@ -6,11 +6,36 @@ export function getDeviceProcesses(processes, deviceId){
...
+
+export function getGroupStreams(device, groupName){
+  return device.streams.filter(stream => stream.groupName === groupName)
+}
+
+export function getStreamById(device, streamId){
+  return device.streams.find(stream => stream.id === streamId)
+}
+
+export function getDeviceUpStreams(device){
```
DanielSeehausen: For future reference (no need to change now), upstream and downstream are single words, so they don't take camelcasing

AG'd this function name to see implications of changing name
and noticed it wasn't being used anywhere. (same with downstreams)

src/components/device/dashboard/Dashboard.jsx
vdiezel: > I am a little less concerned about this one, because we can still provide multiple endpoints (updateDashboardStreams, for example), then do some validation on it, then just call our existing device.update method. We re-use most of the business logic, and adding the additional route/endpoint gives us the safety of asserting only specific parameters get changed

Like this idea!

src/redux/helper/objectMapper.js
vdiezel: I was hoping that one could always use just the `PUT` on device and we don't have to create specific endpoints aka lambda functions for different parameters which also requires creating nexus API call functions. But maybe I'm to greedy here or I missunderstood you. 

I thought of something like this: We have only `PUT` to `device/:deviceId`. There, on incoming requests we check a.) if the propery the user tries to set can be st by him and b.) based on which props should be set apply some validation on this props. That's what I mean by complex validation and I'm afraid this will become a monster...

src/components/device/dashboard/StreamWidget.jsx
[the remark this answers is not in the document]
vdiezel: good catch

src/utils/process.js
```diff
@@ -162,6 +161,16 @@ export function getStreamTarget(process, targetPhaseId, streamId){
     return process.description[targetPhaseId].downstreams[streamId].value
 }
 
+export function getStreamDataPoints(process, streamId) {
+  return process.data.find(streamData => streamData.id === streamId).dataPoints || []
+}
+
+export function getLastStreamDataPoint(process, streamId) {
+  const streamData = getStreamDataPoints(process, streamId)
+  if (!streamData.length) return ''
```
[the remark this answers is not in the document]
vdiezel: agreed - never considered `null` in the past!

src/components/device/dashboard/Dashboard.jsx
vdiezel: On line 25 it returns a copy of the dashboard streams from the device, it does not mutate it. The mutation is happening only once, when we call updateHandler it invokes a thunk. In the thunk, we make a call to the API and after we get the response, we make a call to the reducer where the device is finally mutated. If the API call fails, we are actually returning an error. We could await the return value from the thunk in the component and use it to do display errors - as done it a lot of modals in the WebApp.

src/utils/device.js
vdiezel: I'm thinking what are the advantages of using a static class compared to an object here? With an object, we could also provide nesting, like `Device.dashboard.addStream(...)`

src/components/device/dashboard/Dashboard.jsx
vdiezel: yes!

vdiezel:
- added tests ( could not test adding streams because I can't mock the data that goes into the `onChange` handler in the dropdown.
- added a component where a user sees dashboard streams which are not connected to the device anymore and gives him the option to remove them
- made components smaller

todo: 
- test with a running process
- add tests for redirecting to the data viewer of a process

src/redux/helper/objectMapper.js
DanielSeehausen: increasing security in terms of who can do what sounds good 👍 (instead of making more lambda functions/hambda/etc.)

src/utils/device.js
DanielSeehausen: Good question. I don't know if I see any advantages other than the syntactic sugar! In the end, I think even using `class` is misleading, because it implies we want to use instances in some manner, and we never do. 

I like the idea of namespacing here. A plain object seems the most natural way to do that. I am not sure I see the application of nesting `dashboard` in `DeviceUtils` (as in I am wondering if dashboard should be nested at all - with the alternative being just making use of the `DeviceUtils` api when methods on it are called), but I haven't dug into this nearly as much as you have. 

If you were so inclined (which I am not arguing for), JS now has static properties for classes, which could be assigned to this example `dashboard` and give you the same use result: 

```js
import dashboard from '...'

class DeviceUtils {
  static dashboard = dashboard
}
```

src/components/device/dashboard/Dashboard.jsx
DanielSeehausen: ah - that makes sense. If I am understanding the use of it in Mongoose, it is used to inject those objects/values associated with the ids into the object itself.

In our case, we aren't getting the device back with the dashboard stream objects then nested within. Instead, we are getting only the dashboard streams. Given the arguments, you might see how one would expect the return of `Device.populateStreams(device)` to be a device object itself.

src/components/device/dashboard/Dashboard.jsx
DanielSeehausen: oops my mistake. thank you for explaining to me!

src/components/device/dashboard/Dashboard.jsx
vdiezel: agreed!

PR #270, opened 2019-11-06
Issue 141 edit device name description

### LINKED ISSUE #141: add edit button for title and description of device
image

Per jan request: Please add a button that turns both the title and the description texts into text edit fields. On blur, the fields should post to the appropriate endpoint and update the device `title` and `description`. There may be device schema and/or nexus/backend endpoint work to fulfill this ticket! Investigate that first, and create separate tickets for it if so!

O-Susanoo:
Creating a separable component for editing header , the component now allow users to modify a text like this

image

Also added test for the new component to test the following behaviors : 
- check render of label
- check render on edit
- check call of the change name handler

Closes #141


package.json
```diff
@@ -23,6 +23,7 @@
     "react-beautiful-dnd": "^11.0.4",
     "react-copy-to-clipboard": "^5.0.1",
     "react-dom": "^16.8.6",
+    "react-jss": "^10.0.0",
```
O-Susanoo: Adding jss styling to project

src/components/device/DeviceWidget.jsx
```diff
@@ -111,14 +118,24 @@ const DeviceWidget = ({ device: { connectionState, isVirtual, name, state, runni
     fetchAndContinuouslyUpdateRunningData(runningProcessId, setRunningData)
   }
 
+  const renameDevice = async (newName) => {
+    await callUpdateDevice(id, { name: newName })
```
DanielSeehausen: something to think about in the future:

we will want a standard way of handling responses for the front end to display meaningful results to the user. Roughly something like this:

```js
const { res, err } = await callUpdateDevice(id, { name: newName })
if (err) flash(`renaming failed ${err.msg}`)
if (res.statusCode === 200) flash('success')
```

not to be implemented now, but something to noodle on as we will need it for the more restricted/trickier operation that renaming a device.

src/components/device/DeviceWidget.jsx
```diff
@@ -13,6 +13,20 @@ import bioreactorOfflineImage from "images/bioreactor_offline_cropped.png"
 
 import DurationDisplay from "components/misc/DurationDisplay"
 import LEDCircle from "components/device/spinners/LEDCircle"
+import EditableHeader, { EditableHeaderButtonClassName } from 'components/utility/EditableHeader'
+import { callUpdateDevice } from '../../redux/thunks/thunks'
+import { connect } from "react-redux"
+import injectSheet from 'react-jss'
```
DanielSeehausen: a non-blocking comment on the imports here:

we tend to organize the imports into (generally) two groups:
```
import packageA from 'node_modules_package_A'
import packageB from 'node_modules_package_B'

import oneOfOurFilesA from 'components/componentA'
import oneOfOurFilesB from 'components/componentB'
```

(this would put `connect` and `injectSheet` up top)

On a related note - we also have paths set up to be able to import stuff from `src` using absolute paths. Take a look at how `DurationDisplay` is imported from `src` with just `components/misc/DurationDisplay`. This allows us to avoid relative imports, making moving files less brittle!

In this case, only the `callUpdateDevice` import would change:
`...from 'redux/thunks/thunks`

DanielSeehausen: (you'll see its not consistent everywhere because we are taking the same approach of - if you touch the code, take the opportunity to refactor it)

src/components/processviewer/body/builder/PhaseTabContent.jsx
```diff
@@ -1,64 +1,44 @@
...
 import { callUpdateProcess } from 'redux/thunks/thunks'
+import EditableHeader from 'components/utility/EditableHeader'
+import { Icon, Button } from "semantic-ui-react"
 
 const mapStateToProps = state => ({
   displayedProcessPhase: state.displayedProcessPhase
 })
 
 const mapDispatchToProps = dispatch => {
   return {
```
DanielSeehausen: opportunity here to remove the explicit return, and make it implicit like `mapStateToProps`!

src/components/processviewer/body/builder/PhaseTabContent.jsx
```diff
@@ -91,54 +71,36 @@ class PhaseTabContent extends React.Component {
...
     return (
       <div>
-        <Form>
-          {editName ?
-            <Form.Field
-              control="input"
-              autoFocus
-              defaultValue={name}
-              onBlur={this.submitNewPhaseName.bind(this)}
-              onKeyDown={this.submitNewPhaseNameEnter.bind(this)}
```
DanielSeehausen: :laser: 👍

test/components/utility/EditableHeader.test.js
```diff
@@ -0,0 +1,42 @@
...
+            const editBtn = getAllByRole('button')[1];
+            fireEvent.click(editBtn)
+            const inputField = getByRole('textbox')
+            fireEvent.change(inputField, { target: { value: 'Test Title' } })
+            fireEvent.blur(inputField)
+            expect(mockChangeNameHandler.mock.calls.length).toBe(1)
+            expect(mockChangeNameHandler.mock.results[0].value).toBe(changeText);
+        })
+    })
+})
```
DanielSeehausen: git warning us to add newlines on end of every file:

https://evanhahn.com/newline-necessary-at-the-end-of-javascript-files/

https://stackoverflow.com/questions/44704968/visual-studio-code-insert-new-line-at-the-end-of-files

PR #284, opened 2019-11-12
GitHub api module

O-Susanoo:
Add Github API interface


src/utils/githubApi.ts
```diff
@@ -0,0 +1,40 @@
+
+const API_KEY = '18a65dad407c4676f0c82a6125ba2269ba7e5157'
```
DanielSeehausen: this solution looks great
there are two things here we may want to do regarding the api key:
- decouple the authentication for these API calls from any single developer. The first thing that comes to mind is to create an ospin github user and generate they keys for this user, but there may be other solutions!
- hide this api key in a `.env` file! Surprising we have gotten this far without one, but now may be a great time to make it. 
[Here is the short and sweet](https://medium.com/codait/environment-variables-or-keeping-your-secrets-secret-in-a-node-js-app-99019dfff716) 
[and here is the very useful and ubiquitous package they use](https://www.npmjs.com/package/dotenv)

test/mocks/gitHubPRMock.ts
```diff
@@ -0,0 +1,770 @@
+export const gitHubMockData =
```
DanielSeehausen: this file is marked `.ts`, but I don't see any typescript in it! does that matter?

test/utils/github.test.ts
```diff
@@ -0,0 +1,14 @@
...
+describe('github api', () => {
+
+  describe('test github data parsing', () => {
+    it('return array of GitHubPR', async () => {
+      const result = await parseGitData(gitHubMockData)
+      expect(result.length).toBe(1)
+    })
+  })
+
+})
```
DanielSeehausen: this is looking good

not blocking, but in the future, if we only have one test, we want it to be as broad as possible

jest is pretty good at mocking responses from fetches. In this case, we could mock the fetch response from `'https://api.github.com/repos/ospin-web-dev/REACTor/pulls?state=closed'` with the github mock data, then we would get a full coverage test of `getChangeLog`

test/mocks/gitHubPRMock.ts
O-Susanoo: No it doesn't really matter but , since it was a typescript feature i wanted everything to be typescript

Thank for the review

O-Susanoo:
Move github code from frontend to backend

PR #288, opened 2019-11-14
Display Process Run Times on Process Tiles

### LINKED ISSUE #262: FU Feature Request
Wants a time displayed for processes, similar to devices:

"
Ich finde die Anzeige mit der aktuellen Laufzeit im „Devices“-Abschnitt sehr schön! Kann man das Ganze auch im Interface unter „My Processes“ für bereits abgeschlossene Prozesse mit integrieren (also wie aktuell: Name; Datum und zusätzlich Gesamtlaufzeit)?
"

[Please move to in progress then complete once complete](https://trello.com/c/SMglAZfQ/39-process-time-of-processes-in-the-process-view)

DanielSeehausen:
## Process Tile UI Addition

This feature is for our friends over at Freie Universität Berlin!

image

##### Process tiles will now exhibit the following behavior
- **running/paused** processes display:
  -  running duration
  -  **started at day/time
- **executable/finished** processes display: 
  - **creation day/time

\private
closes #262 
also addresses several very minor visual bugs 
removes a stray console statement

maybe we should move the time to the left to be in line with the other meta data.
@vdiezel do you know an easy way to do that?

future work should include 'finished on' for complete processes - but we need to request firmware work to do that I believe.


src/styles/process-tile.css
```diff
@@ -0,0 +1,8 @@
+.process-meta-data: {
+  margin: 0;
+  padding: 0;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap
+}
+
```
vdiezel: We probably want to use `jss` now that we have it!

src/styles/semanticoverrides.css
```diff
@@ -212,6 +212,10 @@ body {
   color: #0E6EB8;
 }
 
+.break-word {
+  overflow-wrap: break-word
```
vdiezel: Didn't know that existed!

DanielSeehausen:
im going to make both changes now @vdiezel :
styling left
jss

DanielSeehausen:
image

Now floating left and style sheet no longer used (now using jss style)

the DurationDisplay wrapper component did not get jss styling because it would require overhauling the DurationDisplay component styling itself, and then changing how we use classes vs. injecting the styling directly. I don't think it makes sense to blow out this Pr and change all that right now.

PR #289, opened 2019-11-14
CI/CD integration for circleCI to publish changelogs 

vdiezel:
Ready to be reviewed!

image


.circleci/config.yml
```diff
@@ -60,12 +60,29 @@ jobs:
...
+      - run:
+          name: install aws-sdk
+          command: npm i aws-sdk
+      - run:
+          name: install node-fetch
+          command: npm i node-fetch
+      - run:
+          name: execute publish changelog script
+          command: node .circleci/publishChanges.js $GITHUB_API_KEY $CIRCLE_JOB $OSPIN_AWS_REGION
+
```
ospin-webapp: new circleCI job that is executed after deployment

.circleci/publishChanges.js
```diff
@@ -0,0 +1,49 @@
...
+    InvocationType: 'Event',
+    Payload: JSON.stringify(requestData)
+  }
+
+  lambda.invoke(params, function(err, data) {
+    if (err) console.log(err, err.stack)
+    else console.log(data)
+  })
+})
+
```
vdiezel: NodeJS script that is executed in the publishing job. It fetches the latest PRs, gets the first one, checks if it has a `changelog` label on it. If it has, uses the aws-sdk to invoke the lambda function directly (so it's not an HTTP request!) that inserts a changelog notification to every user on `dev`. We also pass the url as `originId` to the notification and the PR title as the message.

.circleci/publishChanges.js
```diff
@@ -0,0 +1,48 @@
...
```
vdiezel: NodeJS Script: 
- fetch PRs from GitHub
- Check if the last PR has a changelog label
- if so, invoke a lambda function that puts a changelog notification into every user on `dev` - due to direct invocation we have to create the event that is passed to the lambda function `user-dev-patchAllNotifications` ourselves 

originId for the notification is a pr-url and the message is the PR title!

vdiezel: we are passing down environmental vars using `process.argv` here. The variables themselves are stored in the circleCI project, in settings / environmental variables

.circleci/publishChanges.js
```diff
@@ -0,0 +1,48 @@
+const fetch = require('node-fetch')
+const AWS = require('aws-sdk')
+
+const url = `https://api.github.com/repos/ospin-web-dev/REACTor/pulls?state=closed`
```
DanielSeehausen: url doesn't communicate too much
(not blocking, but for future reference, perhaps something like `CLOSED_PRS` would let the reader know what the endpoint is when it is run on line 15

.circleci/publishChanges.js
```diff
@@ -0,0 +1,48 @@
...
+
+  const params = {
+    FunctionName: 'user-dev-patchAllNotifications',
+    InvocationType: 'Event',
+    Payload: JSON.stringify(requestData)
+  }
+
+  lambda.invoke(params, function(err, data) {
+    if (err) console.log(err, err.stack)
+    else console.log(data)
```
DanielSeehausen: I see this in their docs as well

would we be able to switch console.log to console.error if it fails? I'd like this to be as loud as possible if it fails!

.circleci/publishChanges.js
vdiezel: done!

PR #299, opened 2019-11-19
fixing missing device subscriptions when entering an already loaded d…

### LINKED ISSUE #298: When going from devices view to the dashboard, data is not being populated in the widgets
Data is working however when you refresh WITHIN the dashboard view it doesnt load the process data

vdiezel:
…evice

closes #298


src/redux/thunks/thunks.js
```diff
@@ -426,57 +426,57 @@ export function callLoadDevice(loadedDevice, deviceId, userId, processesInitiali
       else {
         const deviceData = await nexus.device.get(deviceId)
         device = createDeviceObject(deviceData)
+      }
```
vdiezel: I had the subscriptions within the `else` case, which happened when a user tries to open a device that is not loaded into application state. This case was for the only reason of loading the right device when the user refreshes the browser window within a device view. Somehow I also put all subscription there too so our device was only subscribed when we actually refreshed the browser within the device view. This also caused the dashboard view not to work, because we never subscribed to the datapoints topic! I wonder how it didn't break the whole application ... :fire_engine:

DanielSeehausen: I believe this also fixes some other updating bugs! nice!

PR #306, opened 2019-11-20
Test circle ci contexts

vdiezel:
pure test purpose - no review required!


vdiezel:
was for tetsing purposes only.

PR #308, opened 2019-11-21
Fix for real time changes to running processes on phase transitions

### LINKED ISSUE #290: Process synchronization when ad hoc changes are staged but the device publishes a phase transition
relates to #283

vdiezel:
To prevent synchronization errors between the device and the web application for running processes which have unapplied real time changes when a device is transitioning into a new phase, unapplied changes will now be automatically discarded on phase transitions. Manual transitions will warn the user when unapplied changes are recognized.

image

\private 
closes #290

On phase transition checks in the reducer handler if there are unapplied changes to a process first, and if so,  restore the original process description from the snapshot. Displays a warning for manual transitions when unapplied changes are recognized - for incoming time based transitions there is no warning implemented yet. The best way would be to make use of the implementation for #211


vdiezel:
Don't know what happened here, somehow got two old commits in that branch that I created this morning...

vdiezel:
replaced with #309

PR #312, opened 2019-11-21
update selected phase option to find the correct option

### LINKED ISSUE #260: FU Bug - Processes not displaying once started
FU Bug:

Beim Starten dieser Prozesse gelange ich in ein Fenster, was vollkommen weiß ist. Der Reaktor läuft zwar an, aber man kann ihn nicht steuern/sieht keine Daten. Man muss in diesem Fall den Reaktor komplett ausschalten.

believe it is one of these processes:

* 191028_Adjusted PID Heater/Casing/Plate #4 - Mon Oct 28 2019 13:32:41
* 191028_Equilibration - Mon Oct 28 2019 17:44:40

image

[move this to in progress and close it once finished plz](https://trello.com/c/G2y9TjMG/37-ui-blank-when-starting-a-process)

I suggest trying to clone the processes, then starting them, or something to try to recreate behavior. We can also look at the lambda error logs, as the timestamps are provided

EDIT: Check process data in the data base to not malformed somehow (id, progression)

DanielSeehausen:
#### Process View Bug Fix
Rarely, attempting to view a process that had been modified resulted in a blank screen.
Thanks to our friends at FU, we were able to identify and fix this bug. Processes should no longer (ever) be blank when clicked on.
\private
closes #260 
also tacked on the nexus ref update (eslint stuff, so it didn't really belong to any reactor PR)

this bug was happening because we were using a phaseId to look up a phase by index in an array that has serialized a process.progression object. The index did not correspond to the phaseId

this led to a return of `undefined`, and ultimately `undefined.text` call that blew everything up

now, we look up the phase directly on the process, instead of in the serialized array

(separately, this changes the component's references of `selectedPhase` to `selectedPhaseId` to reduce ambiguity)


src/components/processviewer/body/dataviewer/Stream.jsx
```diff
@@ -248,9 +249,12 @@ class Stream extends React.Component {
           </Button.Group>
           {' '}
           <Button.Group compact>
-            <Dropdown trigger={<Button
-                positive={mode === 'phase'}
-              >{phaseOptions[selectedPhase].text}</Button>}
+            <Dropdown
+              trigger={
+                <Button positive={mode === 'phase'}>
+                  {selectedPhaseName}
```
DanielSeehausen: this is now just a string
everything else in the return method is just formatting

src/components/processviewer/body/dataviewer/Stream.jsx
```diff
@@ -221,6 +221,7 @@ class Stream extends React.Component {
 
     const noData = streamData.length === 0
 
+    const selectedPhaseName = activeProcess.description[selectedPhaseId].name
```
DanielSeehausen: this line is the whole PR, and the value is now used in the render method in place of the `undefined` index lookup

PR #317, opened 2019-11-22
set proper zoom boundaries when zoom is triggered but no data reload …

### LINKED ISSUE #283: Process Data View
```diff
! Whoever picks this up, please point it yourself and remove the 'triage required' once done!
```

**Describe the bug**
I was viewing OD-006 Berlin Bioreactor process data while the process ''PIDBestimmung - 70 rpm, 90 rpm, 100 rpm'' was still running on its last phase (prod environment). I clicked on ''process'' to view the whole process data. When I click and highlight a portion of the data to zoom in on the part, the zooming in works initially, but after about 1 - 5 seconds the graph automatically jumps back to a fully zoomed out view. 
It seems it can happpen on any graph in the data, at any point on the graph. I tried to discover a pattern to it (like if it only happens when I zoom in on certain time frames or certain phases), but there doesn't seem to be one. 
To see if this was an isolated event, I viewed data on a running process on OD-001 Athen (dev environment). The same behavior occurred. This process has also been running for quite some time, but only has one phase.
I then ended the process on the Berlin Bioreactor and started a new one. I viewed the data while the process was running just like before. There was no issue. I added a new phase and started the new phase and there was still no issue. After letting the process run for almost an hour, I experienced the same issues. (Maybe the issue only occurs after processes have been running for some time?)

**To Reproduce**
1. Run a process (at least half an hour to an hour).
2. After some time view process data (not live data) while the process is still running.
3. Click and highlight an area of the graph to zoom in.
4. See it jump back to zoomed out. May have to repeat. Occasionally it stays zoomed in.

**Expected behavior**
The graph should show a zoomed in view of the highlighted area and shouldn't zoom out unless the user causes it to do so by clicking one of the refresh buttons.

**Screenshots**
Zoomed out view of process data
https://user-images.githubusercontent.com/57398264/68668458-795acf00-0548-11ea-98a5-572a7d696cb1.png

Highlighting an area (the zoom in will occur and then it will zoom back out)
https://user-images.githubusercontent.com/57398264/68670821-19ffbd80-054e-11ea-8ed7-fa3a3e970269.png

**Environment**
Both.
 - [ ] Development (https://dev.ospin-app.com)
 - [ ] Production (https://ospin-app.com)

Berlin
46a5504b-0438-4757-a1d9-06ae0e73fbec / 0545cef9-46b8-48c7-aa93-5da474f84baf
Athen
a8207dab-43ab-44be-93eb-308a60df1f29 / de8a8d3d-63a1-48d1-adf3-5a7268d246be

**Additional context**

vdiezel:
closes #283 

Bug reported by our big supporter Jared. Zoom viewport was not set properly when we zoomed in into a grpah but no reload of data required causing viewports to reset after zoom!


src/components/processviewer/body/dataviewer/Stream.jsx
```diff
@@ -162,6 +161,9 @@ class Stream extends React.Component {
         await this.loadHistoricData(this.props.activeProcess.id, this.props.stream.id, this.CONSTANTS.RELOAD_COUNT, start, end)
         this.setState({zoomParams, loadingData: false})
       }
+      else {
```
DanielSeehausen: eslint demands `} else {`

src/components/processviewer/body/dataviewer/Stream.jsx
```diff
@@ -162,6 +161,9 @@ class Stream extends React.Component {
...
+        this.setState({ zoomParams })
```
DanielSeehausen: this was throwing an eslint error, but its ridiculous and shouldn't. I updated the eslint config to allow this in a commit

.eslintrc.js
```diff
@@ -12,5 +12,6 @@ module.exports = {
       "named": "ignore",
       "asyncArrow": "always"
     }],
+    "object-curly-newline": ["error", {"multiline": true}],
```
DanielSeehausen: https://eslint.org/docs/rules/object-curly-newline

ctrl+f `Examples of correct code for this rule with the { "multiline": true } option`
to see what this enables

PR #327, opened 2019-11-25
Issue 307 fix resizing in modals

### LINKED ISSUE #307: Fix resizing in modals to contain the messages and button within the modal
image

It's also visible when editing committing changes to phase

image

UPDATE: This needs to be done for all modals!

O-Susanoo:
Closes #307


DanielSeehausen:
will look again once the second go around is added!

O-Susanoo:
Ok so i have made the change , i have also fixed those who don't seem to have that issue but the moment you have an unpredictable error message you will surprised how the modal will turn out to be 

I have a note about the dev do you guys think it should be better to have a class with those properties and instead of styling the component i can just link it to a class , if so which file do i need to put the class definition in it (semanticoverrides.css) ? if it's not a problem then i think we are done with this ticket

src/components/device/setup/DeletePopup.jsx
```diff
@@ -7,6 +7,11 @@ class DeletePopup extends React.Component {
     showPopup: false,
   }
 
+  confirmClick = () => {
+    this.props.confirmHandler()
```
DanielSeehausen: believe this is throwing an eslint error!

O-Susanoo: no no i would like to be clean however according to it i have this error
confirmHandler' is missing in props validation eslint

i think it want an the prope typs package ?

O-Susanoo:
I will change the name you are right handleClick is more logical , i have already checked on all forms in my localhost but of course once deployed i will double check that it's the rule

vdiezel:
> Ok so i have made the change , i have also fixed those who don't seem to have that issue but the moment you have an unpredictable error message you will surprised how the modal will turn out to be
> 
> I have a note about the dev do you guys think it should be better to have a class with those properties and instead of styling the component i can just link it to a class , if so which file do i need to put the class definition in it (semanticoverrides.css) ? if it's not a problem then i think we are done with this ticket

It not a problem - but in case you want to use a class, it could either use the semanticoverrides (which slowly becomes just a place where everybody dumps his css) or create a new css and import it (I know that one can integrate bundling for css files so you only need to import one css file that contains all css code) OR use JSS and create a reusable style that you import into every modal

O-Susanoo:
I guess that one day we should go throw all the code of the ui and standardize the css part , for now i will leave it like this , because i think if i have to do it then it will be using jss.

Thank you for the note

PR #336, opened 2019-11-26
Patch for process tile duration

DanielSeehausen:
when process tiles have a duration into the days, the display got super bad. I am not saying its the best, but its much better and it didn't take time to do. It doesn't change how it displays when under 1 day. We can find a more polished display later:

# then

image

# now

image

# now without days

image


.eslintrc.js
```diff
@@ -16,6 +16,7 @@ module.exports = {
     }],
     "padded-blocks": "off",
     "quotes": ["warn", "single", { "avoidEscape": true }],
+    "react/prop-types": "off",
```
DanielSeehausen: we aren't using prop types to my knowledge anywhere right now

src/components/device/processes/TileDurationDisplay.jsx
```diff
@@ -4,14 +4,15 @@ import DurationDisplay from 'components/misc/DurationDisplay'
...
   justifyContent: undefined,
+  display: 'block',
   alignItems: undefined,
   width: undefined,
   padding: '9px 0px',
   float: 'left'
 }
 
-const ProcessTileDurationDisplay = ({ ms }) => (
+const TileDurationDisplay = ({ ms }) => (
```
DanielSeehausen: name did not match the file

.eslintrc.js
vdiezel: that's correct!

PR #337, opened 2019-11-27
Bugfix: Disabled UI controls for running phase modifications

### LINKED ISSUE #328: Fix Enable Phase Editing in current running process
It seem to me that somehow we cache the state of editing of a device and only when a user refresh a page we bring the new state and thus allowing us to edit the device parameters

image

vdiezel:
Occasionally, running phases would present a disabled UI control in the process builder which prevented the user from modifying them. This bug now has been fixed.
\private 
closes #328 

We had two issues: 
  1. We were always creating duplicate subscriptions on `callLoadDevices` 
  2. A really weird bug in `callLoadDevices` I ran into already a couple of times: We are subscribing to five different topic, but we were not awaiting any of those. This caused this crazy behaviour: The callbacks on every topic were executed 1 time for every subscription we had (so 5 times) when a message arrived for that topic; e.g. when a new phase arrived from the device in the UI on the topic `phases`, because we had 5 total subscriptions ( `phases` and four other ones), the callback for adding a phase was executed 5 times in total, which led to 5 phases entries in the progression array and this confused the UI about the process history. Now the fun part: I added await to the topic subscriptions and the issue was gone. But in fact, adding await just to ONE of those topics also resolves the issue (no idea why). For consistency and because it's not really a performance issue I use awaits in all subscriptions.


src/components/devices/Devices.jsx
```diff
@@ -15,7 +15,7 @@ const mapStateToProps = state => ({
 })
 
 const mapDispatchToProps = dispatch => ({
-  callGetDevices: () => dispatch(callGetDevices()),
+  callGetDevices: (cachedDevices) => dispatch(callGetDevices(cachedDevices)),
```
vdiezel: passing cachedDevices to prevent duplicate subscriptions in the thunk

src/redux/thunks/thunks.js
```diff
@@ -473,7 +473,7 @@ export function callLoadDevice(loadedDevice, deviceId, userId, processesInitiali
       }
 
       if (device.dataSubscription === null){
-        device.dataSubscription = nexus.device.subscribe(`dataPoints/${device.id}`, message => {
+        device.dataSubscription = await nexus.device.subscribe(`dataPoints/${device.id}`, message => {
           const data = message.value
           dispatch(addLiveStreamData({data}))
         })
```
vdiezel: weird `await` issue described in the PR description

vdiezel:
> looks good! are there any cases off the top of your head where behavior will change for the worse when we aren't refetching those devices that have been cached? (i.e., could something change between the time it was originally cached, and the time we fetch again, where we are actually relying on the updated device info that would come from refetching it?)
> 
> everything looks good - this is the only thing that jumps out without having looked into the code

This may be the case when we update the the device remotely in anyway but do not apply the changes locally. Currently this never happens, but I agree that caching always add complexity. I#ve been evaluating also other soutions. 
1. One would be simply check only the subscriptions and move them from the cached device to the newly loaded version of it 
2. Unsubscribe from existing subs before loading the devices

I'll prefer 1. for now. I'll fiddle around with it.

O-Susanoo:
It was a tricky one !! nice one

vdiezel:
I changed the logic a to be less scary to sync issues between locally and remotely stored device changes: We now only take over existing subscription from devices, rest of the device data is not cached locally.

PR #349, opened 2019-12-03
resets "success" message in controller modal, resizes the the message…

### LINKED ISSUE #344: take away "save as default" message in controller modal after changing values again and adjust message width

vdiezel:
… and reformatted according to eslint

closes #344


.eslintrc.js
```diff
@@ -17,7 +24,10 @@ module.exports = {
     "padded-blocks": "off",
     "quotes": ["warn", "single", { "avoidEscape": true }],
     "jsx-quotes": ["warn", "prefer-single"],
+    "react/destructuring-assignment": ["warn"],
+    "react/jsx-tag-spacing": "off",
     "react/prop-types": "off",
+    "react/state-in-constructor": "off",
     "semi": ["warn", "never"],
     "space-before-function-paren": ["warn", {
       "anonymous": "always",
```
vdiezel: I made a couple of rules be more liberal because it has become very difficult to tell syntax errors from formatting issues. I think errors should be only where formatting impacts readability or code structure significantly.

src/components/processviewer/body/builder/ControllerModal.jsx
```diff
@@ -15,109 +18,128 @@ class ControllerModal extends React.Component {
...
     isSuccess: false,
-    errorMessage: ""
+    errorMessage: '',
   }
 
   close() {
     this.props.closeHandler()
   }
 
+  resetSuccessMessage () {
```
vdiezel: this is the the only real change in the PR: this handler is passed down to the controller streams and allows the child components to reset the `success` message when new values are entered into the inpout fields.

src/components/processviewer/body/builder/ControllerModal.jsx
```diff
@@ -15,109 +18,128 @@ class ControllerModal extends React.Component {
...
-            style={{ marginBottom: '15px', paddingBottom: '15px'}}
+            <Form
+              error={isError}
+              success={isSuccess}
+              style={{ marginBottom: '15px', paddingBottom: '15px' }}
             >
-              {Object.keys(controllerStreamGroups).map((controllerName, i) => {
-                return (<div key={i}>
+              {Object.keys(controllerStreamGroups).map((controllerName) => (
+                <div key={controllerName}>
```
vdiezel: Found out that using indices as keys in React Components is a bad thing (thanks eslint). Example: You have 10 elements, you delete the 5th, when rendering again React will think that the 10th element has been deleted. This way React can't apply it's optimization patterns.

src/components/processviewer/body/builder/ControllerModal.jsx
```diff
@@ -15,109 +18,128 @@ class ControllerModal extends React.Component {
...
-                </div>)
-              })}
+                </div>
+              ))}
               <Message
                 error
                 header='Something went wrong.'
                 content={errorMessage}
-                icon="warning circle"
-                className="max-width-498"
```
vdiezel: fixes width issues

.eslintrc.js
```diff
@@ -4,10 +4,17 @@ module.exports = {
...
+    "brace-style": "off",
     "comma-dangle": ["error", "always-multiline"],
     "consistent-return": "off",
+    "prefer-template": ["warn"],
     "indent": ["error", 2, { "SwitchCase": 1 }],
+    "jsx-quotes": ["warn", "prefer-single"],
+    "jsx-a11y/label-has-associated-control": ["warn"],
     "linebreak-style": ["error", "unix"],
+    "no-restricted-globals": ["warn"],
+    "no-shadow": ["warn"],
```
DanielSeehausen: i'd vote this one to be on error - if a reader (understandable) associates a variable incorrectly because it is declared twice, that can lead to a pretty serious error!

DanielSeehausen: same with `no-restricted-globals`

.eslintrc.js
```diff
@@ -4,10 +4,17 @@ module.exports = {
   "parser": "babel-eslint",
   "rules": {
     "array-bracket-spacing": "off",
+    "arrow-parens": ["warn", "as-needed"],
+    "brace-style": "off",
```
DanielSeehausen: this is a pretty big one to just turn off

with brace style, we should pick something and stay consistent, which is the most important. I believe enabling variation in brace style for new code would be irresponsible :/

.eslintrc.js
```diff
@@ -4,10 +4,17 @@ module.exports = {
   "parser": "babel-eslint",
   "rules": {
     "array-bracket-spacing": "off",
+    "arrow-parens": ["warn", "as-needed"],
```
DanielSeehausen: 👍

.eslintrc.js
```diff
@@ -17,7 +24,10 @@ module.exports = {
     "padded-blocks": "off",
     "quotes": ["warn", "single", { "avoidEscape": true }],
     "jsx-quotes": ["warn", "prefer-single"],
+    "react/destructuring-assignment": ["warn"],
+    "react/jsx-tag-spacing": "off",
```
DanielSeehausen: any chance we could set on warn? these are pretty standard and definitely helpful because JSX/HTML has crap readability. 
Unless you have something specific in mind they are throwing for which is ridiculous (even then, i'd prefer to address the rule). They help us avoid things like this:
`< Hello></ Hello>`, `<Provider>< /Provider>`
and maintain things like this:
`<Hello firstName="John" />` (which aid the eye when reading down jsx)

src/components/processviewer/body/builder/ControllerModal.jsx
O-Susanoo: Yeah i knew that a while ago but sometimes if you push data that is not uniquely identified , you are stuck with the indices

.eslintrc.js
DanielSeehausen: after looking at the actual code in the PR with eslint, am I right to assume you took this off because of it throwing on the shadowing of the imported redux action?

https://stackoverflow.com/questions/37682705/avoid-no-shadow-eslint-error-with-mapdispatchtoprops

would any of these be reasonable solutions? No.2 looks good (even makes sense), but I defer to you.

vdiezel: I agree on the no-restricted-globals. I deactivated it because eslint was complaining about `isNaN`. I replaced it with `Number.isNaN()`. The `no-shadow` rule blows up on when you desctructure props and functions from `mapDispatchToProps` because it can't tell the difference between the thunks that were imported and the dispatch function that ends up in the props via `connect`:

https://stackoverflow.com/questions/37682705/avoid-no-shadow-eslint-error-with-mapdispatchtoprops

For now, I just let the error be there in the file.

.eslintrc.js
DanielSeehausen: (brace-style I mean. arrow-parens looks 👍 )

vdiezel: Ok - I talked with Mo and we are actually more the `stroustrup` kind of people but I decided to use the `one true` brace style

https://eslint.org/docs/rules/brace-style

that is predefined in the air-bnb configration. The reason is this: I think we should use our own eslint file only to soften up the predefined rules of air-bnb where we consider them to be too strict. This way we will not run into the problem, that after making a file pass linting, we will need to touch it again because rules changed in our `.eslintrc` file.  You and Mo prefer one consistent style for braces, I thought it doesn't really matter. The best solution in my view is therefore to stay with the `one true` brace style.

vdiezel:
I adjusted the eslint rules after the discussion.

PR #350, opened 2019-12-03
ADD Filter functionality to device + Enable Process filtration onChange

### LINKED ISSUE #87: Devices - add filter funtionality to provided UI
When a user enters values into the searchbar, devices should be filtered based on chosen property

O-Susanoo:
Closes #87


src/components/devices/DeviceList.jsx
```diff
@@ -11,18 +11,24 @@ const renderDeviceWidget = (device, clickHandler, idx) => (
   />
 )
 
-const DeviceList = props => (
+const DeviceList = (props) => {
+  const { filterQuery, devices, noDevices, deviceClickHandler } = props
+  const queryDeviceByName = (filterQuery !== undefined && filterQuery !== '') ? (name) => new RegExp(filterQuery, 'i').test(name) : () => true
```
vdiezel: like this filter functio

O-Susanoo: I thought about it actually but it's one line function , so i keep it there

PR #354, opened 2019-12-04
Check process is found before phase switch

### LINKED ISSUE #340: waiting on the process view ends up throwing an error - see screenshot
after a certain amount of time, just sitting on a non-running process screen throws this error. I have never seen it occur directly - only when I have left the tab and come back to it do I see this. Some investigation re: how long it actually takes and confirmation that's the screen that is the culprit is required!

image

O-Susanoo:
Closes #340

It seem that when user open two tabs and start a new process the first subscribed tab will not see immediately the new process but yet still try to find the diff to it's latest state.

The risk of the fix we made is that it can destroy unstaged changes to a device , we will be working on this error on separate ticket


src/redux/reducers/reducer.js
```diff
@@ -212,14 +212,18 @@ function rootReducer(state = initialState, action) {
 
     const { processes, processDescriptionSnapshots } = state
     const { processId, phase } = action.payload
+    const process = processes.find(p => p.id === processId)
```
DanielSeehausen: we have a function to do just this (and the file is already being imported in `reducer`), just not the function
see `getProcessById` in `utils/process`

While it may seem unnecessary in terms of lines of code, we want to stick to using and growing a standardized interface into the `process` model (and, in a more general sense, into our model layer in general). This helps separate our controller and model layers!

src/redux/reducers/reducer.js
```diff
@@ -212,14 +212,18 @@ function rootReducer(state = initialState, action) {
...
+
+    if (process === undefined){
```
DanielSeehausen: don't forget to check eslint!

src/redux/reducers/reducer.js
O-Susanoo: Will do even tough this is supposed to be a very simple fix , i will peer with viktor on this one because it's more than meets the eye

src/redux/reducers/reducer.js
O-Susanoo: this entire file is eslint bomb so i conserve stuff as they are but i will try to do the efforts to fix only this part

O-Susanoo:
Check out the new pr text

PR #358, opened 2019-12-04
357 validation errors on adhoc

### LINKED ISSUE #357: validation errors not popping up when ad hoc changes don't pass validation

vdiezel:
Had wrong reference in the ChangesFooter for the action creator to add validation errors to the store.

closes #357 

Second commit is for eslint changes and formatting.


vdiezel:
Tried to provide some tests for this one - turned out to be no reasonable. Semantic UIs Accordion poses the same challenge as their Dropdown: One cannot fire events to trigger them using React Testing Library. Therefore I can't unfold the inptut fields for downstreams t enter any vaues...

src/components/device/Device.jsx
```diff
@@ -52,56 +77,51 @@ class Device extends React.Component {
...
 
     if (activeProcess) {
-      return <ProcessScreen {...{ activeProcess, activeDevice }} />
-    } else if (activeDevice) {
+      return <ProcessViewer activeProcess={activeProcess} activeDevice={activeDevice}/>
+    }
+    if (activeDevice) {
       console.warn(`process: ${this.props.match.params.processId} not found! This should not be seen! Alert an admin!`)
       this.props.history.replace(`/device/${activeDevice.id}/processes`)
       return null
```
DanielSeehausen: I just added a commit on top to move this `return null` out of the conditional. I was the last person to touch it on blame so I didn't want to bother requesting it. Since this is a render, wanted to keep with best practice of returning `null` not `undefined` in the unlikely case that neither `if` conditionals are met

DanielSeehausen: I read up on it:
in the past, react complained when `undefined` was rendered. As I understand, this has since been patched, and it now takes `false`, `true`, `undefined`, `null` without issue

This also answers the question: "why prefer ternary over short circuit" (seems there is no reason to prefer one over the other now, since the change)

I still see `null` being recommended as best practice, but good to know!

PR #368, opened 2019-12-06
Convert Error to string so that it pass backend check

### LINKED ISSUE #361: Call to logClientError fails validation
bug_3

O-Susanoo:
Closes #361


src/components/utility/ErrorWrapperProvider.jsx
```diff
@@ -5,11 +5,12 @@ import * as nexus from 'nexus'
 class ErrorWrapperProvider extends React.Component {
 
   componentDidCatch(error, errorInfo) {
-    nexus.utils.logClientError(error, errorInfo.componentStack)
+    const errString = error !== undefined ? error.toString() : 'Undefined error'
+    nexus.utils.logClientError(errString, errorInfo.componentStack)
     this.props.history.push({
```
DanielSeehausen: looks great - don't neglect the blank lines. They are useful to group code. (i.e. one might make sense on line 9.5 - separating the post of error data from the react behavior). Not worth making the change, just wanted to point it out because I notice it a lot. Even your man [Stroustrup](https://regmedia.co.uk/2015/04/27/bjarne_stroustrup.jpg?x=442&y=293&crop=1) is [all about that](http://www.stroustrup.com/3rd_code.html)

O-Susanoo: Thanks for the observation i will start doing that for sure

PR #370, opened 2019-12-06
adapt to the new error response API

### LINKED ISSUE #367: error messages coming from the server are not displayed as they should anymore.
Example for a failed validation: It was a 422 because the comment field was emty ( the joi schema was too strict here,  ut this was fixed). The user should see the error message `validation failed: data.comment not allowed to be empty`

image (1)

vdiezel:
Changes the access of the error message in `logAndExtractError` in the thunks to fit the new error response API.

closes #367


src/redux/thunks/thunks.js
```diff
@@ -33,7 +33,7 @@ import {
 import { subscribeToDeviceTopics } from '../helper/subscribe'
 
 const logAndExtractError = error => {
-  const errorMessage = error.response ? error.response.data : error
+  const errorMessage = error.response ? error.response.data.message : error
```
vdiezel: data is read from the response body ( after `JSON.parse` was caled on it ). `message` is the key i've chosen on the backend.

DanielSeehausen: just double checking `data` is guaranteed if `response` is present

PR #375, opened 2019-12-06
creates dataSeries only when there is phase data

### LINKED ISSUE #366: Legend in graph for a chosen phase shows also all phases which come after it
Bildschirmfoto von 2019-12-05 16-44-28

vdiezel:
When creating data series entries all phases were added to the display which started after the first one due to artificial data point insertion to prevent gaps. Fixed this.

closes #366 

first commit is the one with logic changes.


src/utils/process.js
```diff
@@ -427,35 +427,38 @@ export function generatePhaseSeriesData(process, datapoints, clickHandler = null
...
+          // unshift less efficient than reverse.push.reverse
 
-        slice.reverse()
-        slice.push(lastDataPoint)
-        slice.reverse()
+          slice.reverse()
+          slice.push(lastDataPoint)
+          slice.reverse()
+        }
       }
```
vdiezel: artificial data point insertion to prevent gaps between phases.

src/utils/process.js
```diff
@@ -427,35 +427,38 @@ export function generatePhaseSeriesData(process, datapoints, clickHandler = null
...
+      data.push({
+        click: clickHandler ? (e) => clickHandler(slice, e) : null,
+        markerSize: 0,
+        type: "line",
+        toolTipContent: "{x}: {y}",
+        showInLegend: true,
+        legendText:  phaseName,
+        dataPoints: slice
+      })
+    }
```
vdiezel: data entries now only pushed to to the display when data the phase is visible in the selection. Also restores our missed `No datapoints yet` in the graph view - completely forgot we had that!

src/utils/process.js
```diff
@@ -427,35 +427,38 @@ export function generatePhaseSeriesData(process, datapoints, clickHandler = null
     const start = getPhaseStart(process, (phase.phaseId).toString())
     const end = getPhaseStart(process, getNextPhaseId(process, (phase.phaseId).toString()))
 
+
     const phaseName = getPhaseName(process, (phase.phaseId).toString())
 
-    let slice = datapoints.filter(p => (p.x >= start && p.x <= end))
+    let slice = datapoints.filter(p => (p.x >= start && p.x < end))
```
O-Susanoo: why the last data point is not taking ?

vdiezel: I want to prevent overlaping datapoints because this would mean that two phases will have a common point. Example: You have phase 1, phase 2 and phase 3. The data slice for phase 2 would also contain the last data point of phase 1 and the first of phase 3. In the legend of the graph you would then see all three phases when you choose phase 2, because phase 1 and phase 3 also contribute each 1 datapoint.

DanielSeehausen: makes sense

PR #377, opened 2019-12-06
displays the default value of a controller value in the modal

### LINKED ISSUE #269: Display default values for controller parameters in the controller modal
requested by marcus from FU - I think it is a valid request

vdiezel:
closes #269 

Bildschirmfoto von 2019-12-07 00-17-33


src/components/processviewer/body/builder/ControllerStream.jsx
```diff
@@ -1,51 +1,61 @@
...
   return (
     <Form.Field error={target > stream.max || target < stream.min}>
       <Popup
-        trigger={<label>{stream.name}</label>}
+        trigger={
+          <label>
+            {stream.name}
+            <span style={{ color: 'grey', marginLeft: '8px' }}>
+              {`[ default: ${defaultValue}${unit} ]`}
+            </span>
```
vdiezel: adds the default value + unit to the label of the input

src/utils/device.js
```diff
@@ -96,6 +96,18 @@ export class Device {
...
+      .find(defaultEntry => defaultEntry.streamId === stream.id)
+    if (foundDefaultEntry) return foundDefaultEntry.value
+    return stream.default
+  }
+
+  static getStreamUnit(stream) {
+    if (stream.unit === '-') return ''
+    return stream.unit
+  }
+
```
vdiezel: two new helper functions to fetch a default value and the stream unit

src/utils/device.js
```diff
@@ -96,6 +96,18 @@ export class Device {
     return stream.name
   }
 
+  static getStreamDefault(device, stream) {
+    const foundDefaultEntry = device.streamDefaults
+      .find(defaultEntry => defaultEntry.streamId === stream.id)
+    if (foundDefaultEntry) return foundDefaultEntry.value
+    return stream.default
```
O-Susanoo: ```return foundDefaultEntry ?  foundDefaultEntry.value : stream.default```

saving typing energy :) !!

DanielSeehausen: 🕵

src/utils/device.js
DanielSeehausen: Understanding a stream should likely never be without a '-' or a unit, should we have a default return value instead of undefined if it is somehow missing?

src/utils/device.js
vdiezel: Yes, thanks. In fact I have to check explicitly that it is different from `undefined` because it might be set by the customer to `0`.

PR #380, opened 2019-12-08
211 flash messages

### LINKED ISSUE #211: As a developer, I have a component I can invoke that provides a flash banner across the top of the application

vdiezel:
Did this as a small project this weekend:

- Implements a wrapper component that displays flash messages in the bottom right corner using TypeScript
- features 4 different types of messages: error | warning | info | success
- animated entry/exit using semantic ui `Transition` wrapper with the `fly left` aninmation
- messages can be closed prematurely, and display duration is adjustable
- messages are stackable
- adds a `FlashMessenger` utility class that makes firing messages pretty easy
- implements 2 examples of the `FlashMessenger`
- provided test coverage

closes #211 

Bildschirmfoto von 2019-12-08 16-05-31

Bildschirmfoto von 2019-12-08 19-37-56

Bildschirmfoto von 2019-12-08 21-21-18


src/components/device/Device.jsx
```diff
@@ -102,7 +103,7 @@ class Device extends React.Component {
       return <ProcessViewer activeProcess={activeProcess} activeDevice={activeDevice}/>
     }
     if (activeDevice) {
-      console.warn(`process: ${this.props.match.params.processId} not found! This should not be seen! Alert an admin!`)
+      FlashMessenger.warning(`process: ${this.props.match.params.processId} not found!`)
```
vdiezel: fixes this as requested in the issue #211 - shows an example use case for the messenger

src/components/routing/Routes.jsx
```diff
@@ -8,11 +8,13 @@ import ResetPassword from '../authentication/ResetPassword'
 import PrivateRoute from "./PrivateRoute"
 import BoardConfigBuilder from "components/admin/board-config-builder/Main"
 import ErrorWrapperProvider from "components/utility/ErrorWrapperProvider";
+import FlashMessagesContainer from "components/utility/FlashMessagesContainer";
 
 const Routes = () => (
 
   <BrowserRouter forceRefresh={false}>
     <ErrorWrapperProvider>
+      <FlashMessagesContainer/>
```
vdiezel: New TypeScript component as a sibling to our `Routes` makes it possible to post on top of everything besides `ErrorWrapperProvider`.  We could also move it out of the error boundaries.

src/components/utility/FlashMessage.tsx
```diff
@@ -0,0 +1,162 @@
...
+    justifyContent: 'space-between',
+    alignItems: 'center',
+  },
+  messageIcon: {
+    marginRight: '16px !important',
+  },
+  closeButton: {
+    marginLeft: '16px !important',
+  }
+}
```
vdiezel: Styles moved to the bottom

O-Susanoo:
There is actually a lot of packages that do all of this stuff , the first one that cross my mind is [react-notifications](https://www.npmjs.com/package/react-notifications) and there are also others way cooler than this one , but if kompir is my food kryptonite than the same is for coding , i can't argue with someone who loves typescript , great work happy you enjoyed playing with it this week-end

src/components/device/Device.jsx
O-Susanoo: So now the user will see this messages , is that OK ?

src/components/utility/FlashMessage.tsx
```diff
@@ -0,0 +1,164 @@
...
+  message : Message,
+}
+
+type DispatchProps = {
+  dispatchDeleteFlashMessage: (messageId: number) => void
+}
+
+type IconType = 'check' | 'info' | 'warning' | 'ban'
+
+type MessageIcons = {
```
O-Susanoo: Awesome you are using indexers !!!

src/components/utility/FlashMessage.tsx
```diff
@@ -0,0 +1,164 @@
...
+export type Message = {
+  type: 'warning' | 'error' | 'info' | 'success'
+  text: string,
+  persistTime: number,
+  id: number,
+}
+
+type FlashMessageState = {
+  visible: boolean,
+  persistTimeout: any,
```
O-Susanoo: you can do like this persistTimeout : number | undefined , this way it will also accept undefined , and the reader will understand it's eventually will get a number value

test/components/utility/FlashMessagesContainer.test.js
```diff
@@ -0,0 +1,85 @@
+import React from 'react'
```
O-Susanoo: i guess it's better ti gave the ts extension for tests that are testing a typescript component

test/components/utility/FlashMessagesContainer.test.js
```diff
@@ -0,0 +1,85 @@
...
+import { Provider } from "react-redux"
+import configureMockStore from 'redux-mock-store'
+
+import FlashMessagesContainer from 'components/utility/FlashMessagesContainer.tsx'
+import { initialState } from 'redux/initialState'
+import * as actions from 'redux/actions/actions'
+
+const mockStore = configureMockStore()
+
+describe('FlashMessagesContainer.jsx', () => {
```
O-Susanoo: I guess you mean FlashMessagesContainer.tsx

src/components/utility/FlashMessagesContainer.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+    return (
+      <div
+        style={{
+          position: 'fixed',
+          bottom: '5%',
+          right: '5%',
+          display: 'flex',
+          flexDirection: 'column',
+          zIndex: 1000,
```
O-Susanoo: hope there is no other element with higher zIndex

test/components/utility/FlashMessagesContainer.test.js
DanielSeehausen: ++ `.tsx`

src/components/routing/Routes.jsx
```diff
@@ -8,11 +8,13 @@ import ResetPassword from '../authentication/ResetPassword'
 import PrivateRoute from "./PrivateRoute"
 import BoardConfigBuilder from "components/admin/board-config-builder/Main"
 import ErrorWrapperProvider from "components/utility/ErrorWrapperProvider";
+import FlashMessagesContainer from "components/utility/FlashMessagesContainer";
```
DanielSeehausen: stray semis

src/components/utility/FlashMessagesContainer.tsx
```diff
@@ -0,0 +1,44 @@
...
+        {flashMessages.map(message => (
+          <FlashMessage
+            key={message.id}
+            message={message}
+          />
+        ))}
+      </div>
+    )
+  }
+}
```
DanielSeehausen: im a `tsx` noob, so forgive me if there is an obvious reason why this can't be, but:

can this component be functional? seeing as it only has a render method?

src/utils/flashmessenger.js
```diff
@@ -0,0 +1,38 @@
+import store from 'redux/store'
```
DanielSeehausen: this file name is missing camelcase

test/components/utility/FlashMessagesContainer.test.js
```diff
@@ -0,0 +1,85 @@
...
+        generateMessage('message2'),
+      ]
+      const store = { flashMessages: messages }
+
+      const { getByText } = renderFlashMessagesContainer(store)
+
+      messages.forEach(msg => getByText(msg.text))
+    })
+  })
+})
```
DanielSeehausen: these look awesome and really well structured. Super easy to read. I don't have anything to contribute structure wise, so i'll try to add something content wise:

I understand we are testing:

1. single message text appears on screen (user experience confirmation)
2. multiple message texts appear on screen (user experience confirmation)
3. a delete method is called (unit test)

I think the only thing that is missing is the user experience confirmation of number 3 (some text displays, then after a time it no longer displays). Not a requirement or blocking request, just something that jumped out, (if we could only have one of 'assert delete method was called' or ' assert the user experience was that the thing disappeared', its the latter that gives us broader coverage)

PR #381, opened 2019-12-09
Reset Running data

### LINKED ISSUE #371: Running process doesn't disappear when a process is finished
Reproduce: Start a process with one phase a a relative short duration (e.g. 30s) - go to the devices view: Even after the process has finished, the device displays the process data and also keep counting up the duration. Here an example with a phase that ran for 30 seconds:

bug_4

O-Susanoo:
Closes #371


src/components/device/DeviceWidget.jsx
```diff
@@ -125,6 +125,8 @@ const DeviceWidget = ({ device: { connectionState, isVirtual, name, state, runni
   useEffect(() => {
     if (isActive) {
       fetchAndSetRunningData(runningProcessId, setRunningData)
+    } else {
```
O-Susanoo: I hesitated between giving it a new useEffect or combining it with this one , i think it's better to have it either so that we don't check twice

O-Susanoo: BTW the issue is just we keep the old running data in the state

PR #385, opened 2019-12-09
Device Widget Feature 

### LINKED ISSUE #384: When I click on a device I see the process currently running and not the process list

**Describe the bug**
When I click on a device I see the process currently running and not the process list.

**To Reproduce**
1. I click on the device
2. I land on the current process running and not on the process list. 
This is not 100% reproducible: out of 10 times, 9 times I landed on the running process and 1 time on the process list. 

**Expected behavior**
I should see the process list.

**Screenshots**
No screenshots. 

**Environment**
Production (https://ospin-app.com)

DeviceID: 5602794b-6eeb-449b-9f6e-868a66fd1547
ProcessID: 360acc26-d983-4e49-9149-b81afbf329e7

**Additional context**

O-Susanoo:
Users with the appropriate process access can now jump straight to a running process from the device widget by hovering on the process details!

image

Closes #384


DanielSeehausen:
I pinged nina about this, as it was not a bug but a feature. 

if we take action on tickets like this (this was from Nina's request) we should be communicating 
with whoever opened the ticket (in this case Nina), re: the solutions we propose/identify. 

I'll follow up with her on this, but let's not take action until we hear confirmation from her that this is a solution that makes sense.

DanielSeehausen:
@O-Susanoo would you be able to update the visual? This is prolly going to go to a changelog too, so anything polished like the current top post in changelog re: the gif asset would really impress the users!

https://dev.ospin-app.com/changelog

if you are able to do this, please keep in mind it shouldn't have any actual client's names on the devices (like celltainer)

DanielSeehausen:
you know I just typed that last comment and I refreshed and I see you already did it -- it looks really good. :p
only request is to remove celltainer as an identifiable name (if you like just inspect and change it to whatever goofy or not goofy name you like)

src/components/device/Device.jsx
```diff
@@ -107,6 +107,15 @@ class Device extends React.Component {
     const activeProcess = getProcessById(processes, processId)
     const activeDevice = getActiveDevice(devices, deviceId)
 
+    const userHasReadAccessToProcess = this.props.processes.some(process => process.id === processId)
```
DanielSeehausen: this should be in `utils/process` I believe!

DanielSeehausen: or `utils/access` - @vdiezel would be the one to ask

src/components/device/Device.jsx
```diff
@@ -107,6 +107,15 @@ class Device extends React.Component {
...
+
+    if (!userHasReadAccessToProcess && processId === activeDevice.runningProcessId) {
```
DanielSeehausen: this should also be in a util. 
Perhaps `utils/device` 
would something like `isRunningProcess(device, process) make sense?

It might seem unnecessary in terms of lines of code, but the intent here is to standardize the way we ask these questions across the code base with a uniform interface - and there is good reason for that at least!

src/components/device/DeviceWidget.jsx
```diff
@@ -62,16 +62,39 @@ const renderVirtual = () => (
...
+      : (
+        <div
+          onMouseLeave={mouseEvent.onMouseLeave}
+          className="meta"
+          style={{
+            justifyContent: 'center',
+            alignItems: 'center',
+            height: '57px',
+            display: 'flex',
+          }}
```
DanielSeehausen: inconsistent usage of ' " here! not blocking, but please be sure to make use of the linter!

src/components/device/DeviceWidget.jsx
```diff
@@ -62,16 +62,39 @@ const renderVirtual = () => (
   </Label>
 )
 
-const renderRunning = ({ startedAtDay, startedAtTime, startedBy, name, duration }) => (
+const renderRunning = ({ startedAtDay, startedAtTime, startedBy, name, duration }, hovered, mouseEvent, clickHandler) => (
```
DanielSeehausen: eslint erroring on this
consider something like:
```js
const renderRunning = (runningData, hovered, mouseEvent, clickHandler) => {
  const { startedAtDay, startedAtTime, startedBy, name, duration } = runningData
}

File: src/components/device/DeviceWidget.jsx [hunk 5]
```diff
@@ -149,7 +178,7 @@ const DeviceWidget = ({ device: { connectionState, isVirtual, name, state, runni
           />
         </Card.Header>
         <Card.Meta>{type}</Card.Meta>
-        {runningData ? renderRunning(runningData) : renderNotRunning(connectionState)}
+        {runningData ? renderRunning(runningData, hovered, mouseEvent, runningProcessClickHandler) : renderNotRunning(connectionState)}
```
DanielSeehausen: eslint erroring here, consider:
```jsx
        {
          runningData
            ? renderRunning(runningData, hovered, mouseEvent, runningProcessClickHandler)
            : renderNotRunning(connectionState)
        }
```

src/components/devices/Devices.jsx
```diff
@@ -40,13 +40,14 @@ class Devices extends React.Component {
...
+    event.stopPropagation()
     this.props.resetActiveDevice()
     const device = this.props.devices.find(device => device.id === deviceId)
-    const userHasReadAccessToProcess = this.props.processes.some(process => process.id === device.runningProcessId)
-    if (userHasReadAccessToProcess) {
+    if (viewRunningProcess) {
       this.props.history.push(`/device/${deviceId}/process/${device.runningProcessId}`)
-    } else {
+    }
+    else {
```
DanielSeehausen: eslint OTBS

src/components/devices/Devices.jsx
```diff
@@ -40,13 +40,14 @@ class Devices extends React.Component {
...
       this.props.history.push(`/device/${deviceId}/processes`)
```
DanielSeehausen: I believe now, under all circumstances, when a user clicks on a device, it should take them to the device/processes.  

Do we still need to be checking anything else like `viewRunningProcess`? That should all be caught now when they click on the button we provided.

The passing of `viewRunningProcess` and `deviceClickHandler` adds a lot of complexity that vanilla react (without router/redux) might require, but in our application all of that passing up and down the component tree is unnecessary with proper use of the two.

DanielSeehausen: in fact, I believe we might be able to move the click handler in general down to the device widget...would that be accurate?

src/components/devices/DeviceList.jsx
```diff
@@ -7,7 +7,8 @@ const renderDeviceWidget = (device, clickHandler, idx) => (
   <DeviceWidget
     key={idx}
     device={device}
-    clickHandler={() => clickHandler(device.id)}
+    clickHandler={(event) => clickHandler(event, device.id, false)}
+    runningProcessClickHandler={(event) => clickHandler(event, device.id, true)}
```
DanielSeehausen: I believe we should avoid passing these callbacks, and instead affix them directly on to the `deviceWidget`, which is the thing that is being clicked after all. For the most part, its just a redirect.

src/components/device/DeviceWidget.jsx
```diff
@@ -62,16 +62,39 @@ const renderVirtual = () => (
...
   <Card.Description>
-    <Card.Content extra className="meta ui internally celled grid">
-      <div style={processMetaDataStyle}>
-        <strong>{name}</strong><br />
-        started by <strong>{startedBy}</strong><br />
-        on <strong>{startedAtDay}</strong>at <strong>{startedAtTime}</strong>
-      </div>
-      <DurationDisplay ms={duration} />
-    </Card.Content>
+    {!hovered
```
DanielSeehausen: you can remove the inversion here by swapping the ? and the :
no need to ! it
low cost way to reduce reading complexity.

src/components/device/DeviceWidget.jsx
```diff
@@ -62,16 +62,39 @@ const renderVirtual = () => (
...
+      ? (
+        <Card.Content onMouseEnter={mouseEvent.onMouseEnter} extra className="meta ui internally celled grid">
```
DanielSeehausen: we are defining these mouseEvents far away from where they are used. 
We can simplify this and reduce ambiguity by just passing `setHover` and calling
`onMouseEnter={() => setHover(true)}`

If you'd like to keep the `mouseEvent` object, it should be moved into this function so the definition is closer to the use.

src/components/device/DeviceWidget.jsx
DanielSeehausen: related - would you be able to break both of those options out into new `render<Name>` functions here? This file is already really big and we work in it a lot, so that means we need to keep the code as modular and clean as possible (either we keep working in it a lot, or we end up breaking it out into separate files. Either way, code quality is going to pay off here)

src/components/device/Device.jsx
O-Susanoo: Nope doesn't exist on both files

src/components/device/processes/Processes.jsx
```diff
@@ -95,88 +94,85 @@ class Processes extends React.Component {
```
O-Susanoo: just removed the Loader because it wasn't actually used

src/components/device/Device.jsx
```diff
@@ -107,6 +107,15 @@ class Device extends React.Component {
     const activeProcess = getProcessById(processes, processId)
     const activeDevice = getActiveDevice(devices, deviceId)
 
+    const userHasReadAccessToProcess = Process.hasProcess(processes, processId)
+
+    if (!userHasReadAccessToProcess && DeviceUtil.isRunningProcess(activeDevice, processId)) {
+      FlashMessenger.error(`you don't have read access to process: ${this.props.match.params.processId}!`)
+
+      history.replace(`/device/${activeDevice.id}/processes`)
+      return null
```
vdiezel: Just curious - why do you need to return null here?

PR #387, opened 2019-12-09
Disable Edit for a finished process

### LINKED ISSUE #386: Deactivate settings values for streams in a finished process
Currently, users can enter values for streams also for a finished process. This shouldn't be posisble. Either we have a bug in the function that cares of this (similar to the one that disbales UI controls for finished phases during the running process) or we don't have anything in place right now.

O-Susanoo:
Closes #386


test/utils/process.test.js
```diff
@@ -483,7 +483,7 @@ describe('phaseIsModifiable', () => {
 
   it('returns true for a running or future phase', () => {
     const phaseIds = ['1', '2']
-
+    mockedProcess.state = 'executable'
```
O-Susanoo: currently the way we determine if a process is running or not is a little bit weird you can check the function deriveProcessState , this is a quick fix for test to be green but the function deriveProcessState shoud be rewritten at some point

vdiezel: I remembered the logic and I think it is the right (slightly weird way to do it) because the process state is always derived from two sources: 
1. Device `state` and `runningProcessId` will tell you which process is running
2. Process `progression` can tell you if the process was already started or not. 

When the process is running (we now this from from the `runningProcessId` of the device), we don't care about progression when we want to know the state. When it is not running, we use the `progression` to see if this process has already run (`finished`) or not (`executable`). We did this, because there was no sense in keeping a seperate process state variable because a process `running` state is always connected to the device state `running` while the rest could be derived from `progression`

O-Susanoo: Ok thank you for the clarification but it seem that you already do keep track of process state because the process object has a state property

DanielSeehausen: mo brings up a good point. It looks like we are deriving process state in multiple places (one on `process.state` and one from `deriveProcessState`

Currently, `deriveProcessState` is only used to set `process.state` in the `objectMapper.js`.
As its currently used, no issue, but if it jumps out as scary that makes sense because there is obvious room for developer error and miss-use of the `deriveProcessState`. 

I will push up a commit to address it.

test/mocks/processDataMock.js
```diff
@@ -4,7 +4,7 @@ export const getMockedProcessData = props => {
     deviceId: 'abcde',
     deviceName: 'mockkDeviceName',
     name: 'mockProcessName',
-tabcomment: 'mockComment',
+    tabcomment: 'mockComment',
```
vdiezel: This should be just `comment`

DanielSeehausen: updated

Issue #192, opened 2019-09-18
Calibration history


vdiezel:
Tricky one:

When a user calibrates in the UI, we always create a fresh calibration based on the raw data. When there is no calibration, we can simply use the device data - easy.

But when there is an existent calibration, we have to strip off mathematically the calibration from the data being sent to the UI (see `getRawDataHere` here https://github.com/ospin-web-dev/REACTor/blob/master/src/utils/device.js) so we can compute the new fresh calibration.

The problem arises when a user applies several calibrations in a row in the same process: When the user uses a data set A to do the calibration and there is no existent calibration, we simply take the raw data and create a calibration entry. But when the user now uses the same set of data to calibrate again, we will apply the calibration reversal on this data set because the UI doesn't know anymore that this was raw data being sent without any calibration applied - there is no history for the calibration. this will distort the computation. The user can circumvent this by always using fresh data that went though the newest calibration - which is ok for now, but in the the future we will need a calibration history we exactly know where 1.) we received already raw data and where we received calibrated data, and 2.) which calibration was applied in which part of the data. This way we can infere how to compute the raw data.

DanielSeehausen:
@vdiezel was this ticket closed?
could it be linked to Adel's request if not?

vdiezel:
So after thinking about it, I think the simplest solution (still quite bit of work) is this: 

We need to store calibration information when a user claibrates within a process. The the question is where are we going to store this. We could do it in device or in process. Logically a calibration belongs more to the device than to the process. This is how I would go about it:

Either create a `calibrationHistory` object on the device or start implementing an `event` property on the device whihc takes events of type `calibration`. When a user creates a calibration, fire an event that contains the absolute time stamp of the event, the relative timestamp within the process and processId. When a user tries to calibrate within the process, we can parse from the history which calibration was at work here and apply the math required.

vdiezel:
Postponed the implementation of a complete calibration history until we an `events` API on processes/devices. This would take several days to implement and would address a very rare problem in the webapp. 

Instead I provided instructions for recalibration of streams within the `CalibrationModal`: #393

PR #394, opened 2019-12-10
Modification of the calibration process for sensors

### LINKED ISSUE #393: Provide Calibration instructions for multiple calibrations on the same stream
A workaorund for #192 that simply tells the user not to use data for calibration that was generated under a previous calibration but instead always use raw device data.

vdiezel:
Previously, users were able to make calibrations using already calibrated data. With little instruction provided to the user, this led to unnecessary complexity and distortions in the calculations of calibrated values. 

In order to simplify this and improve the user experience, the following changes have been made:
 - it is now mandatory to delete existing calibrations before applying a new calibrations 
 - we have provided short instructions on how to recalibrate a sensor

In the future, we will plan to implement a full calibration history for every sensor that will allow customers to calibrate with any data and trace back all applied calibrations on their data.

Calibration_Popup_small
\private
…or recalibration and blocked calibration on top of another
one

This is a workaround for #192 

Now we never apply a calibration on top of another one, but we always reset the current calibration first and then the user has either to use raw data that was created before the calibration or has to wait until the device produces fresh data that can be used for calibration again. However there is no programmatic validation that the users does so.

I deactivated calibrating within finsihed processes because this is counter intuitive with this approach, because we would like to encourage the user to calibrate always on top of fresh data coming from the device, not on old process where the user probably can't recall anymore which data went through calibration and which didn't.

closes #393 
Calibration_Popup


src/components/processviewer/modals/CalibrationModal.jsx
```diff
@@ -488,11 +480,6 @@ class CalibrationModal extends React.Component {
         p2 = [parsedAvrgLower, parsedLowerRef]
       }
 
-      if (existingCalibration){
-        p1[0] = getRawDataValue(activeDevice, stream.id, p1[0])
-        p2[0] = getRawDataValue(activeDevice, stream.id, p2[0])
-      }
-
       params.offset = linear.offset(p1, p2)
```
vdiezel: Now that we are always applying calibration on top of raw data by default and never on top of an existing calibration, we don't need to call `getRawData`

src/components/processviewer/modals/CalibrationModal.jsx
```diff
@@ -522,6 +509,31 @@ class CalibrationModal extends React.Component {
...
+          calibration or in case of a repeated calibration within the same process
+          you can use data that was generated prior to your calibration again.&nbsp;
+          <strong>
+            Do not use data points that were created using your previous calibration. This will distort the result!
+          </strong>
+        </List.Item>
+      </List>}
+    />
+  )
+
```
vdiezel: Popup as seen in the picture

src/components/processviewer/modals/CalibrationModal.jsx
```diff
@@ -802,6 +815,7 @@ class CalibrationModal extends React.Component {
                       primary
                       type='submit'
                       loading={calibrating}
+                      disabled={!!existingCalibration}
```
vdiezel: Now disabled creating a new calibration while there is still an old one existing. You need to delete the previous calibration first.

src/components/processviewer/modals/CalibrationModal.jsx
DanielSeehausen: since this component is already ungodly large - would we be able to break this out into a new component and just import it? seeing as it uses no props should be pretty straightforward

src/components/processviewer/modals/CalibrationModal.jsx
```diff
@@ -522,6 +509,31 @@ class CalibrationModal extends React.Component {
     return generatePhaseSeriesData(this.props.activeProcess, slicedStreamData, this.selectPoints.bind(this, type))
   }
 
+  renderRecalibrationInformation = () => (
+    <Popup
+      position='left center'
+      trigger={
+      <div
+        className='toggle-text margin-b-8'
+      >How do I recalibrate?
```
DanielSeehausen: this spacing got funky - does this look more natural?
```html
      <div className='toggle-text margin-b-8'>
        How do I recalibrate?
      </div>
```

src/components/processviewer/modals/CalibrationModal.jsx
```diff
@@ -522,6 +509,31 @@ class CalibrationModal extends React.Component {
...
+      </div>
+      }
+      content={<List ordered>
+        <List.Item>
+          Reset the calibration using the &lsquo;Reset Calibration&rsquo; Button
+        </List.Item>
+        <List.Item>
+          For calibration with data selection either use data that was generated after the calibration reset for the new
+          calibration or in case of a repeated calibration within the same process
+          you can use data that was generated prior to your calibration again.&nbsp;
```
DanielSeehausen: (is this the same statement?)

Use data that was generated **after** the calibration reset for the new calibration
**OR**
Use data that was generated **prior** to your first calibration.

**Do Not** use data that was generated with a calibration applied.

src/components/processviewer/modals/CalibrationModal.jsx
vdiezel: I was afraid somebody will request this. Yes, we probably should split it up big time.

src/components/processviewer/modals/CalibrationModal.jsx
vdiezel: Almost. The latter can only be done when the initial calibration was created within the same process. Otherwise you start off with an existing calibration and therefore there is no data that can be used *prior* the claibration.If we append to the latter *in case of multiple calibration within the same process*

O-Susanoo:
i think we can go horizontal so that popup isn't too big

vdiezel:
Changed information text as requested and abstracted a couple of components. Also fixed a zooming bug we had still linger around!

src/components/processviewer/modals/calibration/RecalibrationInformationPopup.jsx
```diff
@@ -0,0 +1,32 @@
...
+    content={
+      <List ordered>
+        <List.Item>
+          Reset the calibration using the &lsquo;Reset Calibration&rsquo; Button
+        </List.Item>
+        <List.Item>
+          For calibration with data selection either use data that was generated after
+          the calibration reset for the new calibration or in case of a repeated
+          calibration within the same process you can use data that was generated prior
+          to your calibration again.&nbsp;
```
DanielSeehausen: > ...calibration or, in the case of a repeated calibration within the same process, you can...

DanielSeehausen: ignore this - it was resolved in a later commit

PR #397, opened 2019-12-11
378 provide option to env dependent reporting

### LINKED ISSUE #378: provide the possibility to deactivate client error reporting when in local development environment
When in local development, a lot of errors are being thrown and reported to our `client-side-errors` channel in slack when working on components. This messes up the report channel with not relevant data. We need a way to deactivate the reporting when we are on local dev environment.

vdiezel:
closes #378 

Provides a new configuration file that allows to set options for local development, production environment and test or how react calls them `development`, `production` and `test`. It's abit confusing because the `development` env in REACT corresponds with just the local dev environment. Our `dev` and `prod` environments are both considered to be `production` because they are created using the `builld` script of CRA. But this is good because we probably want the same kind of behaviour on `dev` and `prod`.


src/components/utility/ErrorWrapperProvider.jsx
```diff
@@ -1,12 +1,16 @@
...
 
 class ErrorWrapperProvider extends React.Component {
 
   componentDidCatch(error, errorInfo) {
     const errString = error !== undefined ? error.toString() : 'Undefined error'
-    nexus.utils.logClientError(errString, errorInfo.componentStack)
+
+    if (getConfig().reportClientErrors) {
+      nexus.utils.logClientError(errString, errorInfo.componentStack)
+    }
```
vdiezel: Why use a function that needs to be invoked instead of providing just the config immediately? Because this was the only way to make it testable. When I exported just the config file from `env.config.js ` it alway had set the environmental variable `NODE_ENV` to `test`, even though I set the env var to be `development` or `production` in the test file. While the environment variables do propagate nicely to the component, the files the component requires do not have env vars set (maybe one of you knows why?).  So to be able to mock it, I had to go this way.

test/components/utility/ErrorWrapperProvider.test.jsx
```diff
@@ -5,22 +5,53 @@ import { Router } from 'react-router-dom'
...
+  switch (process.env.NODE_ENV) {
+    case 'development':
+      return envConfig.localEnv
+    case 'test':
+      return envConfig.testEnv
+    case 'production':
+      return envConfig.deployedEnv
+    default: return envConfig.localEnv
+  }
+}
```
vdiezel: This looks like it's double work, why not resuing the `getConfig` function from `env.config.js` and pass in the environment as an argument into it? Because we would then need to pass in the environment whereever we call this in the code. I'd prefer to keep this abstracted in the `env.config.js` and take the hit in the tests.

DanielSeehausen:
@vdiezel did you mean to put @O-Susanoo on as reviewer and not the team account?

test/components/utility/ErrorWrapperProvider.test.jsx
```diff
@@ -29,28 +60,62 @@ describe('ErrorWrapperProvider.jsx', () => {
...
       try {
         renderModal(true)
-      } catch (err) {
-        expect(nexus.utils.logClientError).toHaveBeenCalledTimes(1)
-      }
+      } catch {}
+
+      expect(nexus.utils.logClientError).toHaveBeenCalledTimes(0)
     })
   })
```
vdiezel: Fixes the tests.

test/components/utility/ErrorWrapperProvider.test.jsx
```diff
@@ -5,22 +5,53 @@ import { Router } from 'react-router-dom'
...
+  return <div>Works</div>
+}
+
+describe('ErrorWrapperProvider.jsx', () => {
+
+  let defaultEnv
+
+  const renderModal = (throwError, passedHistory) => {
+
+    const history = passedHistory || createMemoryHistory()
```
DanielSeehausen: can this be a default argument?
```...throwError, passedHistory = createMemoryHistory())```

File: test/components/utility/ErrorWrapperProvider.test.jsx [hunk 5]
```diff
@@ -29,28 +60,62 @@ describe('ErrorWrapperProvider.jsx', () => {
   beforeEach(() => {
     jest.spyOn(console, 'error')
     console.error.mockImplementation(() => {})
+    defaultEnv = process.env.NODE_ENV
   })
 
   afterEach(() => {
     console.error.mockRestore()
```
DanielSeehausen: I believe this may be unnecessary if I am reading it correctly - currently we are:
before every test, spy/mock `console.error`
after every test, restore `console.error` to its original state

luckily, we don't even need to do it in an `afterAll`, because the mock won't apply in other test files, so we don't need to worry about ensuring it is cleared after the last test

I am unsure what we are using this for right now - do we just want to silence `console.error`?
I believe something like `console.error = jest.fn()` should do the trick.

please let me know if I misunderstood the implementation here

test/components/utility/ErrorWrapperProvider.test.jsx
```diff
@@ -29,28 +60,62 @@ describe('ErrorWrapperProvider.jsx', () => {
...
+      expect(pushMock).toHaveBeenCalledWith(expect.objectContaining({
+        pathname: '/report',
+        error: 'Error: Error test',
+        errorInfo: expect.any(Object),
+      }))
+    })
+
+    test('It calls nexus log function in production environment', async () => {
+      process.env.NODE_ENV = 'production'
+      nexus.utils.logClientError = jest.fn()
```
DanielSeehausen: this is already run [here](https://github.com/ospin-web-dev/REACTor/pull/397/files#diff-925e1fdff680453e06fad7063c029bbeR57)

test/components/utility/ErrorWrapperProvider.test.jsx
```diff
@@ -5,22 +5,53 @@ import { Router } from 'react-router-dom'
...
   }
+}
 
-  const history = createMemoryHistory()
-  const renderModal = (throwError) => (
-    render(
+const renderErrorComponent = (throwError) => {
+  if (throwErrror) throw new Error('error text.')
+  return <div>Works</div>
+}
```
DanielSeehausen: this is never used - looks like a refactor of the class component declaration above, but was never implemented

test/components/utility/ErrorWrapperProvider.test.jsx
```diff
@@ -29,28 +60,62 @@ describe('ErrorWrapperProvider.jsx', () => {
...
+      try {
+        renderModal(true, history)
+      } catch {}
+
+      expect(pushMock).toHaveBeenCalledTimes(1)
+      expect(pushMock).toHaveBeenCalledWith(expect.objectContaining({
+        pathname: '/report',
+        error: 'Error: Error test',
+        errorInfo: expect.any(Object),
+      }))
```
DanielSeehausen: would you mind updating the test description string to include this assertion, or putting this into a second test?
currently it is: "When an error has been thrown in a child component"

test/components/utility/ErrorWrapperProvider.test.jsx
```diff
@@ -29,28 +60,62 @@ describe('ErrorWrapperProvider.jsx', () => {
...
+      nexus.utils.logClientError = jest.fn()
+      try {
+        renderModal(true)
+      } catch {}
+
+      expect(nexus.utils.logClientError).toHaveBeenCalledTimes(1)
+    })
+
+    test('It calls nexus log function in test environment', async () => {
+      nexus.utils.logClientError = jest.fn()
```
DanielSeehausen: already run above in beforeEach

test/components/utility/ErrorWrapperProvider.test.jsx
```diff
@@ -29,28 +60,62 @@ describe('ErrorWrapperProvider.jsx', () => {
...
-      }
+      } catch {}
+
+      expect(nexus.utils.logClientError).toHaveBeenCalledTimes(1)
     })
 
-    test('It calls nexus log function', async () => {
+    test('It does not call nexus log function in local environment', async () => {
+      process.env.NODE_ENV = 'development'
       nexus.utils.logClientError = jest.fn()
```
DanielSeehausen: already run above in beforeEach

test/components/utility/ErrorWrapperProvider.test.jsx
```diff
@@ -5,22 +5,53 @@ import { Router } from 'react-router-dom'
 import { createMemoryHistory } from 'history'
 import * as nexus from 'nexus'
 
-describe('ErrorWrapperProvider.jsx', () => {
-  const TestComponent = ({ throwError }) => {
-    if (throwError) throw 'Test error'
+// need to import envConfig here to be able to change 'test' NODE_ENV to
+// the env vars we would like to have by mocking getConfig
```
DanielSeehausen: if im understanding you correctly, are you saying we are unable to override `process.env.NODE_ENV` without importing this here? See changes requested for a potential solution

vdiezel:
> @vdiezel did you mean to put @O-Susanoo on as reviewer and not the team account?

yes!

vdiezel:
> While they aren't all from this PR, I have a few requested changes in the test file (see inline comments). If I am misunderstanding the `mockRestore` implementation, please forgive me.
> 
> Not written inline:
> 
> The config file addition looks great. A few things jump out that I wanted to raise before this goes in:
> I believe this should be outside of `/src`, and instead in the top level.
> 
> Understanding our environment config can only grow (I feel very confident it will grow again), I wanted to ask if you looked in to [this package](https://www.npmjs.com/package/config) at all. It is a pretty standard and lightweight implementation, with documentation and testing, etc etc etc
> 
> They also take care of resolving the NODE_ENV by using different config files, all children of a top level directory. a la: `config/<env>.json`
> 
> If it looks good, now might be a good time to implement it. please let me know what you think!

I tested couple of things. The `config` package is restricted to NodeJS backends and relies on modules of it like the `fs` module. I also checked out his: https://create-react-app.dev/docs/adding-custom-environment-variables/

but that is not exactly what we want. We don't wont set environmental variables in this case, but depending on the `NODE_ENV` being set, create a configuration file which can be used in our source code to trigger certain behaviour. It also appeared to be more tricky to test. 

I finally implemented https://www.npmjs.com/package/react-global-configuration. It does the job in a similar way like my implementation, but provides a bit better debugging and also allows easier reset of the configuration for testing purposes.

re: putting the `env.config.js` file outside of `src`: Per default CRA blocks importing files from outside `/src`. You can bypass this by ejecting the app which I didn't want t do. I also didn't want to install another package to bypass this as there is a reason why the devs of CRA do it: 

https://stackoverflow.com/questions/44114436/the-create-react-app-imports-restriction-outside-of-src-directory

One idea one could follow is to abuse the `NODE_PATH` to also point an `env` file at top level. But I'd like to prevent abusing this, too.

src/env.config.js
```diff
@@ -0,0 +1,21 @@
...
+    reportClientErrors: true,
+    target: 'production',
+  },
+  'test': {
+    reportClientErrors: true,
+    target: 'test',
+  },
+}
+
+const defaultConfigArgs = { freeze: true }
```
vdiezel: when freeze is set to false, it allows overwriting it in the test files.

test/components/utility/ErrorWrapperProvider.test.jsx
vdiezel: good :eyes:

O-Susanoo:
While it may be tempted to use the config package because it's the defacto on this kind of stuff , but viktor is right it seem that React global configuration is what we can do for now , for the folder location there is a trick that i use , i create a symlink inside the src folder to files outside the src folder it usually worked

DanielSeehausen:
looks great - thank you for the thought out response. Sounds like you found the best solution, and I certainly learned more from reading what you took the time to research and communicate back!

PR #399, opened 2019-12-12
389 additional flash notifications

DanielSeehausen:
We have provided additional flash messages in order to improve your experience running processes! Flash messages have been provided/improved for:
- Phase Transitions
- Process Start/End/Pause/Resume

\private
easiest to view each commit - they are broken up by the bullets defined in the tickets!
(please be critical - I am not as confident in this area of the code!)


src/redux/reducers/reducer.js
```diff
@@ -212,15 +212,15 @@ function rootReducer(state = initialState, action) {
...
+    const { processes, processDescriptionSnapshots, flashMessages } = state
     const { processId, phase } = action.payload
     const process = getProcessById(processes, processId)
 
     if (process === undefined) {
       return state
     }
 
-    const updatePhasesFn = process => progressToPhase(process, phase)
+    const updatePhasesFn = selectedProcess => progressToPhase(selectedProcess, phase)
```
DanielSeehausen: this is the only change that isn't related to one of the deliverables -- eslint complained as `process` was already defined, which threw me off when reading it so I updated

src/redux/reducers/reducer.js
```diff
@@ -235,7 +235,34 @@ function rootReducer(state = initialState, action) {
...
+      flashMessages: updatedFlashMessages,
+    }
+  }
+  else if (action.type === ACTIONS.STOP_PROCESS) {
+    const { processes, flashMessages } = state
+    const { processId } = action.payload
+
+    const process = getProcessById(processes, processId)
+
+    const msg = `Process: ${process.name.slice(0, 15)} has been stopped`
```
vdiezel: If you slice the process name it would be nice to provide some `...` s

DanielSeehausen:
paired on with @vdiezel

src/utils/string.js
```diff
@@ -0,0 +1,6 @@
+export function ellipseString(string, max) {
+  const numberOfEllipses = 3
+  if (max < numberOfEllipses) return '...'
+
+  return string.length > max ? `${string.slice(0, max - numberOfEllipses)}...` : string
+}
```
DanielSeehausen: @O-Susanoo you may find this util function useful - I expect we will use it elsewhere in the code base!

src/redux/helper/subscribe.js
```diff
@@ -1,71 +1,102 @@
...
+        ? STATE_CHANGE_DISPATCH_MAP.resumed({ processId: runningProcessId })
+        : STATE_CHANGE_DISPATCH_MAP.started({ processId: data.value.processId, startedAt: Date.now() }) // TODO: started at should be passed from the device
+    } else if (state === 'idle') {
+      if (runningProcessId) {
+        STATE_CHANGE_DISPATCH_MAP.ended({ processId: runningProcessId })
+      }
+    } else if (state === 'paused') {
+      STATE_CHANGE_DISPATCH_MAP.paused({ processId: runningProcessId })
+    }
+
```
O-Susanoo: if i recall good enough at some point viktor and i found out that sometimes state is in weird state , that is non of the above then i guess we should also add an else for such a case but not blocking

DanielSeehausen: good point mo!

these are emitted from the device itself (this is a subscription) - if it doesn't have state we are in big trouble because they sent us a bad payload. What is the safest behavior for us if that were the case?

Issue #396, opened 2019-12-11
Th units in some graphs are not shown


aahmed2021:
All information you provide is _extremely_ helpful to us! Thank you for taking the time.

Actually I talked with Nina and Dennis, Dennis will change the units in the Firmware

**Describe the bug**
-	All Heaters, including the heating plate
-	All Fans
-	Glucose value
-	Lactate value
-	Metabolite temperature value
-	Ph value
-	the name of po2 grahp is not correct

**To Reproduce**
see any finished process

**Expected behavior**
-	All Heaters, including the heating plate: percent [%]
-	All Fans: percent [%]
-	Glucose value : [mMol]
-	Lactate value: [mMol]
-	Metabolite temperature value: [°C]
-	Ph value: [-]
-	the name of po2 grahp should: Oxygen Saturation

**Screenshots**
![image](https://user-images.githubusercontent.com/56728314/70618178-81269580-1c12-11ea-900b-3a0ff61540a2.png)
![image](https://user-images.githubusercontent.com/56728314/70619254-ee3b2a80-1c14-11ea-8661-d917bd1295a2.png)

**Environment**
 - [ ] Development (https://dev.ospin-app.com)
 - [ x ] Production (https://ospin-app.com)

vdiezel:
solved by firmware engineers.

PR #400, opened 2019-12-12
linting the graph and parsing the stream unit consistently with the builder

vdiezel:
This doesn't resolve the issue #396 by Adel as the origin of this issue is on firmware side (the units are simply missing in the publsihed device description)

We did - however - not display units consistently on the process buider and the graphs. I fixed this and linted the graph component on the way.


DanielSeehausen:
> nice

nice

PR #419, opened 2019-12-17
Add Additional info to error message

### LINKED ISSUE #390: Provide more information on client errors that we capture
In addition to the error name and stack trace, we could make use of, whenever applicable, the following:

- [ ] the userId + userName (if any) that was active when it threw
- [ ] deviceId + deviceName (if any) that was active when it threw
- [ ] processId + processName (if any) that was active when it threw
- [ ] perhaps the last few (5?) router url histories (so we can see the what page they were likely on, as well as how they got there
- [ ] the time it threw on the client (while slack has a time, this should be recorded from the client)
- [ ] anything else? imagine we were debugging from one of these - what is the most important stuff that you might want that we have access to?

O-Susanoo:
Closes #390


src/components/utility/ErrorWrapperProvider.jsx
```diff
@@ -2,15 +2,26 @@ import React from 'react'
 import { withRouter } from 'react-router-dom'
 import * as nexus from 'nexus'
 import config from 'react-global-configuration'
+import { connect } from 'react-redux'
+
+const mapStateToProps = state => (
+  { user: state.user }
+)
```
DanielSeehausen: I don't have any strong opinion on this, 
but to highlight consistency, the codebase tends to:

```js
  user: state.user,
})
```

src/components/utility/ErrorWrapperProvider.jsx
```diff
@@ -2,15 +2,26 @@ import React from 'react'
...
 
 class ErrorWrapperProvider extends React.Component {
 
   componentDidCatch(error, errorInfo) {
     const errString = String(error)
 
+    const { userName } = this.props.user
+    const pathName = this.props.history.location.pathname
+    const errorTime = (new Date()).toISOString().replace('T', ' ').split('.')[0]
```
DanielSeehausen: might make sense to abstract this into an exported function within `utils/time`, (or use an appropriate one there, if there is one already)

src/components/utility/ErrorWrapperProvider.jsx
```diff
@@ -2,15 +2,26 @@ import React from 'react'
...
+    const additionalErrorInfo = `\n    User : ${userName} \n    Path : ${pathName} \n    Time : ${errorTime} \n ${errorInfo.componentStack}`
```
DanielSeehausen: understanding our message goes:
`browser` -> `server` -> `slack`

the current implementation is allowing the end use (`slack`) to determine how we are structuring the message and work we are doing where the message originates, which is a code smell as it breaches separation of concern.

I believe it would be closer to best practice if we passed an object here. Something like:
```js
{
  userId: ...,
  userName: ...,
  pathName: ...,
  createdAt: ...,
  stackTrace: ...,
}
```

Any data serialization for view stuff should be done as close as possible to the view (in our case, when the server emits)

src/components/utility/ErrorWrapperProvider.jsx
O-Susanoo: it was like that actually but the linter keep saying it didn't want it like that

src/components/utility/ErrorWrapperProvider.jsx
vdiezel: I have a big refacttoring of the `time/utils` here: https://github.com/ospin-web-dev/REACTor/pull/423. If my PR goes first into code base, you can rebase and add it there. otherwise you will need to resolve conflicts if you modify it. But I agree with daniel.

src/components/utility/ErrorWrapperProvider.jsx
```diff
@@ -2,24 +2,56 @@ import React from 'react'
...
-      error: errString,
-      errorInfo,
+      errorObject,
     })
   }
 
+  serializeErrorData(error, errorInfo) {
+    if (!this.props.user) {
+      this.props.user = { userName: 'unauthenticated user' }
+    }
```
vdiezel: setting props manually here is probably considered an anti-pattern. I think the better solution would be checking if user is defined in props after destructuring it and doing something like this :
` userName = user ? user.userName : 'unauthenticated user'`

src/components/utility/Report.jsx
```diff
@@ -59,7 +60,7 @@ const Report = ({ location: { error, errorInfo } }) => (
         </Grid.Row>
       </Grid>
     </Segment>
-    {errorInfo != null ? RenderStackTrace(error, errorInfo) : ''}
+    {errorObject != null ? RenderStackTrace(errorObject) : ''}
```
vdiezel: Is there a reason why we use `!=` instead of `!==``?

test/components/utility/ErrorWrapperProvider.test.jsx
```diff
@@ -56,8 +62,13 @@ describe('ErrorWrapperProvider.jsx', () => {
...
       expect(pushMock).toHaveBeenCalledWith(expect.objectContaining({
         pathname: '/report',
-        error: 'Error: Error test',
-        errorInfo: expect.any(Object),
+        errorObject: expect.objectContaining({
+          errorString: 'Error: Error test',
+          errorTime: expect.any(Date),
+          pathname: '/',
+          stackTrace: expect.any(String),
+          userName: undefined,
```
vdiezel: If possible, we shoudl assert that the correct userName actually arrives here. So in case of an authenticated user we could pass the user to the store, of in case of unauthenticated user we expect `unauthenticated user`

src/components/utility/ErrorWrapperProvider.jsx
O-Susanoo: actually we will never have this case as we don't show the error page to non login users , but you are definitely right i have made the changes

src/components/utility/Report.jsx
O-Susanoo: not existing any more

test/components/utility/ErrorWrapperProvider.test.jsx
O-Susanoo: thanks i changed it but for the other case as mentioned before will not happen unless the user miss manually with the store

PR #422, opened 2019-12-17
Reset Iot Provider on invitation and revoke

### LINKED ISSUE #382: Page reload needed to see correct status of a shared bioreactor

**Describe the bug**
When user1 gives access to a bioreactor to user2, then user2 needs to reload the page to see the correct bioreactor status. 

**To Reproduce**
1. User1 gives access to the bioreactor to user 2.
2. User2 accepts the invitation.
3. User to goes back to the device page.
4. New device is gray, while it should be blue (process running). 
5. If the page is reloaded then the device color turns into blue. 

**Expected behavior**
No page refresh.

**Screenshots**
No screenshots.

**Environment**
Production (https://ospin-app.com)

Device ID: 5602794b-6eeb-449b-9f6e-868a66fd1547

**Additional context**

O-Susanoo:
The IotProvider is used to connect to our Iot Broker on AWS: https://aws-amplify.github.io/docs/js/pubsub. Viktor had cached the IotProvider at the first call to subscribe to prevent recreating it. The problem is that when we accept a device invitation, a new Iot policy is attached to the user on AWS, but because the IotProvider is cached on client side, it doesn't know about the new IoT policy and therefore forbids to subscribe to the new devices topics immediately.  

It seems that there is a hidden, built-in refresh mechanism in the IoTProviider, because sometimes the user actually got access to the topics, but in 4 out of 5 cases it didn't work. 

Resetting the IotProvider solves this: When a device invitation was accepted successfully (or a device access has been revoked) we create a new IoTProvider on the next subscription attempt, which will then have the updated Iot policies.

Closes #382


src/components/devices/DeviceList.jsx
```diff
@@ -21,8 +21,7 @@ const DeviceList = (props) => {
           ? <p className='text-center margin-top-50'>You don't have access to any device yet.</p>
           : (
             <Card.Group>
-              { filteredDevices.map((device, idx) => renderDeviceWidget(device,
-                deviceClickHandler, idx)) }
+              { filteredDevices.map((device, idx) => renderDeviceWidget(device, idx)) }
             </Card.Group>
```
vdiezel: What was this change for?

O-Susanoo: We have connected the deviceWidget to the store so we didn't need to pass the click handler down however i forgot to remove that from the last dev but i did it this time

PR #423, opened 2019-12-17
Finished processes will now display their duration and end time in

### LINKED ISSUE #398: Show total time of the process
In the page with the different processes, we would like to have an overview of the total run time of an already finished process. 
With that information, the user can understand how long the process was and is able to easy distinguish between them.

I hope it is clear.

**Describe the bug**

**To Reproduce**
_(providing device and process IDs is very helpful when applicable! You can find them in the url)_

**Expected behavior**

**Screenshots**
(an easy way to do this is to make sure an image is copied to clipboard, then you can just `ctrl + v` right in line here and GitHub will do the rest!)

**Environment**
 - [ ] Development (https://dev.ospin-app.com)
 - [ ] Production (https://ospin-app.com)

The IDs can be found in the browser in the url: `ospin-app.com/device/{deviceId}/procecss/{processID}`

**Additional context**

vdiezel:
Finished_Process

Processes which were created before this update will not expose their duration and end time.
\private
Frontend part for the ticket #398 
closes #398 

- Renders the process tile now so that finished processes display there finished date and duration
- Backwards compatible, processes which don't have the `finishedAt` prop (https://github.com/ospin-web-dev/hambda/issues/173) will be rendered as they used to be
- Linted components a bit 
- refactored `utils/time` + test coverage 

Review commit-wise. Last commits are mainly refactoring.


src/components/device/DeviceWidget.jsx
```diff
@@ -126,15 +124,15 @@ const renderProcessDetails = (
...
+      <strong> {ellipseString(startedBy, 8)}</strong>
       <br />
       on
-      <strong>{startedAtDay}</strong>
+      <strong> {startedAtDay}</strong>
       at
-      <strong>{startedAtTime}</strong>
+      <strong> {startedAtTime}</strong>
     </div>
     <DurationDisplay ms={duration} />
```
vdiezel: small display clean up in this component

src/components/device/processes/ProcessTile.jsx
```diff
@@ -65,24 +64,25 @@ const continuouslyUpdateRunningDuration = (startedAt, setRunningDuration) => (
...
       <div className={styleClassName}>
         Started on <strong>{day}</strong>at <strong>{time}</strong>
       </div>
-      <TileDurationDisplay ms={duration} />
+      {duration ? <TileDurationDisplay ms={duration} /> : null}
     </>
   )
 }
 
-const renderNotRunning = (createdAt, styleClassName) => {
```
vdiezel: This got split up into rendering an executable and a finished process now.

src/components/device/processes/ProcessTile.jsx
```diff
@@ -92,9 +92,37 @@ const renderNotRunning = (createdAt, styleClassName) => {
   )
 }
 
+const renderFinished = (createdAt, startedAt, finishedAt, styleClassName) => {
+
+  // legacy for processes which don't have the finishedAt property set
+  if (!finishedAt) return renderExecutable(createdAt, styleClassName)
```
vdiezel: renders processes without the `finishedAt` property set (is done on the backend) as before

src/components/processviewer/body/builder/ControllerModal.jsx
```diff
@@ -53,7 +53,7 @@ class ControllerModal extends React.Component {
 
       if (existingDefault && existingDefault.value === parsedDefault) continue
       if (!existingDefault && parsedDefault === stream.default) continue
-      if (Number.isNaN(parsedDefault) || parsedDefault > stream.max || parsedDefault < stream.min) {
+      if (isNaN(parsedDefault) || parsedDefault > stream.max || parsedDefault < stream.min) {
```
vdiezel: learned this (due to tests!):
`Number.isNaN()` and the gloab `isNaN` are actually two different functions. `Number.isNaN()` returns `true` when the argument passed in is `NaN` as the primitive value, e.g. passing in a string returns `false` . `isNaN` checks if the argument is not a number and also can't be casted into a number.

(this change was in a former PR, but I reverted it here because `validateNumber` in `utils/Time` made me aware of the problem)

src/components/device/processes/ProcessRow.jsx
```diff
@@ -1,5 +1,23 @@
...
+import Time from 'utils/Time'
+import { PROCESS_STATES } from 'utils/process'
+
+const getDuration = (state, startedAt, finishedAt) => {
+  if (!finishedAt || !startedAt) {
+    if (state === PROCESS_STATES.finished) return 'Not exposed'
+    return 'N/A'
+  }
+  return Time.stringFromDuration((finishedAt - startedAt) / 1000)
+}
```
vdiezel: Processes that are finished but don't have the `finishedAt` property set will show `not exposed` in the ProcessTable component for duration and finishedAt. We could, theoretically, calculate their duration from the datapoints timestamps, but thats too much work for now. Open for better naming here.

src/components/device/processes/Processes.jsx
```diff
@@ -89,7 +89,6 @@ class Processes extends React.Component {
       { text: 'in Name', value: 'name' },
       { text: 'in Comment', value: 'comment' },
       { text: 'in State', value: 'state' },
-      { text: 'in Created', value: 'createdAt' }
```
vdiezel: Because we now always store timestamps as milliseconds in the models (see `objectMapper.js`), users can't query by time the same way they used to be able to do it. We need a better implementation where users can enter time ranges!

src/components/device/processes/ProcessesTable.jsx
```diff
@@ -155,7 +157,7 @@ class ProcessesTable extends React.Component {
                   />
               </span>
             </Table.HeaderCell>
-            <Table.HeaderCell colSpan='2'>
+            <Table.HeaderCell colSpan='3'>
               <span>
                   Select action: {' '}
                 <Dropdown
```
vdiezel: two additional columns in the ProcessTable (go to device, in the processes view, select `Table` from the dropdown at the top left corner.

src/redux/helper/objectMapper.js
```diff
@@ -18,7 +18,7 @@ export const createUserLogObject = userLogData => ({
     level: userLogData.level,
     topic: userLogData.topic,
     content: userLogData.content,
-    timestamp: (new Date(userLogData.createdAt)).toString().substr(0,24)
```
vdiezel: this will be now always applied on View layer for consistency reasons. `Time.getTrimmedDateString()` does this now.

src/redux/reducers/reducer.js
```diff
@@ -395,7 +395,10 @@ function rootReducer(state = initialState, action) {
 
     const process = getProcessById(processes, processId)
 
-    const updateFn = targetProcess => Object.assign(targetProcess, { state: 'finished' })
+    const updateFn = targetProcess => Object.assign(targetProcess, {
+      state: 'finished',
+      finishedAt: Date.now(),
```
vdiezel: Setting the `finishedAt` property when a device finishes a process in the application state for real time update.

src/utils/Time.js
```diff
@@ -0,0 +1,85 @@
...
+      return durationString.replace(/^(0[dhms]\s){1,3}/g, '')
+    }
+
+    return durationString
+  }
+
+  static getTrimmedDateString(msSinceEpoch) {
+    return (new Date(msSinceEpoch)).toString().substr(0, 24)
+  }
+}
```
vdiezel: Abstracted time utils into a class and slightly simplified it.

src/utils/Time.js
```diff
@@ -0,0 +1,85 @@
...
+      throw new Error('Passed duration is negative.')
+    }
+
+    if (duration === 0) return '0'
+
+    const durationObj = this.parseTimeElapsedFromSeconds(duration)
+    const durationString = `${durationObj.days}d ${durationObj.hours}h ${durationObj.minutes}m ${durationObj.seconds}s`
+
+    if (trim) {
+      return durationString.replace(/^(0[dhms]\s){1,3}/g, '')
```
vdiezel: trims away leading  zeros for days, hours and minutes. See tests for exact behaviour.

src/components/device/processes/ProcessRow.jsx
O-Susanoo: i don't have a good text either

src/components/device/processes/ProcessRow.jsx
```diff
@@ -1,5 +1,23 @@
...
+
+const getFinishedString = (state, finishedAt) => {
+  if (!finishedAt) {
```
O-Susanoo: Great job

but listening to our inner mean voice , it's better to check the if on the truth not the not :) of a variable

src/utils/Time.js
```diff
@@ -0,0 +1,85 @@
...
+
+  static validateNumber(number) {
+    if (number === 0 || number === '0') return
+    if (!number || isNaN(number.toString())) {
+      throw new Error(`argument: ${number} is not a valid second value`)
+    }
+  }
+
+  static forceDoubleDigits(integer) {
+    return integer < 10 ? `0${integer}` : integer // 7 -> 07
```
O-Susanoo: do we also want to show 00 ?

src/utils/Time.js
```diff
@@ -0,0 +1,85 @@
...
+      seconds: Math.floor(seconds % 60),
+    }
+  }
+
+  static parseTimeElapsedFromMS(ms) {
+    this.validateNumber(ms)
+    return this.parseTimeElapsedFromSeconds(ms / 1000)
+  }
+
+  static timeToSeconds(value, unit) {
```
O-Susanoo: a dictionary could be great here

src/utils/Time.js
vdiezel: Yes, we do!

src/components/device/processes/ProcessRow.jsx
vdiezel: I agree and I fixed this!

src/utils/Time.js
vdiezel: Done!

PR #425, opened 2019-12-19
ADD Modal to show user when offline

### LINKED ISSUE #363: Device Widget shows to be offline but a running process is displayed
MQTT sockets are closed under certain conditions, e.g. when a user closes his laptop for a certain amount of time. We should have a sanity check in place that tells the user that he needs to either refresh the browser window or provide a possibility to subscribe again to the topics. 

Bildschirmfoto von 2019-12-05 14-35-38

O-Susanoo:
Peek 2019-12-19 16-42

Closes #363


O-Susanoo:
We will talk about the message and probably after that right a simple test

src/redux/reducers/reducer.js
```diff
@@ -556,6 +556,9 @@ function rootReducer(state = initialState, action) {
       message => message.id !== messageId
     )]}
   }
+  else if (action.type === ACTIONS.SHOW_CONNECTION_LOSS_MODAL){
```
O-Susanoo: we need to talk about this part because i don't all errors deserve to refresh for now , i can confirm that error code 8 is one for connection loss

src/components/main/Main.jsx
```diff
@@ -109,6 +114,21 @@ class Main extends React.Component {
...
+              It seems that you have lost connection to our cloud server , please refresh again
+            </p>
+          </Modal.Content>
+          <Modal.Actions>
+            <Button color='green' inverted onClick={this.refreshPage}>
+              <Icon name='sync' />
+              Refresh
+            </Button>
+          </Modal.Actions>
+        </Modal>
```
vdiezel: We should abstract this into a sperate component because `Main.jsx` is a very high level component and we should keep it as clean as possible.

src/redux/reducers/reducer.js
vdiezel: For now, we should use the error code of 8 and only in this case fire the connectionLost event . I dont know which other error codes might pop up and I'm afraid we might block the application completely for somebody because of something erroring we don't understand completely. I personally haven't seen any other error popping up yet.

src/components/main/Main.jsx
DanielSeehausen: ++

src/redux/initialState.js
```diff
@@ -20,4 +20,5 @@ export const initialState = {
   displayedProcessPhase: '0',
   processBuilderValidationErrors: [],
   flashMessages: [],
+  connectionLost: false,
```
DanielSeehausen: if we are putting something like this in global state, it needs to be more specific. 
in this case, imagine someone else is reading this. They might as "connection to what?"

in thinking ahead, we will probably be concerned with tracking several connection states "is the internet connected? are we receiving updated from the device? etc."

This _might_ warrant considering something like an object as a top level here (just spitballing ideas):

```js
  network: {
    deviceHeartbeats: {
      <id>: <bool>
    }
  }
```

src/redux/reducers/reducer.js
```diff
@@ -556,6 +556,9 @@ function rootReducer(state = initialState, action) {
...
+    return { ...state, connectionLost: true}
+  }
```
DanielSeehausen: there are some eslint errors here

PR #429, opened 2019-12-30
Device Ownership Transfer

### LINKED ISSUE #428: Create the frontend part for transfering device ownership
Provide UI and thunk/reducer function to transfer the device ownership

vdiezel:
The owner of a device can now pass the device ownership to another user which grants them the highest authorization level on the device. As only one user can be the device owner, the previous owner will be demoted to a regular member of the 'admin' group, which is the second highest authorization level. You have access to this function in the 'Access' tab of your device.

transfer_ownership_with_cursor

\private
closes https://github.com/ospin-web-dev/hambda/issues/134
closes #428

- provides a modal to pass ownership of a device to another device user
- implements an Authorizer utility class that replaces the authorization utility file
- removes `deviceAccess` from global state, now always derived on execution from `Authorizer`

AFTER REVIEW CHANGES: 

- renamed 'owners' to 'admin' in the UI (only on the view layer!) for device and process groups
- requested clean up in the code


src/components/device/manageaccess/User.jsx
```diff
@@ -27,35 +30,38 @@ const User = ({ deviceId, user, access, isOwner, itsMe, deviceAccess,
...
+          disabled={!I.canModifyAccessOf(user).to(activeDevice)}
         />
       </Table.Cell>
       <Table.Cell collapsing>
         <DeletePopup
           confirmHandler={() => deleteAndClose(user.id, accessGroup.name)}
-          disabled={itsMe || !deviceAccess.withdrawDeviceAccess}
+          disabled={!I.canModifyAccessOf(user).to(activeDevice)}
           iconName='user times'
           position='left center'
```
vdiezel: usage of the new Authorizer utility

src/utils/Authorizer.js
```diff
@@ -0,0 +1,46 @@
...
+    isOwnerOf: resource => this.isOwnerOf(resource, user.id),
+    canModifyAccessOf: targetUser => ({
+      to: resource => {
+        if (user.id === targetUser.id) return false
+        if (this.isOwnerOf(resource, user.id)) return true
+
+        const myRights = this.getUserGroupRights(resource, user.id)
+        const targetUserRights = this.getUserGroupRights(resource, targetUser.id)
+        return this.hasManageRightsAccess(myRights) && !this.hasManageRightsAccess(targetUserRights)
+      },
```
vdiezel: I can modify the access of somebody if : 

1. it's not me (line 34)
2. I'm the owner of the reosurce
3. I have the right authorization level for managing access rights AND the person I'd like to change access of doesn't have the right to manage access rights itself (considered to be on an equal level)

vdiezel: Will be changed up a bit in the following comits

src/components/device/manageaccess/AdminToolbar.jsx
```diff
@@ -20,7 +20,7 @@ const AdminToolbar = ({ activeDevice, user }) => (
         <Menu.Item>
           Admin Tools
         </Menu.Item>
-        { user.roleActions.openSsh ? renderSSHAccess(activeDevice) : null }
+        {Authorizer.user(user).hasDeveloperAccess() ? renderSSHAccess(activeDevice) : null }
```
vdiezel: `roleActions` was removed and replaced by Authorizer

src/components/device/manageaccess/User.jsx
```diff
@@ -27,35 +30,41 @@ const User = ({ deviceId, user, access, isOwner, itsMe, deviceAccess,
...
         />
```
vdiezel: usage of the `Authorizer`

src/components/device/manageaccess/User.jsx
```diff
@@ -27,35 +30,41 @@ const User = ({ deviceId, user, access, isOwner, itsMe, deviceAccess,
...
+            </Label>
+          </span>
+        )}
       </Table.Cell>
       <Table.Cell>{email}</Table.Cell>
       <Table.Cell>
         <Dropdown
           onChange={(event, data) => modifyAccess(id, accessGroup.name, event, data)}
-          defaultValue={accessGroup.name}
+          value={accessGroup.name}
```
vdiezel: changed to value otherwise it doesn't update from state changes

src/redux/reducers/reducer.js
```diff
@@ -298,6 +298,31 @@ function rootReducer(state = initialState, action) {
...
+      deviceId, devices, [ updateFnRemoveFromGroup,  updateFnAddToGroup, updateOwnerId ]
+    )
+
+    const updateFnDeviceUsers = deviceUser => updateDeviceUserGroup(
+      deviceUser, newOwnersNewGroupName
+    )
+    const updatedDeviceUsers = findAndUpdate(newOwnerId, deviceUsers, [ updateFnDeviceUsers ])
+
+    return { ...state, devices: updatedDevices, deviceUsers: updatedDeviceUsers }
+  }
```
vdiezel: we need to update in three places: 
1. `ownerId` of the device
2. Move the new owner from his previous group to the owners group
3. update the `accessGroup` of the new owner in `deviceUsers`.

src/utils/Authorizer.js
```diff
@@ -0,0 +1,61 @@
...
+    },
+    canModifyAccessOf: targetUser => ({
+      to: resource => {
+        if (user.id === targetUser.id) return false
+        if (this.isOwnerOf(resource, user.id)) return true
+
+        const userRights = this.getUserGroupRights(resource, user.id)
+        const targetUserRights = this.getUserGroupRights(resource, targetUser.id)
+        return this.manageRightsBitSet(userRights) && !this.manageRightsBitSet(targetUserRights)
+      },
```
vdiezel: I can modify the access of a user to a resource when : 
1. The user is NOT me (you can modify your own rights)
2. You are the owner(then you can do everything)
3. You have the correct authorization level  and the target user CANNOT modify access rights on this resource (user are considered to be equal otherwise)

src/components/device/manageaccess/User.jsx
```diff
@@ -27,35 +30,41 @@ const User = ({ deviceId, user, access, isOwner, itsMe, deviceAccess,
...
           options={groupOptions}
-          disabled={itsMe || isOwner || !deviceAccess.modifyDeviceAccess}
+          disabled={!I.canModifyAccessOf(user).to(activeDevice)}
```
DanielSeehausen: clever use of syntax
when I saw 'I' being used I was going to complain, but this is too nice

src/components/device/modals/AccessDeviceModal.jsx
```diff
@@ -108,14 +108,12 @@ const AccessDeviceModal = props => {
...
               content={errorMessage}
               icon='warning circle'
-              className='max-width-498'
             />
             <Message
               success
               header='Success!'
               content={<p>User {newUserEmail} was invited to device.</p>}
               icon='check circle'
-              className='max-width-498'
```
DanielSeehausen: not needed now, but explanation (simple image would do, or text w/e) would be helpful to know why this was removed because it's a visual only thing and reader can't see in code.

src/components/device/modals/TransferOwnershipModal.jsx
```diff
@@ -0,0 +1,127 @@
...
+  const close = () => {
+    closeHandler()
+  }
+
+  const handleUserChange = (event, data) => {
+    setNewOwner(data.value)
+  }
+
+  const submitNewOwner = async () => {
+    setRequestState({ ...defaultState, isLoading: true })
```
DanielSeehausen: is it intentional that this always spreads the default state, which would overwrite any other state changes?

would it make sense to spread the previous state instead?
```js
  setRequestState(prevState => {
    return { ...prevState, isLoading: true }
  })
```

src/components/device/modals/TransferOwnershipModal.jsx
```diff
@@ -0,0 +1,127 @@
...
+
+    const res = await dispatchCallTransferDeviceOwnership(activeDevice.id, newOwnerId)
+    if (!res.success) {
+      setRequestState({ ...defaultState, isError: true, errorMessage: res.message })
+    } else {
```
DanielSeehausen: reverse conditionals here to avoid the `!`
```js
if (res.success) {
  setRequestState({ ...defaultState, isSuccess: true })
} else  {
  setRequestState({ ...defaultState, isError: true, errorMessage: res.message })
}
```

src/components/device/modals/TransferOwnershipModal.jsx
DanielSeehausen: (same applies for below, if that is helpful)

src/utils/Authorizer.js
DanielSeehausen: this text should be inline documentation in the file

src/utils/Authorizer.js
```diff
@@ -0,0 +1,61 @@
...
+
+  static viewBitSet = rights => rights[0] === '1'
+
+  static editBitSet = rights => rights[1] === '1'
+
+  static operateBitSet = rights => rights[2] === '1'
+
+  static deleteBitSet = rights => rights[3] === '1'
+
+  static manageRightsBitSet = rights => rights[4] === '1'
```
DanielSeehausen: any chance of changing these to just:

```js
static viewRights
// or maybe more explicit
static canView
// or
static hasViewRights
```

reason I bring it up is the 'BitSet' doesn't have much context. It may look like a bit set, but it's ambiguous especially because it's a string.

DanielSeehausen: any chance of changing to `...manageAccess...` instead of just `manageRights...` to be clear this is for editing other user access

src/utils/Authorizer.js
DanielSeehausen: spaces between functions could help reading

src/components/device/manageaccess/ManageAccess.jsx
```diff
@@ -1,51 +1,60 @@
...
+      {Authorizer.user(user).hasDeveloperAccess() ? renderAdminToolbar(activeDevice) : null}
+      <Grid.Row key={0}>
+        <Menu>
+          <Menu.Item>
+            <Button
+              primary
+              icon
+              labelPosition='left'
+              onClick={() => toggleNewUserModal(true)}
+              disabled={!Authorizer.user(user).hasManageAccessRights(activeDevice)}
```
O-Susanoo: i don't know how but at some point this code throw an error it seem that sometime activeDevice is undefined which may cause an error when trying to access the access property of the device ,

again i didn't go throw all of it but i thought you may want to know that

O-Susanoo:
Nice :+1:

src/components/device/manageaccess/User.jsx
vdiezel: yes, had the same train of thought

src/components/device/modals/AccessDeviceModal.jsx
vdiezel: This was an artifact where messages would leave the boundary of the modal. Mo fixed this issue by removing the `image` prop form the modals globally, so messages now stay within the modal and this custom className to limit the Message width is not required anymore.

src/components/device/modals/TransferOwnershipModal.jsx
vdiezel: It is intentional! The default state contains exactly the parameter I want to set here.

src/utils/Authorizer.js
vdiezel: the reason I did this is because I wanted to decouple rights from the bits in the bit mask. Those functions are only supposed to be used internally in this module to actually determine if I have certain rights. 

I decoupled it because I wanted to reserve names like `hasViewRights` to be called together with resources as in line 39 and 43 of `Authorizer.js`

src/components/device/manageaccess/ManageAccess.jsx
vdiezel: activeDevice should never be undefined when we are in this component as we are within the device component. I'll keep an eye on it.

src/utils/Authorizer.js
vdiezel: done!

src/components/device/modals/TransferOwnershipModal.jsx
vdiezel: done!

src/utils/Authorizer.js
vdiezel: done

vdiezel:
Bildschirmfoto von 2020-01-09 14-44-32

src/utils/Authorizer.js
DanielSeehausen: this explanation makes sense to me! especially as it has had time to marinate in my :Gehirn:

DanielSeehausen:
```diff
! one more quick thing! would you be able to update the changelog to reflect that the previous 'owners' group is now 'admin' group. It may be helpful to mention the reason we made the distinction was to avoid confusion between who/what the owner was and the other owners.

PR #448, opened 2020-01-07
Add Dyanmic xAxis Time to graph

O-Susanoo:
Add dynamic xAxis to view time in different format dynamically 
Previously user can view time only in seconds like this

Screenshot from 2020-01-07 15-28-37

Now user will have time format dynamically like this

Dynamic-Time


src/components/processviewer/graphs/Graph.jsx
```diff
@@ -1,9 +1,39 @@
...
+
+  return moment.utc(ms).format(derivedTimeDisplayFormat)
+}
+
+function getTotalDuration(data) {
+  const lastDataSetPoints = data[data.length - 1].dataPoints
+  const lastXValue = lastDataSetPoints[lastDataSetPoints.length - 1].x
+
+  return lastXValue
+}
```
vdiezel: While the formatting function is probably best located here, I can think of use cases where using the helper functions would be interesting in other parts of the application, especially the `deviceTimeDisplayFormat`. We might consider moving them to `TimeUtils` in the future.

PR #452, opened 2020-01-08
Support for Alternative Units

### LINKED ISSUE #98: As a user I can pick a unit for downstreams and upstreams
This needs to be discussed as a team first I think - Ticket is likely bigger than it seems, and will involve some persistence, so it should be broken up into multiple!

When entering values to downstreams a dropdown with compatible units should be provided and also when displaying data for upstreams and a unit conversion should be applied

On logic level we want to keep the streams default unit so we want the conversion applied only on the view layer

vdiezel:
Default units provided by the device can now be overwritten by compatible alternative units. Under device settings you will find a new option that allows users to define new defaults either by single streams or globally for the whole device.

Does your team have a particular need for a certain unit conversion? Get in touch with us and let us know!

unit_conversion_2

\private 
closes #98 

This one turned out to be huge. I'm sorry.

- implements a unit conversion utility 
- implements a new DeviceConfig class that takes care of all customer set device configurations
- implements a new prop on the device called `unitConfig` that contains information about customer set units (updated via a `updateDevice` call (on the backend the only thing that needs to be done is to allow `unitConfig` in the Joi Schema!)
- implements unit conversion in all relevant parts of the application
- implements UI to assign default units by stream or globally by device unit

Things got messy sometimes. Especially in calibration (This component will be refactored again soon, we have work assigned to a ticket on this component anyway)

## UPDATES

- refactored DeviceConfig to have a single responsibility
- improved code smell after review
- refactored Stream.jsx to work with adhoc changes
- changed setting default unit to display the unit type e.g. 'temperature'  instead of the unit '°C'
- cleaned up touched code a bit more
- full test coverage of the new DeviceConfig class


src/appConstants.js
```diff
@@ -0,0 +1,4 @@
+export default {
+  MAX_MS_TIME_BETWEEN_HEARTBEATS: 3000,
+  NUMBER_INPUT_DECIMALS: 3,
+}
```
vdiezel: Renamed this file to appConstants (can't name it `constants` because this package name is taken!

src/components/device/modals/setunitmodal/SetByStreamMenu.jsx
```diff
@@ -0,0 +1,90 @@
...
+        options={unitOptions}
+        onChange={handleUpdate}
+        placeholder='Choose new unit'
+        value={streamUnit}
+      />
+    </Form.Group>
+  )
+}
+
+export default SetByStreamMenu
```
vdiezel: Components allows to chose an existing device unit thta provides alternatives and set this as a default for all stream that have this device unit as their current default. E.g. set globally Celsius to Fahrenheit

src/components/device/modals/setunitmodal/SetByUnitMenu.jsx
```diff
@@ -0,0 +1,73 @@
...
+          placeholder='Choose unit'
+          value={altDeviceUnit}
+        />
+      </Form.Group>
+      {renderAffectedStreams(activeDevice, deviceUnit)}
+    </div>
+  )
+}
+
+export default SetByUnitMenu
```
vdiezel: Components allows to chose an existing device unit thta provides alternatives and set this as a default for all stream that have this device unit as their current default. E.g. set globally Celsius to Fahrenheit

src/components/device/setup/CalibrationTable.jsx
```diff
@@ -13,29 +14,13 @@ const mapDispatchToProps = dispatch => ({
...
-    const aliasEntry = streamNameAliases.find(aliasEntry => aliasEntry.streamId === calibEntry.streamId)
-
-    if (aliasEntry) {
-      return aliasEntry.alias
-    }
-    return streamData.name
-  })
-
-  const streamsData = calibration.map((calibEntry, i) => ({...calibEntry, name: streamNames[i]}))
+  const calibrationEntries = DeviceConfig.getCalibrations(activeDevice)
```
vdiezel: DeviceConfig returns device props (always) overriden with customer set defaults like name aliases, units etc..

src/components/device/setup/UnitsTable.jsx
```diff
@@ -0,0 +1,107 @@
...
+          closeHandler={this.toggleShowSetUnitModal.bind(this)}
+          activeDevice={activeDevice}
+          open={showSetUnitModal}
+        />
+      </>
+    )
+  }
+}
+
+export default connect(null, mapDispatchToProps)(UnitsTable)
```
vdiezel: Component that displays the customer set units in a similar fashion sream defaults, calibration and aliases are displayed in the Device -> Settings view

src/components/processviewer/body/builder/Stream.jsx
```diff
@@ -96,4 +138,3 @@ class Stream extends React.Component {
   }
 }
 export default connect(null, mapDispatchToProps)(Stream)
```
vdiezel: struggled a lot with this component. JavaScript doesn't have a good behaviour when you want float numbers as an input. The number input type with `step=any` behaved strange when something like `100.` was entered (it dispatched a value of undefined in this case). So I was stuck with text input and custm function for rounding etc. 

We also now have to store the target stream value in local state, and store it only (in the default unit of the stream) when we call `save` because we have to wait with the unit conversion and runding until the user is done doing his input because when he enters a decimal seperator it would always removed after unit conversion so users wouldn't e able to enter floats...

src/components/processviewer/body/dataviewer/Stream.jsx
```diff
@@ -235,20 +255,23 @@ class Stream extends React.Component {
...
+    const unitConverter = convertFrom(stream.unit).to(selectedUnit)
 
     let lastDataPoint = 'No Data'
     if (completeStreamData.length > 0) {
-      lastDataPoint = completeStreamData[completeStreamData.length - 1].y
+      lastDataPoint = unitConverter(completeStreamData[completeStreamData.length - 1].y)
     }
 
     const streamData = this.computeSlice(completeStreamData)
+    applyUnitConversionOnData(streamData, unitConverter)
```
vdiezel: I went quite far with the labelFormatter approach, but in the end it was much more complicated, especially with the Calibration. So I convert the data before I pass them to the graphs. 

As streamData is only used in the components, we still can consider this to be just on the view layer.

src/components/processviewer/modals/calibration/CalibrationModal.jsx
```diff
@@ -283,7 +296,7 @@ class CalibrationModal extends React.Component {
         showInLegend: true,
         legendText: 'Interval 1',
         color: 'red',
-        dataPoints: slice
+        dataPoints: applyUnitConversionOnDataPoints(slice, unitConverter),
```
vdiezel: Those slices are generate based on the unconverted data and need conversion to be displayed properly next to the data graphs. This are the lines that are drawn in the calibration when you select points.

src/components/processviewer/modals/calibration/CalibrationModal.jsx
```diff
@@ -636,6 +664,11 @@ class CalibrationModal extends React.Component {
 
     const streamData = this.computeSlice(completeStreamData)
 
+    const unit = DeviceConfig.getStreamUnit(activeDevice, stream)
+    const unitConverter = convertFrom(stream.unit).to(unit)
+
+    applyUnitConversionOnData(streamData, unitConverter)
```
vdiezel: When using the Calibration with the labelFormatter, I had to receive values from the customer that were in the custom unit, parse them back to the original unit to store in the state and parse them back to customer units when I was rendering them. That resulted in a huge effort as there are many input fields. Also I would need to implement onBlur as I can't convert values immediately on the `onChange`` event as this would not allow entering floats again...

This way I have the data from input and selection already in the right format and I have to apply a bit of conversion math whenever the user wants to submit a calibration.

src/redux/helper/objectMapper.js
```diff
@@ -66,7 +66,8 @@ export const createDeviceObject = deviceData => ({
   stateChangeSubscription: null,
   descriptionUpdateSubscription: null,
   dataSubscription: null,
-  phaseChangeSubscription: null
+  phaseChangeSubscription: null,
+  unitConfig: deviceData.unitConfig || [],
```
vdiezel: new prop hat contains the unit config

src/utils/number.js
```diff
@@ -0,0 +1,30 @@
...
+  const trailingZerosPattern = /(?<=\.[0-9]*)0+$/
+
+  if (isInteger.test(string)) {
+    return string.replace(isInteger, '')
+  }
+
+  return trailingZerosPattern.test(string)
+    ? string.replace(trailingZerosPattern, '')
+    : string
+}
```
vdiezel: 4.0120 -> 4.012
4.0000 -> 4

src/utils/number.js
```diff
@@ -0,0 +1,30 @@
...
+  if (INVALID_VALUES.includes(number)) return number
+
+  let roundedNumber = typeof number !== 'number'
+    ? number
+    : number.toFixed(maxDecimals)
+
+  roundedNumber = collapseTrailingZeros(roundedNumber)
+
+  return revertToNumber ? parseFloat(roundedNumber) : roundedNumber
+}
```
vdiezel: A better version of `toFixed()` which has the issue of turning a numbers into strings and filling up missing deccimas with zeros.  This one fixes both.

src/utils/units.js
```diff
@@ -1,70 +1,144 @@
...
+       * in math operations which results e.g. for empty strings in input fields to be
+       * converted to 32 in Fahrenheit (<3 JS)
+       */
+
+      if (sourceUnit !== DEFAULT_UNITS[physicalEntity]) {
+        return value => {
+          if (INVALID_VALUES.includes(value)) return value
+          if (round) {
+            return roundToTheNthDecimal(
+              defaultToUnit[targetUnit](unitToDefault[sourceUnit](value))
```
vdiezel: conversion from any unit (rounding applied per default)

src/utils/units.js
```diff
@@ -1,70 +1,144 @@
...
+            )
           }
-          else return value => defaultToUnit[targetUnit](value)
+          return defaultToUnit[targetUnit](unitToDefault[sourceUnit](value))
+        }
+      }
+      return value => {
+        if (INVALID_VALUES.includes(value)) return value
+        if (round) {
+          roundToTheNthDecimal(defaultToUnit[targetUnit](value))
```
vdiezel: conversion from default unit

src/utils/units.js
```diff
@@ -1,70 +1,144 @@
...
   }
 }
 
-
-
-
-
+export const replaceNoUnitCharacter = (unit, replacement = '') => (
+  unit === NO_UNIT_CHAR ? replacement : unit
+)
```
vdiezel: repaces the no unit character we receive from firmware (`-`) with nothing for displays where you dont want to show this character as a unit, like `5-`

src/utils/units.js
```diff
@@ -1,70 +1,144 @@
...
+    ...p,
+    y: converter(p.y),
+  }))
+)
+
+export const applyUnitConversionOnData = (data, converter) => {
+  data.forEach(dataSlice => {
+    dataSlice.dataPoints = applyUnitConversionOnDataPoints(dataSlice.dataPoints, converter)
+  })
+}
```
vdiezel: heartbreaking for every functional programmer (which I'm not).

test/utils/DeviceConfig.test.js
```diff
@@ -0,0 +1,144 @@
...
+        streamId,
+        unit: '-',
+        value: newDefault,
+        disconnected: true,
+      }
+
+      expect(result).toEqual(exptectedResult)
+    })
+  })
+})
```
vdiezel: A couple of tests. Test coverage will be improved.

src/components/device/modals/setunitmodal/SetUnitModal.jsx
```diff
@@ -0,0 +1,208 @@
...
+      this.setState({ isSuccess: true })
+    } else {
+      this.setState({ isError: true, errorMessage: res.message })
+    }
+  }
+
+  renderInfoMessage(mode) {
+
+    const msgUnitMode = 'This mode will override a device default unit for all affected streams with a compatible unit of your choice'
+    const msgStreamMode = 'This mode will override a stream default unit with a compatible unit of your choice'
```
vdiezel: I'll fix `override` to `overwrite`

DanielSeehausen:
image

image

image

DanielSeehausen:
don't forget to assign project @vdiezel

src/appConstants.js
```diff
@@ -0,0 +1,5 @@
+export default {
+  MAX_MS_TIME_BETWEEN_HEARTBEATS: 3000,
+  NUMBER_INPUT_DECIMALS: 3,
+  NO_UNIT_CHAR: '-',
+}
```
DanielSeehausen: any chance this might have some place in our `env.config` file? Perhaps with an additional application of constants into all the environments, whichever one is picked?

I ask because at least for MAX_MS..., I can see how its natural place could be there (as we could want to edit this between environments)

src/utils/DeviceConfig.js
```diff
@@ -0,0 +1,247 @@
...
+    ))
+  }
+
+  // stream name aliases
+
+  static getStreamNameAlias(device, stream) {
+    const foundAliasEntry = device.streamNameAliases
+      .find(aliasEntry => aliasEntry.streamId === stream.id)
+    return foundAliasEntry ? foundAliasEntry.alias : stream.name
+  }
```
DanielSeehausen: any chance I could lobby to change this to `getStreamDisplayName`?
as it is returning its `name` as a fallback, it seems that is more its prime concern that to get the actual alias, which would return `null` or `undefined` if it didn't actually have one

src/utils/DeviceConfig.js
```diff
@@ -0,0 +1,247 @@
...
+
+  static getNameAliasedStreams(device) {
+    return device.streams.map(stream => {
+      const streamNameAlias = this.getStreamNameAlias(device, stream)
+      return streamNameAlias ? { ...stream, name: streamNameAlias } : stream
+    })
+  }
```
DanielSeehausen: if you do think the above change is a good idea, that could simplify this too:
```js
static getStreamsWithDisplayName(device) {
  // ...
  return {
    ...stream,
    displayName: getStreamDisplayName(device, stream)
  }
}
```

src/components/device/modals/SetStreamDefaultModal.jsx
```diff
@@ -53,27 +54,27 @@ class SetStreamDefaultModal extends React.Component {
...
+    const streamOptions = displayedStreams
       .filter(stream => stream.type === 'downstream')
       .map((stream, i) => {
         return {
           value: stream.id,
           text: stream.name,
           key: i
         }
-      })
+    })
```
DanielSeehausen: indentation was right before change!

src/components/device/modals/setunitmodal/SetByStreamMenu.jsx
```diff
@@ -0,0 +1,90 @@
...
+    ? getUnitOptions(stream.unit)
+      .filter(unit => (
+        currentUnitEntry
+          ? unit !== currentUnitEntry.unit
+          : unit !== stream.unit
+      ))
+      .map(unit => ({ key: unit, value: unit, text: unit }))
+    : []
+)
+
```
DanielSeehausen: i'd lobby to move this into a `utils/stream` if possible! or something similar

src/components/device/modals/setunitmodal/SetByStreamMenu.jsx
```diff
@@ -0,0 +1,90 @@
...
+  const displayedStreams = DeviceConfig.getNameAliasedStreams(device)
+
+  return displayedStreams
+    .filter(stream => getUnitOptions(stream.unit).length > 1)
+    .map((stream, i) => ({
+      value: stream.id,
+      text: stream.name,
+      key: i,
+    }))
+}
```
DanielSeehausen: same thing here re: abstracting to a util file if you think it makes sense

src/components/device/modals/setunitmodal/SetByStreamMenu.jsx
```diff
@@ -0,0 +1,90 @@
...
+  targetStreamId,
+}) => {
+
+  const currentUnitEntry = targetStreamId !== null
+    ? DeviceConfig.getStreamUnitEntry(activeDevice, targetStreamId)
+    : null
+
+  const stream = Device.getStreamById(activeDevice, targetStreamId)
+  const unitOptions = getStreamUnitOptions(stream, currentUnitEntry)
+  const streamOptions = getStreamOptions(activeDevice)
```
DanielSeehausen: here is maybe an argument for moving `getStreamUnitOptions` and `getStreamOptions` into utils, as its already done on line 45

src/components/device/modals/setunitmodal/SetByStreamMenu.jsx
DanielSeehausen: (or I see you have a `utils/units` - dont know where these would be most appropriate between that, device, and stream, but wanted to point it out)

src/components/device/modals/setunitmodal/SetUnitModal.jsx
```diff
@@ -0,0 +1,208 @@
...
+              </Button>
+            </Form>
+          </Modal.Description>
+        </Modal.Content>
+      </Modal>
+    )
+  }
+}
+
+export default connect(null, mapDispatchToProps)(SetUnitModal)
```
DanielSeehausen: im throwing eslint errors in here

src/components/device/modals/setunitmodal/SetUnitModal.jsx
```diff
@@ -0,0 +1,208 @@
...
+
+  async submitByStream() {
+    const { targetStreamId, streamUnit } = this.state
+
+    if (!targetStreamId || !streamUnit) return
+
+    const { activeDevice, dispatchCallUpdateDevice } = this.props
+    const newUnitEntry = { streamId: targetStreamId, unit: streamUnit }
+
+    const updatedUnitConfig = DeviceConfig.addUnitEntries(activeDevice, [newUnitEntry])
```
DanielSeehausen: one clean way to get around this, is to make `addUnitEntries` call a `addUnitEntry` for each entry passed
then you get the other function for free in cases like this and you don't need to wrap in an array. it is cheap to do, because you already have the block you would need for the function right within `addUnitEntries`:

```js
    unitEntries.forEach(unitEntry => {
      const existingEntry = this.getStreamUnitEntry(device, unitEntry.streamId)
      if (existingEntry) {
        existingEntry.unit = unitEntry.unit
      } else {
        unitConfig.push(unitEntry)
      }
    })
```

src/components/device/setup/CalibrationTable.jsx
```diff
@@ -13,29 +14,13 @@ const mapDispatchToProps = dispatch => ({
...
-
-    const aliasEntry = streamNameAliases.find(aliasEntry => aliasEntry.streamId === calibEntry.streamId)
-
-    if (aliasEntry) {
-      return aliasEntry.alias
-    }
-    return streamData.name
-  })
-
-  const streamsData = calibration.map((calibEntry, i) => ({...calibEntry, name: streamNames[i]}))
```
DanielSeehausen: vnice

src/components/device/setup/StreamDefaultsTable.jsx
```diff
@@ -31,26 +32,9 @@ class StreamDefaultsTable extends React.Component {
...
-        return aliasEntry.alias
-      }
-      return streamData.name
-    })
-
-    const streamsData = streamDefaults.map((streamDefault, i) => {
-      return {
-        ...streamDefault, name: streamNames[i]
-      }
-    })
```
DanielSeehausen: +--^----------,--------,-----,--------^-,
 | |||||||||   `--------'     |          O
 `+---------------------------^----------|
   `\_,---------,---------,--------------'
     / XXXXXX /'|       /'
    / XXXXXX /  `\    /'
   / XXXXXX /`-------'
  / XXXXXX /
 / XXXXXX /
(________(                
 `------'

src/components/device/setup/UnitsTable.jsx
```diff
@@ -0,0 +1,107 @@
...
+
+  async deleteStreamUnit(streamId) {
+
+    const { activeDevice, dispatchCallUpdateDevice } = this.props
+
+    const updatedUnitConfig = DeviceConfig.deleteUnitEntry(activeDevice, streamId)
+    const updateParams = { unitConfig: updatedUnitConfig }
+
+    await dispatchCallUpdateDevice(activeDevice.id, updateParams)
+  }
```
DanielSeehausen: once the update resolves, does the new device get pushed into redux?
i was wondering what happens if this fails/internet goes out. it doesn't look like we do any optimistic rendering here, so am I right to assume that? or is the visual updating through some other means?

src/components/processviewer/body/builder/ControllerStream.jsx
```diff
@@ -39,8 +40,8 @@ const ControllerStream = ({
   }
 
   const target = getStreamTarget(activeProcess, phaseId, stream.id)
-  const defaultValue = Device.getStreamDefault(activeDevice, stream)
-  const unit = Device.getStreamUnit(stream)
+  const defaultValue = DeviceConfig.getStreamDefault(activeDevice, stream)
+  const unit = replaceNoUnitCharacter(stream.unit)
```
DanielSeehausen: though you have a default arg for this in the declaration, might make sense to still expose it here as a second arg for the reader so they know what the possibility of this variable value is

src/utils/number.js
```diff
@@ -0,0 +1,30 @@
...
```
DanielSeehausen: eslint desires default exports
I suppose in this case it would mean wrapping your exports in a top level default export named the same as the file and then nesting this inside? unsure just see the rule throwing

src/utils/number.js
DanielSeehausen: any reason not to just have 3 args here? seems the object is excessive. when the name of the other argument is in the title of the function, I think it's fair the function expects a non-default second argument. 

the global constant to a default arg could be misleading, (I believe the global constant NUMBER_INPUT_DECIMALS belongs on the other side of things, where it is passed in, because its a constant that is there to control the view, while this function is generic and has nothing to do with the view. We definitely wouldn't want its default behavior changing in other places this function is being used just because someone changes this constant :O

src/components/processviewer/body/builder/ControllerStream.jsx
DanielSeehausen: (love the function though!)

src/components/device/modals/setunitmodal/SetByStreamMenu.jsx
vdiezel: I don't want to move those into a utils file becuase those optipons are created only for the dropdown a this is the input a semantic ui dropdown expects. I use utils functions like `getUnitOptions`, but in the end I have to map them towards the semantic dropdown API. I don't like the idea of abstracting this away from the component.

src/components/device/setup/UnitsTable.jsx
vdiezel: We are not doing anything when the request fails. The device in the application state gets only updated when the request was successfull. I'm aware that one might think `DeviceConfig.deleteUnitEntry` actually maniplates the device. In fact it just creates a new unitConfig that will be only set when the request was fine. I had to go this way because I wanted to use the `updateDevice` call which means I need to create an updated version of unitConfig BEFORE I do the call to the backend. The alternative would be to create specific endpoints for this, then I could do all required manipulation just in the reducer and update the unitConfig there.

src/utils/DeviceConfig.js
vdiezel: you are right.

src/utils/DeviceConfig.js
vdiezel: I think about this. The name of the stream is always used just for the display so I don't see any issues. So I'm not totally convinced yet :japanese_ogre:

src/utils/number.js
vdiezel: With 2 default arguments an object makes it easier to overwrite defaults, e.g. when you want to overwrite just one and not touching the other, you have to pass undefined as an argument to the parameters if you have to skip them. Also I like to pass objects instead of plain variables when I use boolean flags. I think it is easier to read on the other side: 

``` round(myNumber, 3, { parseToNumber: false })```
is easier to read and understand imho than 
``` round(myNumber, 3, false)```

Comment (vdiezel) [reply src/utils/number.js]: re: the global constant: You are right here. I'll change that.

File: src/utils/number.js (hunk 32 above)
Comment (vdiezel) [reply src/utils/number.js]: I could export an object as default and provide it with a prop named like the function and pass a reference to the function.

Comment (vdiezel) [reply src/utils/number.js]: If I do a default export I can't import it like this : 
`import { roundToTheNthDecimal } from 'utils/unit'`
anymore because import Syntax doesn't allow destructuring: https://github.com/babel/babel-loader/issues/194
because ES6 import syntax (unlike CommonJS) is not using an object, but a so called ImportList. I would need to import the default object first, and then destruct the function in a second line.

File: src/components/processviewer/body/builder/Stream.jsx [hunk 33]
```diff
@@ -5,76 +5,117 @@ import { Select, Form, Grid, Popup } from 'semantic-ui-react'
...
-    inputError: false,
-    errorMessage: '',
-    showControllerModal: false,
-    hasController: !!this.props.stream.controller
+    selectedUnit: this.props.stream.unit,
+    displayedTarget: roundToTheNthDecimal(getStreamTarget(
+      this.props.activeProcess, this.props.phaseId, this.props.stream.id
+    )),
+    displayedMin: roundToTheNthDecimal(this.props.stream.min),
+    displayedMax: roundToTheNthDecimal(this.props.stream.max),
```
DanielSeehausen: there are linter errors throwing here and I think you were right to ignore them.  This is a good chance to mention that if/when we do turn on a hook for linting (maybe we should try to do that before Conor arrives), [eslint provides a nice interface to ignore rules either in functions or at the top of a file](https://eslint.org/docs/2.13.1/user-guide/configuring#disabling-rules-with-inline-comments). Not implying this is a requested change right now, just an opportunity to mention it in case you haven't seen it used. 

Alternatively, you could kill two birds with one stone: its getting pretty crowded with the function calls nested in function. Because we have a complex initialization of state, and because this is already a class, you could throw `state` in a constructor and destructure there, breaking out the more complex assignment of `displayedTarget`.

https://eslint.org/docs/2.13.1/user-guide/configuring#disabling-rules-with-inline-comments

src/components/processviewer/body/dataviewer/Stream.jsx
```diff
@@ -9,6 +9,8 @@ import {
   generatePhaseSeriesData,
```
DanielSeehausen: changes need some lintin

src/components/processviewer/body/dataviewer/Stream.jsx
```diff
@@ -224,6 +228,22 @@ class Stream extends React.Component {
...
+        <span className='float-right'>
+          <strong>
+            {val!== 'No Data' ? val.toFixed(2) : val}
+          </strong>
+          {val !== 'No Data' ? replaceNoUnitCharacter(unit) : ''}
+        </span>
+      </Segment>
+    )
+  }
+
```
DanielSeehausen: @O-Susanoo will this break with the -500/null changes?
We might want to consider a function which does what the first part of the ternary is really asking
(which might be something like: "isValidNumber"). Will this stuff potentially break once it goes on top of your PR?

We can invert the exclusion here and be a lot safer by switching away from asking 
"is this not a bad value that I know could occur" 
to
"is this a valid value"

src/components/processviewer/body/dataviewer/Stream.jsx
```diff
@@ -224,6 +228,22 @@ class Stream extends React.Component {
     return generatePhaseSeriesData(this.props.activeProcess, slicedStreamData)
   }
 
+  renderlastDataPoint(val, unit) {
```
DanielSeehausen: camelcase on l

src/components/processviewer/body/dataviewer/Stream.jsx
```diff
@@ -235,20 +255,23 @@ class Stream extends React.Component {
...
+    const unitOptions = getUnitOptions(stream.unit)
+      .map(unit => ({ text: unit, value: unit }))
+
+    const unitConverter = convertFrom(stream.unit).to(selectedUnit)
 
     let lastDataPoint = 'No Data'
     if (completeStreamData.length > 0) {
-      lastDataPoint = completeStreamData[completeStreamData.length - 1].y
+      lastDataPoint = unitConverter(completeStreamData[completeStreamData.length - 1].y)
     }
```
DanielSeehausen: understanding you only changed the one line, this might be a little out of scope:
using a ternary you can avoid the sandwich code in which lastDataPoint is declared outside of the if block

additionally, you can bury this in the new `renderlastDataPoint`, because that's the only place its used in the render method by passing `completeStreamData` only: `{this.renderlastDataPoint(completeStreamData)}` (the other value that is currently being passed, `selectedUnit`, is in state, so no need to pass it

src/components/processviewer/body/dataviewer/Stream.jsx
```diff
@@ -337,75 +361,69 @@ class Stream extends React.Component {
...
-              </>
+                    </>
+                    }
+                  </Form.Group>
+                </Form>
+              </Segment>
+            </>
           }
         </Grid.Column>
       </Grid.Row>
```
DanielSeehausen: this render method has gotten huge and has nested ternaries and is hard for to follow because I don't have the horsepower in my Gehirn. can any of these easily be made new functional components in other classes? I know this older stuff and the component was huge to begin with, and because this PR is already massive it might not be prudent to do it now, but I wanted to point it out.

src/components/processviewer/modals/calibration/AverageSelectionMenu.jsx
```diff
@@ -29,32 +31,33 @@ const AverageSelectionMenu = ({
...
       zoomParams={zoomParams}
       reloadDataOnRangeChangeHandler={reloadDataOnRangeChangeHandler}
+      unit={unit}
     />
     <Divider/>
     <Form.Group widths='equal'>
       <Form.Input
         fluid
-        label='Start: Lower Value'
+        label='Start: Lower Value [s]'
```
DanielSeehausen: Lower Value(s)

src/components/processviewer/modals/calibration/AverageSelectionMenu.jsx
```diff
@@ -29,32 +31,33 @@ const AverageSelectionMenu = ({
...
         placeholder='click point'
-        value={pointSelectionAtoP.startLower[0]}
+        value={roundToTheNthDecimal(pointSelectionAtoP.startLower[0])}
         onClick={() => resetValue(type, 'startLower')}
       />
       <Form.Input
         fluid
-        label='End: Lower Value'
+        label='End: Lower Value [s]'
```
DanielSeehausen: Lower Value(s)

src/components/processviewer/modals/calibration/AverageSelectionMenu.jsx
```diff
@@ -64,27 +67,27 @@ const AverageSelectionMenu = ({
     <Form.Group widths='equal'>
       <Form.Input
         fluid
-        label='Start: Upper Value'
+        label='Start: Upper Value [s]'
```
DanielSeehausen: Lower Value(s)
[I wasn't sure so I looked up at least one reference that satisfied](https://www.chicagomanualofstyle.org/qanda/data/faq/topics/Plurals.html?old=Plurals11.html)

src/components/processviewer/modals/calibration/AverageSelectionMenu.jsx
```diff
@@ -64,27 +67,27 @@ const AverageSelectionMenu = ({
...
         placeholder='click point'
-        value={pointSelectionAtoP.startUpper[0]}
+        value={roundToTheNthDecimal(pointSelectionAtoP.startUpper[0])}
         onClick={() => resetValue(type, 'startUpper')}
       />
       <Form.Input
         fluid
-        label='End: Upper Value'
+        label='End: Upper Value [s]'
```
DanielSeehausen: Lower Value(s)
[I wasn't sure so I looked up at least one reference that satisfied](https://www.chicagomanualofstyle.org/qanda/data/faq/topics/Plurals.html?old=Plurals11.html)

src/components/processviewer/modals/calibration/CalibrationModal.jsx
```diff
@@ -488,14 +501,25 @@ class CalibrationModal extends React.Component {
...
     this.setState({calibrating: true})
+
+    const unit = DeviceConfig.getStreamUnit(activeDevice, stream)
+    const convertToOriginal = convertFrom(unit).to(stream.unit)
+
+    // offset needs to be converted in relation to 0 of the chosen unit
+    // slope is a relation and independent of the unit, so no conversion required - or?
+
+    const convertedOffset = convertToOriginal(params.offset) - convertToOriginal(0)
+
```
DanielSeehausen: next time we are touching this stuff, which I am sure will be soon, we need to think about abstracting a solid unit conversion API and decouple all this work.

src/components/processviewer/modals/calibration/GraphSelection.jsx
```diff
@@ -10,25 +10,29 @@ const GraphSelection = ({
...
+        stream={stream}
+        data={streamData}
+        zoomParams={zoomParams}
+        Ymin={null}
+        Ymax={null}
+        includeZero
+        reloadDataOnRangeChangeHandler={reloadDataOnRangeChangeHandler}
+      />
+    </div>
+)}
```
DanielSeehausen: seeing as this switched to a non implicit return, you can wrap these with the fragment shorthand (`<></>`) to get the same effect and avoid switching to a `return (...)`

src/utils/DeviceConfig.js
```diff
@@ -0,0 +1,247 @@
...
+/* applying custom configurations like aliasing, display units, etc to the default
+ * device configuration
+ */
+
+import { Device } from 'utils/device'
+import { convertFrom } from 'utils/units'
+
+
+class DeviceConfig {
+
```
DanielSeehausen: If this class is a layer on top of `Device` this is a great opportunity to `extends` the Device class!

src/utils/number.js
DanielSeehausen: i agree re: the options in an object - especially after reading some sandi this morning:
my vote would go to the one you mention there: `round(myNumber, 3, { parseToNumber: false })`, with the options passed as object and the first two directly

src/utils/number.js
DanielSeehausen: i'd be game to turn off/edit that rule in our config then, if you wanted. if this is the only case its throwing in, I wouldn't object! destructure imports or BUST

src/utils/units.js
```diff
@@ -1,70 +1,144 @@
...
+  UNITLESS: 'UNITLESS',
+  PERCENTAGE: 'PERCENTAGE',
 }
 
-export function unitToPhysicalEntity(unit){
-  if (unit === "RPM") return PHYSICAL_ENTITIES.SPEED
-  else if (unit === "K" || unit === "C" || unit === "F") return PHYSICAL_ENTITIES.TEMPERATURE
+const TEMPERATURE_UNITS = ['K', '°C', '°F']
+const PUMP_UNITS = ['RPM']
+const UNITLESS = [CONSTANTS.NO_UNIT_CHAR]
```
DanielSeehausen: seeing as you are controlling the input from the device into something standard on our here (which is awesome and good thinking), what do you think about replacing calls to the `CONSTANTS.NO_UNIT_CHAR` throughout the front end code to instead reference this?

The reason I ask is this seems to be our API and control station for units in general, so the more we can reference it as the source of truth, the more stable and centralized our unit stuff may become

Any objections to either changing this to not be wrapped in an array, OR changing the name to imply its a collection of `unitless` units?

src/utils/units.js
```diff
@@ -1,70 +1,144 @@
...
+import { roundToTheNthDecimal } from 'utils/number'
+import CONSTANTS from 'appConstants'
+
+const INVALID_VALUES = [null, undefined, '', false]
+
 const PHYSICAL_ENTITIES = {
-  TEMPERATURE: "temperature",
-  SPEED: "speed"
+  TEMPERATURE: 'TEMPERATURE',
+  SPEED: 'SPEED',
```
DanielSeehausen: I think we need to be a little more specific re: things like speed/pump_units/rpms
I would imagine a pump unit to be closer to things like volume over time than revolutions, for instance
and speed (or velocity) to be closer to things like distance over time

while its not a problem with our current sensors since we only have RPM right now, we know this api will grow as our devices grow in sophistication/get attached to other things!

src/utils/units.js
```diff
@@ -1,70 +1,144 @@
...
+ *  that takes a value as an input and maps it from sourceUnit to
+ *  targetUnit. We are using default units for intermediate conversion.
  */
 
+import { roundToTheNthDecimal } from 'utils/number'
+import CONSTANTS from 'appConstants'
+
+const INVALID_VALUES = [null, undefined, '', false]
+
 const PHYSICAL_ENTITIES = {
```
DanielSeehausen: Understanding this added `unitless` and `percentage`, we may either miss-using the PHYSICAL_ENTITIES map, or its naming could be more accurate.

src/utils/units.js
```diff
@@ -1,70 +1,144 @@
...
+  PERCENTAGE: 'PERCENTAGE',
 }
 
-export function unitToPhysicalEntity(unit){
-  if (unit === "RPM") return PHYSICAL_ENTITIES.SPEED
-  else if (unit === "K" || unit === "C" || unit === "F") return PHYSICAL_ENTITIES.TEMPERATURE
+const TEMPERATURE_UNITS = ['K', '°C', '°F']
+const PUMP_UNITS = ['RPM']
+const UNITLESS = [CONSTANTS.NO_UNIT_CHAR]
+const PERCENTAGE = ['%']
```
DanielSeehausen: is there an appropriate object these four (19-22) could also be wrapped in to?

src/utils/units.js
```diff
@@ -1,70 +1,144 @@
...
+    case (PERCENTAGE.includes(unit)):
+      return PHYSICAL_ENTITIES.PERCENTAGE
+    default:
+      return undefined
+  }
 }
 
-export const unitOptions = {
-  TEMPERATURE: ["K", "C", "F"]
+const unitOptions = { TEMPERATURE: TEMPERATURE_UNITS }
```
DanielSeehausen: `supportedUnitTypeOptions`

src/utils/units.js
DanielSeehausen: perhaps `UNIT_TYPES`?

src/utils/units.js
DanielSeehausen: perhaps the object is `UNIT_TYPES` or osmething like this?

src/utils/validation.js
```diff
@@ -105,11 +106,18 @@ export function validateStreamTargets(process) {
           message: `Missing target value in group ${streamRef.groupName} at stream ${streamRef.name}`
         })
       } else if (target > streamRef.max || target < streamRef.min) {
+
+       const streamUnit = DeviceConfig.getStreamUnit(device, streamRef)
+
+       const { displayedTarget, displayedMin, displayedMax } = DeviceConfig.getUnitAdjustedStreamValues(streamRef, streamRef.unit, streamUnit, target)
+
```
DanielSeehausen: please lint meeee!

test/utils/DeviceConfig.test.js
DanielSeehausen: awesome start. its super helpful when at least something is there, so that when something is extended or a bug is fixed its easy to jump off from somewhere

src/components/processviewer/body/dataviewer/Stream.jsx
vdiezel: Stream is - after the Calibration - the next component that needs refactoring. Didn't touch it for the reason of not blowing up the PR even further.

src/components/processviewer/modals/calibration/AverageSelectionMenu.jsx
vdiezel: This is a unit and stands for seconds :grin:

test/utils/DeviceConfig.test.js
```diff
@@ -0,0 +1,144 @@
...
+  describe('getNameAliasedStreams', () => {
+    test('returns the streams with applied name aliases', () => {
+
+      const mockedDevice = createDeviceObject(getMockedDeviceData())
+      const streamId = 'streamId3'
+      const stream = Device.getStreamById(mockedDevice, streamId)
+
+      const alias = 'Mr.Nobody'
+      const streamAliasEntry = { streamId, alias }
+      mockedDevice.streamNameAliases = [ streamAliasEntry ]
```
DanielSeehausen: Understanding a lot of work is going in to centralizing and abstracting our interface into these objects into single functions, this jumped out as an opportunity to do just that. As discussed, these seems to be the closest thing we have. Not needed, but if we are mutating these objects we probably want to make sure we are always going through the standard way to do so!

`setStreamConfigurationProperty(device, streamId, prop, values){ ...`

Would be pretty cheap to make a new one as well, to append/add newStreamNameAlias

src/components/processviewer/modals/calibration/AverageSelectionMenu.jsx
vdiezel: see comment at first occurence

src/components/processviewer/modals/calibration/AverageSelectionMenu.jsx
vdiezel: see comment at first occurence

src/components/processviewer/modals/calibration/AverageSelectionMenu.jsx
vdiezel: see comment at first occurence

test/utils/DeviceConfig.test.js
DanielSeehausen: read them again - very awesome

test/utils/device.test.js
```diff
@@ -60,35 +60,6 @@ describe('utils/device', () => {
...
-      const mockedDevice = createDeviceObject(getMockedDeviceData())
-      const streamId = 'streamId3'
-      const stream = Device.getStreamById(mockedDevice, streamId)
-
-      const result = Device.getStreamNameAlias(mockedDevice, stream)
-
-      expect(result).toBe(stream.name)
-    })
-  })
-
```
DanielSeehausen: 🔫

test/utils/number.test.js
```diff
@@ -0,0 +1,36 @@
+import { roundToTheNthDecimal } from '../../src/utils/number'
+
+describe('roundToTheNthDecimal', () => {
+  describe('when receives valid input', () => {
+
+    // function uses 3 decimals + rounding per default
```
DanielSeehausen: will this still be accurate after the change you mentioned you made?

src/components/processviewer/modals/calibration/GraphSelection.jsx
vdiezel: thats just a re-refactoring left over and is fixed.

test/utils/units.test.js
```diff
@@ -0,0 +1,130 @@
+import { convertFrom, getUnitOptions, replaceNoUnitCharacter } from 'utils/units'
+import CONSTANTS from 'appConstants'
+
+/* as learned from poodr2, we are going to test only functions that are
+ * being exported and used within the application; internal helper functions
+ * are considered to have been covered implicitly by this
+ *
+ * test coverage for every physical entity should be added!
```
DanielSeehausen: no rush to get this in if you would like to add them for this PR!

test/utils/units.test.js
```diff
@@ -0,0 +1,130 @@
+import { convertFrom, getUnitOptions, replaceNoUnitCharacter } from 'utils/units'
+import CONSTANTS from 'appConstants'
+
+/* as learned from poodr2, we are going to test only functions that are
+ * being exported and used within the application; internal helper functions
+ * are considered to have been covered implicitly by this
```
DanielSeehausen: image

src/utils/DeviceConfig.js
vdiezel: As a OOP I didn't think of it. I'll give it a shot.

src/utils/number.js
vdiezel: agreed.

src/utils/DeviceConfig.js
DanielSeehausen: 👍

src/components/device/modals/setunitmodal/SetByStreamMenu.jsx
DanielSeehausen: sounds good!

src/utils/number.js
vdiezel: turned it into a warning.

src/components/device/setup/UnitsTable.jsx
DanielSeehausen: sounds great 👍 - seems like a better solution that optimistically rendering here from my end

src/components/processviewer/body/dataviewer/Stream.jsx
DanielSeehausen: 👹

src/components/device/modals/SetStreamDefaultModal.jsx
vdiezel: fixed.

src/components/processviewer/modals/calibration/AverageSelectionMenu.jsx
DanielSeehausen: im a newb
forgive me
👹

src/utils/DeviceConfig.js
DanielSeehausen: 💇

src/components/device/modals/setunitmodal/SetUnitModal.jsx
vdiezel: done!

src/components/device/modals/setunitmodal/SetUnitModal.jsx
vdiezel: fixed.

src/components/processviewer/body/builder/ControllerStream.jsx
vdiezel: fixed.

src/components/processviewer/body/dataviewer/Stream.jsx
vdiezel: fixed.

test/utils/units.test.js
DanielSeehausen: (even if its just a single one, it will help someone extend it a lot when this grows/errors/changes)

src/components/processviewer/body/dataviewer/Stream.jsx
vdiezel: cleaned it up!

src/utils/units.js
vdiezel: I'm actually not controlling the Unit from the device into something standard. `NO_UNIT_CHAR` stands actually for the chracter the firmware sends us when there is no unit that can be applied

src/utils/units.js
vdiezel: done!

src/utils/units.js
vdiezel: done!

src/redux/helper/objectMapper.js
vdiezel: was replaced by a `config` prop instead now.

src/components/processviewer/modals/calibration/CalibrationModal.jsx
vdiezel: I agree.

src/utils/units.js
vdiezel: done.

src/components/device/modals/SetStreamDefaultModal.jsx
```diff
@@ -53,27 +54,27 @@ class SetStreamDefaultModal extends React.Component {
...
+    const displayedStreams = DeviceConfig.getNameAliasedStreams(activeDevice)
 
-    const streamOptions = streamData
+    const streamOptions = displayedStreams
       .filter(stream => stream.type === 'downstream')
       .map((stream, i) => {
         return {
           value: stream.id,
           text: stream.name,
           key: i
```
O-Susanoo: some linting stuff but nothing blocking

src/components/device/modals/setunitmodal/SetByStreamMenu.jsx
```diff
@@ -0,0 +1,90 @@
...
+      .filter(unit => (
+        currentUnitEntry
+          ? unit !== currentUnitEntry.unit
+          : unit !== stream.unit
+      ))
+      .map(unit => ({ key: unit, value: unit, text: unit }))
+    : []
+)
+
+const getStreamOptions = (device) => {
```
O-Susanoo:
```
const getStreamOptions = device => {
```
:100:

src/components/device/modals/setunitmodal/SetByUnitMenu.jsx
```diff
@@ -0,0 +1,73 @@
...
+import { Device } from 'utils/device'
+import { getUnitOptions } from 'utils/units'
+
+const getDeviceUnitOptions = activeDevice => {
+  const deviceUnits = DeviceConfig.getUniqueNonEmptyDeviceUnits(activeDevice)
+
+  return deviceUnits.map(unit => ({ key: unit, text: unit, value: unit }))
+}
+
+const getAlternativeDeviceUnits = (activeDevice, deviceUnit) => (
```
O-Susanoo: if activeDevice is not being used ? than maybe you can remove it ?

otherwise the file look so clean i thought who cleaned my screen

src/components/device/modals/setunitmodal/SetUnitModal.jsx
```diff
@@ -0,0 +1,208 @@
+import React from 'react'
```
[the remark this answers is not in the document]
vdiezel: Did it!

src/components/device/modals/setunitmodal/SetUnitModal.jsx
```diff
@@ -0,0 +1,208 @@
...
+      { value: 'unit', text: 'By device unit', key: 0 },
+      { value: 'stream', text: 'By stream', key: 1 },
+    ]
+
+    return (
+      <Modal
+        size='small'
+        open={this.props.open}
+        onClose={this.props.closeHandler}
+        onClick={event => event.stopPropagation()}
```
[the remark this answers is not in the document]
vdiezel: that is a good question. I'll investigate this.

src/components/device/setup/UnitsTable.jsx
```diff
@@ -0,0 +1,107 @@
...
+                    iconName='close'
+                    position='right center'
+                  />
+                </Table.Cell>
+              </Table.Row>)
+            )}
+          </Table.Body>
+        </Table>}
+        <SetUnitModal
+          closeHandler={this.toggleShowSetUnitModal.bind(this)}
```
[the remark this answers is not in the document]
vdiezel: fixed!

src/components/device/modals/setunitmodal/SetByStreamMenu.jsx
```diff
@@ -0,0 +1,93 @@
...
+      .filter(unit => (excludeCurrentDefaultUnit(unit, currentUnitEntry, stream)))
+      .map(unit => ({ key: unit, value: unit, text: unit }))
+    : []
+)
+
+const getStreamOptions = (device) => {
+  // returns the all streams that do offer at least another possible unit
+
+  const displayedStreams = DeviceConfig.getConfiguredStreams(device)
+  console.log(displayedStreams)
```
vdiezel: line 26 deleted.

src/components/device/modals/setunitmodal/SetByStreamMenu.jsx
vdiezel: fixed

test/utils/units.test.js
vdiezel: done.

PR #472, opened 2020-01-16
460 bug moving phase in running process

### LINKED ISSUE #460: Bugs when moving phases during a running process
this is the triaged result - it has been edited from its original form. history has been preserved.

### Problem
Given a started/running process that has jumped to a phase in the middle of the process:
- future, un-run phases can be dragged before the currently running phase
- the running phase can be moved in its position within the phases
- if a user moves a running phase, then we enter a reality where there can be phases that have not been run _behind_ the currently running phase

----

### Deliverables
- **Current/Active Phase**
  - [x] the running phase should not be able to be moved
- **Past Phases**
  - [x] all phases prior to the running phase _that were started_ should not be move-able. If a phase has not been started, it shouldn't exist in the first place. We can avoid this by ensuring the other deliverables. 
- **Future Phases**
  - [x] should not be placeable _behind_ a running phase. If a user wants to do that, the only use case that we want to make sense is if they are just structuring the process so they can clone it later and have it how they want. If they want to do that, just let them do it after they clone it instead and disable it entirely for running phases.

vdiezel:
Does not allow to move around phases via drag and drop when the phase has already started to change the process history. 

- implements a new function `isValidPhaseOrderChange` to check if the drag'n'drop operation is valid
- implements test to cover the new function
closes #460


.eslintrc.js
```diff
@@ -23,6 +23,7 @@ module.exports = {
     "prefer-template": ["warn"],
     "quotes": ["warn", "single", { "avoidEscape": true }],
     "react/destructuring-assignment": ["warn"],
+    "react/jsx-props-no-spreading": ["warn"],
```
vdiezel: DnD syntax relies on prop spreading.

src/components/processviewer/body/builder/Stream.jsx
```diff
@@ -112,13 +112,7 @@ class Stream extends React.Component {
 
     this.setState({ takingInput: false, currentInput: '' })
 
-    /* we don't want to save the process remotely when it's has been already started
-    /* Ad-hoc changes API will do this
-     */
-
-    if (activeProcess.state !== 'executable') return
-
-    // update remote process
```
vdiezel: removed it here because I realized that saveProcess alreday does that check.

src/utils/process.js
```diff
@@ -604,11 +604,37 @@ export function parseHistoricData(process, streamId, data){
...
+      // don't allow to put future phases before phases that already started
+
+      if (!phaseNeverStarted(process, destPhaseId)) {
+        return false
+      }
+    }
 
-  static getById(processes, processId) {
-    return processes.find(process => process.id === processId)
+    return true
```
vdiezel: New functions to check if moving phase operation is valid

test/utils/process.test.js
```diff
@@ -691,3 +692,94 @@ describe('parseLiveStreamData', () => {
...
+      const futurePhases = [phaseIds[2], phaseIds[3]]
+      const moveFuturePhasesBetweenFuturePhases = createMoveCommands(futurePhases, futurePhases)
+
+      moveFuturePhasesBetweenFuturePhases.forEach(comb => {
+        const res = Process.isValidPhaseOrderChange(fakeProcess, comb.srcId, comb.destId)
+        expect(res).toBe(true)
+      })
+    })
+  })
+})
```
DanielSeehausen: 💣

src/utils/process.js
```diff
@@ -604,11 +604,37 @@ export function parseHistoricData(process, streamId, data){
...
+
+  static phaseNeverStarted = (process, phaseId) => (
+    process.progression.every(entry => entry.phaseId.toString() !== phaseId.toString())
+  )
+
+  static isValidPhaseOrderChange = (process, srcPhaseId, destPhaseId) => {
+
+    if (process.state === 'finished') return false
+
+    if (process.state !== 'executable') {
```
DanielSeehausen: can short circuit here and avoid the block
`...state === 'executable') return true`

PR #477, opened 2020-01-17
Fixes application crashes when multiple users work on the user process simultaniously

### LINKED ISSUE #421: Multiple users working on a process throw errors in the others' UIs
if User 1 create a process and start it , and then user 2 go to devices they will see a running process , but when the user 1 pause the process user 2 will have the following error

image

also mentioned in #433 , which reported:
"When someone else modifies a shared running process I get an OOPSIN and vice versa." 
it has been closed in deference of this one

vdiezel:
\private 
Changelog tag added for testing purposes for the notification bell. Will be removed afterwards.

closes #421 

- implements a check on state updates if user has the process before trying to publish a flash message (that's what caused most of the errors)
- implements a new control flow to manage phase updates: If now a user receives a phase update for process they don't have, app will no crash anymore. Also, when the published phase does not exist in the process (because it was created by another user in adhoc) or doesn't correspond with what we expect as the next phase we set a new 'outdated' flag on the process which will display a small banner in the process viewer and asking for a refresh. Adhoc changes wont be able to be applied unit a user has refreshed.


src/components/processviewer/body/ProcessViewerBody.jsx
```diff
@@ -8,45 +8,71 @@ import Builder from './builder/Builder'
...
+      This process is outdated.
+      <span
+        role='link'
+        className='process-link'
+        onClick={() => global.window.location.reload()}
+      >
+        &nbsp;Refresh
+      </span>
+    </Message>
+    : null
```
vdiezel: Bildschirmfoto von 2020-01-16 17-19-07_cut

vdiezel: Displays this when user receives a phase transition that is not the expected transition for this running process.

src/components/processviewer/body/builder/ChangesFooter.jsx
```diff
@@ -104,6 +104,7 @@ const ChangesFooter = ({
           content='Apply'
           primary
           onClick={() => evaluateChanges(diffs)}
+          disabled={activeProcess.outdated}
```
vdiezel: disables adhoc changes when process was set to outdated.

src/utils/process.js
```diff
@@ -636,4 +636,28 @@ export class Process {
...
+   * another user changes the order of existing phases);
+   * otherwise we assume the process in application state is outdated
+   * because another user changed it (via ad-hoc); in this case set the process
+   * to outdated which will promt the user to update it and not allow adhoc changes
+   */
+
+  static manageProcessProgression = (process, nextPhase) => (
+    this.isValidPhaseProgression(process, nextPhase)
+      ? targetProcess => progressToPhase(targetProcess, nextPhase)
+      : targetProcess => Object.assign(targetProcess, { outdated: true })
```
DanielSeehausen: not blocking:
consider breaking this out to be a more explicitly named function to describe your intent. Same thing that's done on the line above it by calling `progressToPhase`. Perhaps `setProcessOutdated` or something of that nature. While this is the only place its being used, the intended way of doing it is kinda buried in this ternary in an unrelated function. 

It wouldn't be surprising that the next time someone goes to outdate a process somewhere else in the code base, that they don't see this usage here and they duplicate it. Then, we have multiple places where we outdate processes that are not easy to lookup or intuitively know :(.

src/utils/process.js
```diff
@@ -636,4 +636,28 @@ export class Process {
...
+
+  /* check if process phase is in process description
+   * and it is actually the next expected phase of the process (when
+   * another user changes the order of existing phases);
+   * otherwise we assume the process in application state is outdated
+   * because another user changed it (via ad-hoc); in this case set the process
+   * to outdated which will promt the user to update it and not allow adhoc changes
+   */
+
+  static manageProcessProgression = (process, nextPhase) => (
```
DanielSeehausen: seeing where this function is being used, and how it is expected to return update functions, is inconsistent with its name and how you use update functions elsewhere. It looks like this function was written for its usage in the reducer, instead of its application to a process. 

I see this being done elsewhere in the reducer, which seems to make a lot of sense:
`const updateFn = process => parseLiveStreamData(process, data)`

Any reason to diverge, and if not any resistance to recreating that usage in the reducer and changing this to no longer return a function? Then the name makes sense and can be used something like this:
`const updateFn = (process) => Process.manageProcessProgression(process, phase)`

src/components/processviewer/body/builder/ChangesFooter.jsx
DanielSeehausen: nice

src/components/processviewer/body/ProcessViewerBody.jsx
DanielSeehausen: A user might be frustrated by seeing this without context, or information on how they can avoid this scenario. If you think so too, perhaps:
- adding something like "Another user may have edited the process"
or
- adding a hoverable more info tooltip icon

could help increase communication, and keep people from just chalking it up to an engineering fault

DanielSeehausen: If I could use this chance to espouse the virtues of making new components for stuff like this in the future, it can help keep our more actively changing components lean (and I don't mean the bizarre purple drink that seems to now be coming up as the first hit for the word, but instead, the adjective)

vdiezel: done.

src/utils/process.js
vdiezel: done!

src/components/processviewer/body/ProcessViewerBody.jsx
vdiezel: done!

src/utils/process.js
vdiezel: you are right here. I've changed the function to match our pattern again.

PR #491, opened 2020-01-30
483 subscribe to alert publishes

### LINKED ISSUE #483: 2 alert: subscribe to alert emissions and display flash notifications when they are received

DanielSeehausen:
closes #483 

refactors subscriptions

this could be read in commit order, but it might be easier to look at the finished `helper/subscribe` file, as that is where the 🍖 is

this is tested and working for all three cases


src/redux/helper/objectMapper.js
```diff
@@ -60,12 +60,8 @@ export const createDeviceObject = deviceData => ({
   streamDefaults: deviceData.streamDefaults,
   streamNameAliases: deviceData.streamNameAliases,
   dashboardStreams: deviceData.dashboardStreams || [],
+  subscriptions: {},
```
DanielSeehausen: subscriptions are now in a nested object. Easier to work with, and will be easier to provide a device subscriptions utility/interface in the future

src/redux/helper/subscribe.js
```diff
@@ -12,120 +14,155 @@ import {
...
+}
+
+export default async function subscribeToDeviceTopics(device, cachedDevice = {}) {
+  device.subscriptions = mergeSubscriptions(device, cachedDevice)
+
+  const missingSubscriptionTopics = getMissingSubscriptionTopics(device)
+
+  for (const topic of missingSubscriptionTopics) {
+    await attachSubscription(device, topic)
+  }
```
DanielSeehausen: consider starting here!

the main focus here was to DRY everything up and get it in a place where it can be easily structured into an interface later on.

src/redux/helper/subscribe.js
```diff
@@ -12,120 +14,155 @@ import {
...
+  return { ...deviceA.subscriptions, ...deviceB.subscriptions }
+}
+
+function getMissingSubscriptionTopics(device) {
+  const foundTopics = Object.keys(device.subscriptions)
+  return DEVICE_TOPICS.filter(topic => !foundTopics.includes(topic))
+}
+
+export default async function subscribeToDeviceTopics(device, cachedDevice = {}) {
+  device.subscriptions = mergeSubscriptions(device, cachedDevice)
```
DanielSeehausen: combine the cached subscriptions with the existing device subscriptions

src/redux/helper/subscribe.js
```diff
@@ -12,120 +14,155 @@ import {
...
+
+  const missingSubscriptionTopics = getMissingSubscriptionTopics(device)
```
DanielSeehausen: only collect those subscriptions that aren't already present in the subscriptions object

src/redux/helper/subscribe.js
```diff
@@ -12,120 +14,155 @@ import {
...
+ * This has been refactored only so far as to remove duplication.
+ * If anyone is going to pick back up refactoring, one potential place to start
+ * would be by removing the inline functions in DEVICE_TOPIC_SUBSCRIBERS and
+ * replacing them with named functions
+ * This could all go nicely in a class, so we aren't so reliant on the `subscriptions` key of `device` to use its keys to describe subscription topics names, and their values to point to the subscriptions
+ */
 
+/************************************************************************
+                        SUBSCRIPTION HANDLERS
+************************************************************************/
```
DanielSeehausen: these are the actual callbacks to be used when a subscribed topic emits an event
they have been separated from the creation of the device subscriptions

PR #492, opened 2020-01-30
Overhauling the device log page in the device cockpit

### LINKED ISSUE #411: As a user, I should be able to view device logs
### Deliverables
 - [ ] investigate existing react packages for viewing logs. Please provide screenshots/links in this ticket here

As a user on the devices page:
  - [ ] I should be able to navigate to the device's logs in a separate view
As a user on the device logs page:
  - [ ] I should experience a 'loading' while the device logs are fetched (for the MVP, please don't worry about caching, or providing real time push updates)

----

#### Blocked By:

ospin-web-dev/hambda#36
(but dev can start on it at any time!)

### LINKED ISSUE #487: adapt the frontend solution to show device logs that works with the new log service
We already have a view that shows device logs, in device -> log. Load the logs from the new API into this view.

vdiezel:
In preparation of the new feature - the implementation of device logs - the lifeless log page within the device cockpit went through a small overhaul. It will load the most recent device logs and allow you to  query the database for older logs by process and time. Loaded logs can be filtered additionally by level, subject and message content. 

We are going to implement automatic log creation for important events progressively in the coming weeks which will finally populate this part of the application.

device_log_view

\private

Extends the existing device log view to match the new API. The log view loads the most recent device logs into the app state on mounting.

You can: 
- fetch older logs using the 'load more' button
- you can check for more recent logs using 'Refresh' button
- query the database for logs for a specific process, startTime and endTime, in any combination
- filter loaded logs by subject, message, level. Chosen query parameters for process, start- and endTime are also applied as display filter when you set them

it's 700 lines mostly because of breaking up into smaller components and test. I shall not be punished for this.

closes #487 
closes #411


src/components/device/Device.jsx
```diff
@@ -21,7 +22,7 @@ const mapStateToProps = state => ({
   devices: state.devices,
   activeDeviceInitialized: state.activeDeviceInitialized,
   processesInitialized: state.processesInitialized,
-  deviceLogs: state.deviceLogs,
+  deviceLogCollections: state.deviceLogCollections,
```
vdiezel: renames the app state variable to deviceLogCollections which is an array of 
`{ id: deviceId, logs: [ deviceLog1, deviceLog2, ... ] }`

I have chosen this format because and allows me the reuse the reducer pattern with `findAndUpdate` even though it's not immeditately clear that the `id` of the log collection refers to a deviceId.

src/components/device/Device.jsx
```diff
@@ -5,12 +5,13 @@ import { Switch, Route, withRouter, Redirect } from 'react-router-dom'
 
 import ProcessViewer from 'components/processviewer/ProcessViewer'
 import FlashMessenger from 'utils/FlashMessenger'
-import { getDeviceById, getDeviceProcesses } from 'utils/device'
+import { getDeviceById, getDeviceProcesses, Device as DeviceUtils } from 'utils/device'
```
vdiezel: name conflict with `Device` as the component name

src/components/device/Device.jsx
```diff
@@ -5,12 +5,13 @@ import { Switch, Route, withRouter, Redirect } from 'react-router-dom'
...
 import { Process as ProcessUtil, Process } from 'utils/process'
+import DeviceLogs from 'utils/DeviceLogs'
 import { callLoadDevice } from 'redux/thunks/thunks'
 import DeviceMenu from './menu/DeviceMenu'
 import ManageAccess from './manageaccess/ManageAccess'
-import LogList from './log/LogList'
+import LogView from './log/LogView'
```
vdiezel: LogView replaces `LogList.jsx` as the new parent component.

src/components/device/log/LogView.jsx
```diff
@@ -0,0 +1,242 @@
...
+  }
+
+  getFilteredLogs = () => {
+    const { deviceLogs } = this.props
+    return deviceLogs
+      .filter(this.filterByProcess)
+      .filter(this.filterByStartTime)
+      .filter(this.filterByEndTime)
+      .filter(this.filterByAppliedQuery)
+  }
```
vdiezel: Had moved this first into `LogList` but totalPages relies on the total amount of filteredLogs here and I didn't want to put totalPages into the state and then update it there whenever any of the filter parameters changes!

src/components/device/log/LogView.jsx
```diff
@@ -0,0 +1,242 @@
...
+    if (selectedProcess !== 'All') {
+      query.processId = selectedProcess
+    }
+
+    if (startDate !== null) {
+      query.startTime = startDate.getTime()
+    }
+
+    return query
+  }
```
vdiezel: function that creates a new log query for consequent log loading (so loading OLDER logs). If no endTime was provided, we simply take the timestamp of last log we have (logs are always stored sorted).

src/components/device/log/LogView.jsx
```diff
@@ -0,0 +1,242 @@
...
+    await dispatchCallLoadDeviceLogs(activeDevice.id, queryParams)
+    this.setState({ loadingMore: false })
+  }
+
+  async loadMoreDeviceLogs() {
+    const query = this.createLogQuery()
+    await this.loadDeviceLogHandler(query)
+  }
+
+  async refreshDeviceLogs() {
```
vdiezel: load more recent logs by taking the most recent log in app state and using it as the startTime of the query.

src/redux/reducers/reducer.js
```diff
@@ -465,23 +466,29 @@ function rootReducer(state = initialState, action) {
...
   }
   else if (action.type === ACTIONS.SET_ACTIVE_DEVICE){
-    const { devices } = state
+    const { devices, deviceLogCollections } = state
     const { deviceUsers, device } = action.payload
+
+    const updatedDeviceLogCollections = DeviceLogs.normalizeLogCollection(
+      deviceLogCollections, device.id,
+     )
+
```
vdiezel: cant use `findAndUpdate` here because this operation insert an element into the `deviceLogCollections`

src/redux/reducers/reducer.js
```diff
@@ -463,22 +464,28 @@ function rootReducer(state = initialState, action) {
...
-      deviceLogs: [...state.deviceLogs,...newDeviceLogs],
-      allDeviceLogsLoaded: !newDeviceLogs.length,
-      deviceLogsInitialized: true
+  } else if (action.type === ACTIONS.ADD_DEVICELOGS){
+    const { deviceId, fetchedDeviceLogs } = action.payload
+    const { deviceLogCollections } = state
+
+    const normalizedDeviceLogCollection = DeviceLogs.normalizeLogCollection(
+      deviceLogCollections, deviceId,
+    )
```
vdiezel: cant use `findAndUpdate` here because this operation insert an element into the `deviceLogCollections`. Need to create an empty deviceLogCollection if it doesn't exist yet so I can call `findAndUpdate`  in the next step.

src/utils/DeviceLogs.js
```diff
@@ -0,0 +1,46 @@
...
+
+  static mergeLogs = (logCollection, newLogs) => {
+    const oldLogsIds = logCollection.logs.map(log => log.id)
+
+    const logsWithoutDuplicates = newLogs.filter(log => !oldLogsIds.includes(log.id))
+    const mergedLogs = [ ...logCollection.logs, ...logsWithoutDuplicates ]
+    mergedLogs.sort((logA, logB) => logB.createdAt - logA.createdAt)
+
+    logCollection.logs = mergedLogs
+  }
```
vdiezel: because our API is (mostly) not aware of what the user could have in this application state, we sanitize the loaded logs by merging them into the old ones. We remove duplicates and sort by having the most recent logs first.

test/components/device/LogView/LogView.test.js
```diff
@@ -0,0 +1,94 @@
...
+
+      const expectedArgs = [ fakeDevice.id, { endTime } ]
+
+      const loadMoreButton = getByText(loadMoreButtonText)
+      fireEvent.click(loadMoreButton)
+      expect(thunkMock).toHaveBeenCalledTimes(1)
+      expect(thunkMock).toHaveBeenCalledWith(...expectedArgs)
+    })
+  })
+})
```
vdiezel: Can't test a lot of behaviour of the LogView because it's mostly dropdowns....

src/redux/initialState.js
```diff
@@ -13,9 +13,7 @@ export const initialState = {
   heartbeats: { devices: {} }, // i.e. { heartbeats: { devices: { id: <lastHeartbeatTimeInMS> } }
   activeDeviceInitialized: false,
   deviceUsers: [],
-  deviceLogs: [],
-  allDeviceLogsLoaded: false,
-  deviceLogsInitialized: false,
```
vdiezel: `deviceLogsInitialized ` was replaces by a simle check if we alreday have loaded for a device. `allDeviceLogsLoaded `was removed for now. Wanted to make the app state smaller. We could display a flash message instead if no new logs were loaded in the future

src/components/device/Device.jsx
DanielSeehausen: nice!

src/components/device/Device.jsx
DanielSeehausen: this import is a duplicate of `getDeviceById`!
they are doing the same thing it appears. 
|
Would you feel like 💥 `getDeviceById`, seeing as we are moving away from it? I know it would take some reference updates, but maybe it's the appropriate time when we are doing both in the same file. 
What do you think?


I also see we are calling `getDeviceById` as well as `getById` many times in this file (as well as without either function on line 52) 
Is it inconsequential to set the active device once, or is it changing in the lifespan of the component? Looks kinda hairy, so just wanted to ask in case you had an immediate answer.

src/components/device/Device.jsx
```diff
@@ -71,9 +71,14 @@ class Device extends React.Component {
...
-    <LogList activeDevice={getDeviceById(this.props.devices, this.props.match.params.deviceId)}/>
-  )
+  renderLog = () => {
+    const { devices, match, deviceLogCollections } = this.props
+    const activeDevice = DeviceUtils.getById(devices, match.params.deviceId)
+    return (<LogView
+      activeDevice={activeDevice}
+      deviceLogs={DeviceLogs.getLogs(deviceLogCollections, activeDevice.id)}
+    />)
+ }
```
DanielSeehausen: NICE

src/components/device/log/DeviceLogEntry.jsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+import { Table } from "semantic-ui-react"
+import Time from 'utils/Time'
+
+const DeviceLogEntry = props => {
```
DanielSeehausen: `const DeviceLogEntry = { message, subject, createdAt, level } => (`

src/components/device/log/DeviceLogEntry.jsx
```diff
@@ -0,0 +1,19 @@
...
+
+  const { message, subject, createdAt, level } = props.devicelog
```
DanielSeehausen: 🔫

src/components/device/log/DeviceLogEntry.jsx
DanielSeehausen: (if the only thing you want from props is `devicelog`, consider only passing that down instead of whatever is being ignored)

src/components/device/log/DeviceLogEntry.jsx
```diff
@@ -0,0 +1,19 @@
...
+
+  return (
```
DanielSeehausen: 🔫

src/components/device/log/DeviceLogEntry.jsx
```diff
@@ -0,0 +1,19 @@
...
+    <Table.Row warning={level === 'warning'} error={level === 'error'}>
+      <Table.Cell collapsing>{level}</Table.Cell>
+      <Table.Cell>{subject}</Table.Cell>
+      <Table.Cell>{message}</Table.Cell>
+      <Table.Cell>{Time.getTrimmedDateString(createdAt)}</Table.Cell>
+    </Table.Row>
+  )
```
DanielSeehausen: 🔪

src/components/device/log/DeviceLogEntry.jsx
```diff
@@ -0,0 +1,19 @@
...
+}
```
DanielSeehausen: `)`

src/components/device/log/LogView.jsx
```diff
@@ -0,0 +1,242 @@
...
+    const displayedLogs = this.getCurrentPageLogs(filteredLogs)
+
+    return (
+      <Container fluid>
+        <LogViewHeader
+          loadingMore={loadingMore}
+          filterProperty={filterProperty}
+          refreshDeviceLogs={this.refreshDeviceLogs.bind(this)}
+          setFilterProperty={this.setFilterProperty.bind(this)}
+          updateQueryString={this.updateQueryString.bind(this)}
```
DanielSeehausen: eslint throws error on this. React preferences not doing it for performance reasons:
[check out eslint reasoning](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md)

iirc, another option is to use arrow functions in your components instance method declarations.
i.e. 
```js
  refreshDeviceLogs = () => { /* ... */}
```
then it should already be bound to the component, and you can just pass it iirc:
image

[I believe we are compiling this](https://reactjs.org/docs/faq-functions.html#class-properties-stage-3-proposal)

DanielSeehausen: (link to docs is last line)

src/components/device/log/LogViewFooter.jsx
```diff
@@ -0,0 +1,53 @@
...
+            selection
+            options={getPageOptions()}
+            compact
+            onChange={(_, data) => updateResultsPerPage(_, data)}
+          />
+        </span>
+      </Table.HeaderCell>
+      <Table.HeaderCell colSpan='1'>
+        <Button
+          onClick={() => loadMoreDeviceLogs()}
```
DanielSeehausen: I'm a newb and can't remember if we need the arrow function here, or if we can just pass the object

src/redux/initialState.js
DanielSeehausen: good call!

src/utils/DeviceLogs.js
DanielSeehausen: v nice

src/utils/device.js
```diff
@@ -86,6 +86,10 @@ export function setStreamConfigurationProperty(device, streamId, prop, values){
 
 export class Device {
 
+  static getById(devices, deviceId) {
+    return devices.find(device => device.id === deviceId)
+  }
+
```
DanielSeehausen: please 🔫 🔪 the old method if adding this one (see comment above)

src/components/device/log/LogViewFooter.jsx
vdiezel: You are correct.

src/utils/device.js
vdiezel: done.

src/components/device/log/LogView.jsx
vdiezel: Fixed!

PR #495, opened 2020-01-31
436 calibration modal fixes

### LINKED ISSUE #436: Edit calibration modal text/behavior/bugs to reduce ambiguity
There was some user confusion surrounding what was going on here:
- [x] change button text
- [x] edit modal text
- [x] edit 'Two-Point-Calibration' Start/End fields to display the **y values** selected (instead of the times/x values, which are currently displayed case) 
- [x] clear 'Offset' value field back to blank when 'Reset Calibration' is clicked && successful
- [x] fix visual issue with calibration modal spacing
- [x] fix state bug with selected calibration type
- [x] clear notifications on error

**edit button text:**
'Cancel' ->'Close'
'Confirm' -> 'Apply'
image

**edit modal text:**
 - remove the light green smaller text
 - change bolded text to read: 'Calibration deleted successfully'
image

**edit 'Two-Point-Calibration' Start/End fields to display the y values selected (instead of the times/x values, which are currently displayed case)** 
image

**visual issue (happens when text wraps):**
image

**fix state bug with selected calibration type**
to reproduce: open the calibration modal. change to any calibration option that is not the default of 'Offset'. Close the calibration modal (maybe after doing some range selection, I don't know). Re-open the calibration modal. The current calibration option displayed in the dropdown is 'Offset', even though it is showing the previously selected calibration option last time it was opened. Instead, the calibration modal should EITHER: default to offset OR reload the last selected calibration option. Do whichever is lest area/cleaner in the code base for now as we don't have data on the desired behavior.

**the green text should have disappeared:**
image

### LINKED ISSUE #481: Existing calibration doesn't disappear when it is deleted after it has been created in the same Calibration session
In the CalibrationModal, at the top the current calibration is displayed if there is an existing one.
When a calibration is deleted, it should disappear. 

It works when:
 A new calibration is created, it shows up at the top. Refresh browser. Open CalibrationModal for the same stream and delete it using the 'Reset Calibration Button'. It disappears. Instead of refreshing you can also go to another part of the application and return. Then it will work too. 

Without refresh the just deleted calibration remains displayed in the modal and blocks creating a new calibration. It seems a reference issue: After deleting the calibration from the device, the reducer updates the devices correctly, but the 'activeDevice' in the props of calibration modal still has the delted calibration.

vdiezel:
- calibration modal
  - renamed 'Cancel' button to 'Close', 'Confirm' to 'Apply' and 'Reset Calibration' to 'Delete Calibration'
  - provided a new button called 'Reset Inputs' to clear values from input fields
  - fixed bug where error and success messages weren't cleared properly
  - fixed a bug where mode dropdown was not displaying the correct selected mode anymore after the modal was closed and opened again
  - width of modal has been increased to simplify chosing the right datapoints and provide better behaviour on smaller screens when input labels contained a line break
  - selected time ranges in the mode 'Two-Point-Calibration (Average-to-Point)' are now being displayed formatted in the same way as the x-axis of the graphs instead

\private 
closes #436 
closes #481 
issue #481 somehow resolved itself.


src/components/processviewer/graphs/Graph.jsx
```diff
@@ -1,38 +1,14 @@
...
-  const ms = durationInSeconds * 1000
-
-  const days = Math.floor(durationInSeconds / 86400)
-  const displayFormat = (days > 0)
-    ? `[${days}d ]${derivedTimeDisplayFormat}`
-    : derivedTimeDisplayFormat
-
-  return moment.utc(ms).format(displayFormat)
-}
-
```
vdiezel: moved those to `utils/Time` to make use of them in the CalibrationModal!

src/components/processviewer/modals/calibration/AverageSelectionMenu.jsx
```diff
@@ -26,6 +38,11 @@ const AverageSelectionMenu = ({
     zoomParams,
   } = menuParams
 
+  const [startLower, endLower, startUpper, endUpper] = getSelectedTimeRangesArray(
```
vdiezel: parses the seconds to formatted display as on the x-labels.

src/components/processviewer/modals/calibration/CalibrationModal.jsx
```diff
@@ -775,7 +777,7 @@ class CalibrationModal extends React.Component {
                 fluid
                 label='Type'
                 options={calibrationTypes}
-                defaultValue={calibrationTypes[0].value}
+                value={type}
```
vdiezel: fixes the bug where a the selected mode (aka `type`) did not correspond with the displayed controls after closing and opening the modal: We always loaded the first mode when the modal was closed and opened again, while the mode persisted.

src/components/processviewer/modals/calibration/CalibrationModal.jsx
```diff
@@ -797,7 +799,13 @@ class CalibrationModal extends React.Component {
                 disabled={!existingCalibration}
                 onClick={e => this.deleteCalibrationEntry(e)}
               >
-                Reset Calibration
+                Delete Calibration
+              </Button>
+              <Button
+                floated='left'
+                onClick={e => this.resetInputFields(e)}
```
vdiezel: handlers will be fixed to be like this: `onClick={this.handler}`

DanielSeehausen:
if we automate patchnotes we need to have a plan on how it will be parsed

We could re-use the /private parsing plan, where the patchnotes are in first position and everything to be ignored comes after /private. 

A bullet based system sounds good to me. We might want to avoid headers in that case, so patchnotes always parse to a bullet based list. Just spitballing here, but this is something that could parse:

```
PR#101
- fixed bug a 
/private
private information written here for developer meatbags
```

```
PR#103
- fixed bug b 

/private
closes some ticket
```

```
PR#107
- reworked modal calibration 
  - thing about rework
  - thing about rework

/private
closes some ticket
it was easy because reason
```

```
PR#123
- fixed bug d 
```

Ill take this to organ farm

src/components/processviewer/graphs/Graph.jsx
DanielSeehausen: 👍

src/components/processviewer/modals/calibration/CalibrationModal.jsx
```diff
@@ -641,7 +643,7 @@ class CalibrationModal extends React.Component {
     />
   )
 
-  static getCalibrationOptiosn = () => ([
+  static getCalibrationOptions = () => ([
```
DanielSeehausen: 👁

PR #502, opened 2020-02-03
489 unify loader display

### LINKED ISSUE #489: Unify the way we display the loader in the WebApp
When data is loaded we often show a loader - which is a good thing. The loader appears at different places however, depending which part of the application is loading, e.g. for the device view the loader centered, for the Log in device is stuck at the top. This should be solved ith a loading indication that is uniform in the app.

vdiezel:
closes #489 

~~Implements a global LoaderWrapper that allows showing the loader always at the same spot - currently at the middle of the app. 
Implements the new Loader in the main container components.~~

### Update
We keep the loading in the local state for now. Made the Changelog Loader appear the same was as with Devices, and gave the Loader in the LogView a bit more space using a new `DeviceViewLoader` component.
Device Logs

logview_loader

Process Logs
logview_loader_process


src/components/changelog/Changelog.tsx
```diff
@@ -110,6 +118,10 @@ class Changelog extends React.Component<Props, ChangelogState> {
   }
 
   render() {
+
+    const { isLoading } = this.props
+    if (isLoading) return null
```
vdiezel: This trick can be used in components to prevent showing some parts of the component while the rest is still waiting for an API call to resolve.

src/components/profile/Profile.jsx
```diff
@@ -29,143 +24,132 @@ class Profile extends React.Component {
...
+                >
+                  Submit
+                </Form.Button>
+              </Form>
+            </Container>
+          </Grid.Row>}
+        </Grid>
       </Container>
     )
   }
```
vdiezel: Cleaned up this component and removed the whole user log deal - should be replaced by a reusable LogView component!

src/components/utility/LoaderWrapper.jsx
```diff
@@ -0,0 +1,16 @@
...
+
+const mapStateToProps = state => ({ isLoading: state.isLoading })
+
+const LoaderWrapper = ({ isLoading }) => {
+
+  if (!isLoading) return null
+
+  return (
+    <Loader active size="big"/>
+  )
```
vdiezel: We can also have a dimmer on top: https://react.semantic-ui.com/elements/loader/#variations-sizes-inverted

src/components/device/Device.jsx
```diff
@@ -30,35 +31,37 @@ const mapDispatchToProps = dispatch => ({
...
-    this.setState({ initializing: false })
+    await dispatchCallLoadDevice(
+      requestedDevice,
+      match.params.deviceId,
+      user.id,
+      processesInitialized
+    )
+
+    this.setState({ deviceInitialized: true })
+    dispatchSetLoadingState(false)
```
DanielSeehausen: should this dispatch be run after setState finishes its update?

src/components/changelog/Changelog.tsx
```diff
@@ -63,20 +67,24 @@ class Changelog extends React.Component<Props, ChangelogState> {
...
 
     const sortedPrs: GitHubPR[] = gitPrs.sort((a, b) => (
       (new Date(b.closedAt)).valueOf() - (new Date(a.closedAt)).valueOf()
     ))
 
-    this.setState({ gitPrs: sortedPrs, loading: false }, () => {
+    this.setState({ gitPrs: sortedPrs }, () => {
       this.props.callRemoveNotificationsByTopic(this.props.user.id, 'changelog')
     })
+    dispatchSetLoadingState(false)
```
DanielSeehausen: should this dispatch be run after setState finishes its update?

src/components/device/log/LogView.jsx
```diff
@@ -24,18 +27,20 @@ class LogView extends React.Component {
...
     loadingMore: false,
+    updatingRecent: false,
     startDate: null,
     endDate: null,
   }
 
   async componentDidMount() {
-    const { deviceLogs } = this.props
+    const { deviceLogs, dispatchSetLoadingState } = this.props
+    dispatchSetLoadingState(true)
```
DanielSeehausen: should both of these `dispatchSetLoadingState` be inside of `deviceLogs.length` if block?

DanielSeehausen: or even just the first that sets to `true`?

src/components/devices/Devices.jsx
```diff
@@ -17,24 +18,24 @@ const mapDispatchToProps = dispatch => ({
...
+    const { dispatchSetLoadingState } = this.props
+    dispatchSetLoadingState(true)
+
     this.props.history.replace('devices')
     if (!this.props.devicesInitialized) {
       const { devices } = this.props
       await this.props.callGetDevices(devices)
     }
-    this.setState({initializing: false})
+    dispatchSetLoadingState(false)
```
DanielSeehausen: wondering same thing as above, whether or not one or both of these dispatches belong inside of the `if` block

src/components/utility/LoaderWrapper.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+import { Loader } from 'semantic-ui-react'
+import { connect } from 'react-redux'
+
+const mapStateToProps = state => ({ isLoading: state.isLoading })
+
+const LoaderWrapper = ({ isLoading }) => {
+
+  if (!isLoading) return null
```
DanielSeehausen: any benefit to delaying this question:
`if not isLoading: render null`
down into the component instead of up where it is initially called in `Routes`?

The reason I ask, is because if this happens where it is called, we get the bonus of:
1. the reader knows what's going on with that component without looking in it
2. we no longer create a component to then ask if it should exist

DanielSeehausen:
(I did the auto-update button to rebase on changes as they weren't conflicting)

src/components/device/Device.jsx
vdiezel: I think it doesn't really matter. This way you will have the loader covering the contetn for a couple of ms, the other way around you will have a blank screen without loader or content on the screen for a couple of ms.

src/components/device/log/LogView.jsx
vdiezel: Yes, you are right.

src/components/changelog/Changelog.tsx
vdiezel: yep.

src/components/devices/Devices.jsx
vdiezel: yes.

vdiezel:
I never intended the loader in the app state to be the single source for all loading events. We do have local loading indicators, like you see in in a lot of buttons, e.g. in the LogView. The app state loader was intended to be used on the top level container, including device, changelog and profile. Below that level you could have local concurrent loaders for every component.

vdiezel:
So I tried using a HOC and it didn't work as expected. I did this: I created a Loadable HOC that had a state witht an boolean to indicate loading state. It passed a handler to its wrapped component which allowed to set the loading state of the HOC. The HOC returned either a Loader ( `loading === true`) or the wrapped component. 

We use `componentDidMount` to fire API call - now when we use the passed handler form the HOC to indicate loading here, because the the state of the HOC changes by this, it actualy triggers rerender of the wrapped component, triggering the `componentDidMount`again which means if there are no additional conditions in `componentDidMount` , we will be stuck in and endless loop
(simplfied syntax): 

```javascript
function loadable(WrappedComponent) {
  return class extends React.Component {

    state = { loading: false }   

   setLoadingState = loading => this.setState({ loading })

    render() {

     if (loading) return <Loader />

      return <WrappedComponent {...this.props} setLoadingState={this.setLoadingState}/>
    }
  }
}
```
Wrapped Component

```javascript
const WrappedComponent = props => {

   componentDidMount = async() => {
      const { setLoadingState, calltoAPI } = props
      setLoadingState(true)  // will change state chnage of HOC and trigger a new componentDidMount
      await calltoAPI()
      setLoadingState(true) // will change state chnage of HOC and trigger a new componentDidMount
    }

    render() {
      // ....
    }
}
```

DanielSeehausen:
appreciate the detailed answer - I learned a lot from it and our discussion yesterday!

PR #505, opened 2020-02-03
504 make log view reusable

### LINKED ISSUE #504: Refactor LogView component to be reusable also for process and user

vdiezel:
closes #504 

- refactors the LogView to be reusable for user and process
- implemets a Log View in process

User logs can be implemented very easy once the backend provides a way to load logs device independently.

Bildschirmfoto von 2020-02-03 17-56-56


src/components/log/LogView.jsx
```diff
@@ -157,20 +157,20 @@ class LogView extends React.Component {
 
   async loadDeviceLogHandler(queryParams) {
     this.setState({ loadingMore: true })
-    const { activeDevice, dispatchCallLoadDeviceLogs } = this.props
+    const { activeDevice, dispatchCallLoadLogs, defaultQuery } = this.props
 
-    await dispatchCallLoadDeviceLogs(activeDevice.id, queryParams)
+    await dispatchCallLoadLogs(activeDevice.id, { ...queryParams, ...defaultQuery })
```
vdiezel: To make the component flexible it takes a `defaultQuery` prop which will be applied for the current context. in case of the device logs, no default query is passed (the deviceId is a url parameter - I will probably make it also a query parameter in the future to allow fetching logs independently of the device, like user logs). For process logs, it takes `{ processId. 541d1dsaf-13.... }`, for user it will take `{ createdBy: 3feab131-23... }`

src/utils/Logs.js
```diff
@@ -0,0 +1,35 @@
...
+
+    const logsWithoutDuplicates = newLogs.filter(log => !oldLogsIds.includes(log.id))
+    const mergedLogs = [ ...oldLogs, ...logsWithoutDuplicates ]
+    mergedLogs.sort((logA, logB) => logB.createdAt - logA.createdAt)
+
+    return mergedLogs
+  }
+}
+
+export default Logs
```
vdiezel: DeviceLogs -> Logs (changed that everywhere)

src/redux/initialState.js
```diff
@@ -14,7 +14,7 @@ export const initialState = {
   activeDeviceInitialized: false,
   deviceUsers: [],
   logs: [],
-  eisplayedProcessPhase: '0',
```
DanielSeehausen: 🔫 🐕 👨

PR #511, opened 2020-02-07
437 add averaging offset calibration

### LINKED ISSUE #437: Add averaging functionality Offset calibration
Create a new value in the Calibration 'Type' dropdown called 'Offset using Average' (or some other text that makes sense)

It should look identical to the 'Offset' type, but with 4 additional fields:
- it should have 'Start Value' and 'End Value' field, that populate when the user clicks points (see the other calibration types for inspiration)
- it should display the average of the area between + including those two points in a third un-editable field field
- it should provide a fourth editable field 'Desired Average', which defaults to the calculated average (if the user hasn't already entered information)
- when the fourth field is updated (desired average), the fifth field 'Offset (default 0)' should _automatically_ update to display the difference between the fourth field (desired average) and the third field (current average). This should work in reverse. When a user edits the fifth 'Offset' field, it should automatically update the displayed average.

To make some of the control flow easier, one could always disable the 4th and 5th field until values are provided for both the 1st and the 2nd field, and only when those values are populated in the 1st and 2nd field can one edit the 4th or 5th field

vdiezel:
- calibration modal
  - removed the 'Point-to-Point' calibration mode
  - renamed remaining modes to be more on point
  - offset mode now provides a graph selection which calculates the average in the selected range
  - offset mode now expects the correct sensor value as an input instead of the offset itself - offset is calculated automatically from the the correct value and the selected range
  - point selection controls now give a better visual feedback on which point is will be set next

\private
closes #437 

Couldn't really simplfiy the calibration modal as a component because it's so tangled up.... Applied a couple of small changes to the naming, created a Calibration utils where I put a bit of code. It's in a better shape than before. 

Diverged slightly from the implementation of the feature from what the ticket says for 2 reasons: 
1. Semantic-ui complains about inputs switching from`controlled` and `uncontrolled` mode
2. Its more consistent with what we have for the `Linear Calibration (from selection)` mode. Also i think this is more convenient for the customer as usually they see the correct value on their reference sensors and now they can simply put this value into the form instead of calculating the correct value themselves. If they want to enter the offset and slope themselves, they still have the `Linear Calibration (direct input)` mode.

Bildschirmfoto von 2020-02-07 14-33-22


test/utils/Calibration.test.js
```diff
@@ -0,0 +1,72 @@
...
+          secondIntervalReference: '2.2',
+        },
+      }
+
+      const expected = { offset: 0.2, slope: 1 }
+      const res = Calibration.getParams(type, params)
+      const margin = 0.01
+
+      expect(res.offset).toBeWithinRange(expected.offset - margin, expected.offset + margin)
+      expect(res.slope).toBeWithinRange(expected.slope - margin, expected.slope + margin)
```
vdiezel: offset and slope calculation uses float, so don't get exactly the expected values.

src/utils/Calibration.js
```diff
@@ -0,0 +1,223 @@
+import { linear } from 'utils/math'
+import DeviceConfig from 'utils/DeviceConfig'
+import { convertFrom, applyUnitConversionOnDataPoints } from 'utils/units'
+import DataTools from 'utils/DataTools'
+
+
+class Calibration {
```
vdiezel: Simply moved a couple of function from the calibration to this helper file.

src/utils/Calibration.js
```diff
@@ -0,0 +1,357 @@
...
+    return mode.updateAverage(state, props, rawData)
+  }
+
+  attachGraphSlice = (streamData, rawData) => {
+    const { state, props } = this.component
+    this.mode.attachGraphSlice(streamData, rawData, state, props)
+  }
+}
+
+export default Calibration
```
vdiezel: Used this pseudo OOP approach for the calibration for now. It allows to extract a big chunk of logic from the CalibrationModal but has the big disadvantage that within the Modal you don't see anymore what certain state variables are used for (thats why I left at least one ref to every state variable in the component, otherwise eslint will mark them as ''unused'.

The more consistent solution with react would be to make the Calibration menus stateful again and move the logic for every mode to it.

src/utils/calibration/Calibration.js
```diff
@@ -0,0 +1,50 @@
...
+    return mode.updateAverage(state, props, rawData)
+  }
+
+  attachGraphSlice = (streamData, rawData) => {
+    const { state, props } = this.component
+    this.mode.attachGraphSlice(streamData, rawData, state, props)
+  }
+}
+
+export default Calibration
```
vdiezel: Pseudo OOP approach here which relies on modes as ducktypes. In general this was the best way to split up the logic, but it's an antipattern to react: I inject the component state to an instance of calibration to prevent this big amount amount of parameters back and forth. But this way you don't really see in the component itself what is done with its state variables - I had to leave atleast one ref to every state variable in the component so eslint doesn't mark them as `unused` and then somebody cleans them up. 

The more standard react code would be proably to make the calibration menues for each mode stateful and have the mode specific functions and parameters stored there but it would require a lot of duplicate code with the current code structure.

src/components/processviewer/modals/calibration/ExistingCalibrationMessage.jsx
```diff
@@ -0,0 +1,29 @@
...
+import DeviceConfig from 'utils/DeviceConfig'
+import { convertFrom, replaceNoUnitCharacter } from 'utils/units'
+
+const ExistingCalibrationMessage = ({
+  activeDevice,
+  stream,
+}) => {
+
+  const existingCalibration = DeviceConfig.getStreamCalibration(activeDevice, stream.id)
+  const displayedUnit = DeviceConfig.getDisplayedStreamUnit(activeDevice, stream.id)
```
DanielSeehausen: this could go after the early exit on `!existingCalibration`

src/utils/calibration/Calibration.js
DanielSeehausen: discussed this before
looks good - and a good explanation for why standard pattern was circumvented

src/utils/calibration/Calibration.js
```diff
@@ -0,0 +1,50 @@
...
+
+  getParams = () => (
+    this.mode.calculateParams(this.component.state)
+  )
+
+  setPoint = time => (
+    this.mode.setPoint(this.component.state, time)
+  )
+
+  updateAverage = rawData => {
```
DanielSeehausen: this should maybe be: `getUpdatedAverage`
now that we are in the OOP realm, `updateAverage` implies you would be modifying the calibration instance's `average` property, which it does not have (and if it did, it should be declared in the constructor)

src/utils/calibration/Calibration.js
```diff
@@ -0,0 +1,50 @@
...
+      default:
+        throw new Error(`Unknown calibration type: ${type}`)
+    }
+  }
+
+  getParams = () => (
+    this.mode.calculateParams(this.component.state)
+  )
+
+  setPoint = time => (
```
DanielSeehausen: discussed naming here
`set<Prop>` signals a mutation of an object's state

src/utils/calibration/LinearCalibrationSelectionMode.js
```diff
@@ -0,0 +1,166 @@
...
+
+    if (parsedFirstIntervalRef < parsedSecondIntervalRef) {
+      p1 = [firstIntervalAverage, parsedFirstIntervalRef]
+      p2 = [secondIntervalAverage, parsedSecondIntervalRef]
+    } else if (parsedFirstIntervalRef > parsedSecondIntervalRef) {
+      p1 = [secondIntervalAverage, parsedSecondIntervalRef]
+      p2 = [firstIntervalAverage, parsedFirstIntervalRef]
+    }
+
+    return { offset: linear.offset(p1, p2), slope: linear.slope(p1, p2) }
```
DanielSeehausen: the code looks great, and understanding this PR is a huge lift take this with a grain of salt:

consider breaking these out into more named functions while writing. You may find it makes the actual construction of this stuff quicker (certainly does the debugging), and it sets up future devs for success!

test/components/processviewer/CalibrationModal.test.js
```diff
@@ -46,24 +46,23 @@ describe('CalibrationModal.jsx', () => {
...
+      <Provider store={mockStore({ ...initialState, ...store })}>
+        <CalibrationModal
+          open={true}
+          stream={mockedStream}
+          closeHandler={() => {}}
+          activeProcess={mockedProcess}
+          activeDevice={mockedDevice}
+          {...props}
+        />
+      </Provider>
```
DanielSeehausen: 🔫

test/utils/Calibration.test.js
```diff
@@ -0,0 +1,72 @@
...
+        message: () => `expected ${received} not to be within range ${floor} - ${ceiling}`,
+        pass: true,
+      }
+    }
+    return {
+      message: () => `expected ${received} to be within range ${floor} - ${ceiling}`,
+      pass: false,
+    }
+  },
+})
```
DanielSeehausen: PRO

test/utils/Calibration.test.js
```diff
@@ -0,0 +1,72 @@
...
+    if (pass) {
+      return {
+        message: () => `expected ${received} not to be within range ${floor} - ${ceiling}`,
+        pass: true,
+      }
+    }
+    return {
+      message: () => `expected ${received} to be within range ${floor} - ${ceiling}`,
+      pass: false,
+    }
```
DanielSeehausen: consider favoring single return:
```js
const msg = pass 
  ? `expected ${received} not to be within range ${floor} - ${ceiling}`
  ?  `expected ${received} to be within range ${floor} - ${ceiling}`

return {
  message: () => msg,
  pass: !!pass,
}
```

DanielSeehausen: (not worth changing, just a common duplication to point out)

src/utils/calibration/Calibration.js
O-Susanoo: you can continue on the pattern you started with the getter and also have a setter

https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Functions/set

src/utils/calibration/CalibrationMode.js
```diff
@@ -0,0 +1,33 @@
...
+
+class CalibrationMode {
+  static getAverage = (rawData, start, end) => {
+    if (!start || !end) return
+    const slice = DataTools.valuesWithinTimeRange(rawData, start, end)
+    return DataTools.averageValues(slice)
+  }
+
+  static setPoint() {
+    return { params: null }
```
O-Susanoo: is there any reason this function is not setting Point ?

vdiezel: Yes, we have a mode ` LinearCalibration (direct input)` that does not implement a `setPoint` function because you can't select a point in this mode, but the user still could be clicking on a point in the graph. Code would blow up if a user does so without that. By extending this class, every mode has at least a dummy function being called when a user clicks around. Better solution would be probably to not attach the click handler for this mode, but this would push the work to the dev to know when I need to attach the click handler and when not when adding a mode. This way is a bit more failsafe.

test/utils/Calibration.test.js
vdiezel: done!

PR #512, opened 2020-02-07
486 allow user to delete alert from backend

### LINKED ISSUE #486: 9 alert: as a user, I can delete alerts
- [ ] I can delete single alerts
- [ ] I can delete all alerts

O-Susanoo:
We are excited to roll out a new **Alerts** Feature!

image

Users will notice a new 'Alerts' tab on the sidebar. Clicking this will take a user to a new view which presents a centralized place to see all high priority device Alerts. A user will be able to see all Alerts for every device they have access to. Using this view, a user may navigate to the device/process which raised the Alert.

Going forward, Alerts will be added for a variety of high priority, 'push worthy' events, such as:
- A physical warning re: the device (e.g. 'Device Door Open')
- A data warning re: a process (e.g. a user specified maximum/minimum temperature value is breached)
- ...and many more!

Alerts can either be explicitly dismissed by users by using the **'X'**, or automatically closed by non-user related actions, e.g.:
1. A user opens a sensitive 'Tissue Compartment X' on the device
2. A 'Tissue Compartment X is Open!' Alert is populated in the **Alerts** view
3. A user closes 'Tissue Compartment X'
4. The Alert is automatically removed

Is there an Alert worthy type of action/event that you encounter in your work routing with an OSPIN device? Reach out and we can make sure it gets populated in the **Alerts** view when it happens!

\private
close #486


src/redux/thunks/thunks.js
```diff
@@ -651,6 +651,7 @@ export function callRemoveNotificationsByTopic(userId, topic) {
 export function callDeleteUserAlert(alertId) {
   return async dispatch => {
     try {
+      nexus.event.remove(alertId)
```
vdiezel: if we are not awaiting it, we actually never know if it is a success or not - is this intentional ? :japanese_ogre:

DanielSeehausen: here is the server code:
```js
async function deleteEvent(event) {
  // ...
  return success({ id: deletedId })
}
```

```js
    const deleteResult = // ...mongo delete request
    return deleteResult.deletedCount === 1 ? eventId : null
```

src/components/alerts/Alert.jsx
```diff
@@ -0,0 +1,42 @@
...
+import { Message, Icon } from 'semantic-ui-react'
+import moment from 'moment'
+import { Link } from 'react-router-dom'
+import { titleCase } from 'title-case'
+import { sentenceCase } from 'change-case'
+
+const renderDeviceName = (deviceName, deviceId) => (
+  deviceName ? (
+    <Message.Item
+      style={{ display: 'block', marginTop: '7px' }}
```
DanielSeehausen: should this have the same `minHeight` given for the placeholder below?
can also then DRY the style

package.json
```diff
@@ -24,9 +24,11 @@
     "@types/react-router-dom": "^5.1.2",
     "aws-amplify": "^1.1.29",
     "babel-preset-react-app": "^9.0.0",
+    "change-case": "^4.1.1",
     "dateformat": "^3.0.3",
     "deep-diff": "^1.0.2",
     "moment": "^2.24.0",
+    "no-case": "^3.0.3",
```
DanielSeehausen: I see this imported, but not where it is used

src/components/alerts/Alert.jsx
```diff
@@ -0,0 +1,42 @@
...
+      <Message.Header>
+        {titleCase(sentenceCase(alert.name))}
+      </Message.Header>
+      {renderDeviceName(alert.deviceName, alert.deviceId)}
+      <span style={{ marginTop: '10px', float: 'left' }}>
+        <span>{alert.displayMsg}</span>
+      </span>
+      <span style={{ float: 'right', textAlign: 'right', marginTop: '10px' }}>
+        <span style={{ fontWeight: 'bold', display: 'block' }}>{moment(alert.createdAt).format('HH:mm:ss')}</span>
+        <span style={{ display: 'block' }}>{moment(alert.createdAt).format('DD/MM/YYYY')}</span>
```
DanielSeehausen: very nice

src/redux/thunks/thunks.js
O-Susanoo: > left a comment re: the optimistic rendering

well i guess nothing i will change than because it's already optimistic

package.json
O-Susanoo: this one is crucial for test side try to remove this one and you will see that test is failing

src/components/alerts/Alerts.jsx
```diff
@@ -0,0 +1,64 @@
...
+  devices: state.devices,
+  devicesInitialized: state.devicesInitialized,
+})
+
+const mapDispatchToProps = dispatch => (
+  {
+    dispatchGetDevices: cachedDevices => dispatch(callGetDevices(cachedDevices)),
+    dispatchDeleteUserAlert: alertId => dispatch(callDeleteUserAlert(alertId)),
+  }
+)
```
DanielSeehausen: some bracing style inconsistency snuck in here

O-Susanoo: Solved laser eye

src/components/alerts/Alerts.jsx
```diff
@@ -0,0 +1,64 @@
...
+
+      const augmentedAlerts = alerts.map(alert => {
+        const deviceName = Device.getById(devices, alert.deviceId)?.name
+        return { deviceName, ...alert }
+      })
+
+      setAlerts(augmentedAlerts)
+    }
+    loadMissingData()
+  }, [user, devices, dispatchGetDevices, devicesInitialized])
```
DanielSeehausen: code looks great in terms of comprehension
as functions like this grow, especially complex ones like `useEffect`, consider abstracting those parts whiich are logically separable into named functions to further increase reading comprehension (i.e. with an `agumentAlerts`, `sortAlerts` (those two could even used then in a single `serializeAlerts`)

DanielSeehausen: (not a request to change now, just something to noodle on for the future)

src/redux/actions/actions.js
```diff
@@ -241,3 +243,10 @@ export function addFlashMessage(payload) {
 export function deleteFlashMessage(payload) {
   return { type: ACTIONS.DELETE_FLASHMESSAGE, payload }
 }
+
+export function addUserAlert(payload) {
+  return { type: ACTIONS.ADD_USER_ALERT, payload }
+}
+export function removeUserAlertById(payload) {
+  return { type: ACTIONS.DELETE_USER_ALERT, payload }
+}
```
DanielSeehausen: now that we have finished the feature, we may have been off here when planning
Alerts are a standalone resource. While they are bundled into a user, the verbiage makes it seem like this deletes an alert specific to a user. In reality, when a user removes an alert, it will remove it for all users. 

We wouldn't want someone to call `DELETE_USER_ALERT`, then wonder why its gone for all users who also have access to the same alert. Not a request to change, but I wanted to highlight the value of proper naming as signaling to other developers.

src/redux/helper/subscribeHandlers/eventHandler.js
```diff
@@ -0,0 +1,42 @@
+import FlashMessenger from 'utils/FlashMessenger'
+import store from 'redux/store'
+import { addUserAlert, removeUserAlertById } from '../../actions/actions'
```
DanielSeehausen: we are mixing import syntax here :(
consider using same pattern that is on line 1 and 2

src/redux/helper/subscribeHandlers/eventHandler.js
```diff
@@ -0,0 +1,42 @@
...
+    eventType,
+    displayMsg,
+  } = value
+
+  if (!eventType || !displayMsg) {
+    console.warn('received event missing either type or message')
+    return
+  }
+  if (value.revokedId) {
+    handleRevokedAlert({ displayMsg, revokedId: value.revokedId })
```
DanielSeehausen: `revokedId` should probably be parsed out with the rest of the values above on lines 28 - 31

test/components/alerts/Alerts.test.jsx
```diff
@@ -0,0 +1,78 @@
...
+        getByText(titleCase(sentenceCase(alert.name)))
+      })
+    })
+
+    test('it should call deleting alert when user dismiss an alert', () => {
+      const { container } = renderAlerts(user, [device])
+      const firstAlert = container.getElementsByClassName('close')[0]
+      fireEvent.click(firstAlert)
+      expect(thunks.callDeleteUserAlert).toBeCalledTimes(1)
+    })
```
DanielSeehausen: very nice!

test/components/alerts/Alerts.test.jsx
```diff
@@ -0,0 +1,78 @@
...
+
+    test('it should redirect to device when user click on device name', () => {
+      const { getAllByText } = renderAlerts(user, [device])
+      const firstAlert = getAllByText(device.name)[0]
+      fireEvent.click(firstAlert)
+      expect(pushMock).toHaveBeenCalledTimes(1)
+      expect(pushMock).toHaveBeenCalledWith(`/device/${device.id}`)
+    })
```
DanielSeehausen: awesome also

O-Susanoo:
> in the `package.json` you might be importing `change-case`, then two of its dependencies.
> 
> https://github.com/blakeembrey/change-case

there is nothing you can do about this one here in fact the package change-case dose not have title case 
```
export * from "camel-case";
export * from "capital-case";
export * from "constant-case";
export * from "dot-case";
export * from "header-case";
export * from "no-case";
export * from "param-case";
export * from "pascal-case";
export * from "path-case";
export * from "sentence-case";
export * from "snake-case";
```

, and even tough the no case is indeed in the change case package but it's still needed as separate one when testing because it seem that sentence case require it from local path for some reason i end up removing change-case and installing only sentence-case

test/components/alerts/Alerts.test.jsx
```diff
@@ -0,0 +1,78 @@
...
+      fireEvent.click(firstAlert)
+      expect(thunks.callDeleteUserAlert).toBeCalledTimes(1)
+    })
+
+    test('it should redirect to device when user click on device name', () => {
+      const { getAllByText } = renderAlerts(user, [device])
+      const firstAlert = getAllByText(device.name)[0]
+      fireEvent.click(firstAlert)
+      expect(pushMock).toHaveBeenCalledTimes(1)
+      expect(pushMock).toHaveBeenCalledWith(`/device/${device.id}`)
```
DanielSeehausen: when using mock like this, you want to make sure to reset/clear all mocks in an after/before all so tests are extendable without breaking/unexpected behavior.

Alternatively, and maybe even a better pattern that doesn't require doing work before it is needed, simply bring the mock calls down into this test function, as they are only used once. Then, you make the burden of abstracting that obvious for the next comer

test/helpers/UserGenerator.js
```diff
@@ -0,0 +1,22 @@
...
+      userName: faker.internet.userName(),
+      fullName: faker.name.findName(),
+      organization: 'ospin',
+      email: faker.internet.email(),
+      role: 'admin',
+      notifications: [],
+      events: [],
+      ...preMapData,
+    }
+    return { ...createUserObject(userData), ...postMapData }
```
DanielSeehausen: what is the use case for the `postMapData`? It is somewhat intimidating, seeing as `createUserObject` is a control point before entry into our application. Is there a concern that `createUserObject` will overwrite data that we want? It doesn't look like it currently does, and we shouldn't be mutating a user beyond their schema (which `createUserObject` is used to assert)

If I am missing something please let me know here, otherwise, I would lobby to remove it

test/redux/helper/subscribe.test.js
```diff
@@ -0,0 +1,57 @@
...
+      eventHandler({ value: newAlert })
+      const action = mockedDispatch.mock.calls[1][0]
+      expect(mockedDispatch).toBeCalledTimes(2)
+      expect(action).toEqual(expect.objectContaining({
+        type: ACTIONS.ADD_USER_ALERT,
+        payload: newAlert,
+      }))
+    })
+  })
+})
```
DanielSeehausen: very exciting that we have some test coverage here now!

DanielSeehausen:
@O-Susanoo left some comments in the review! I might not have understood the use case for some of the things I left comments on. Please let me know if so!

looking very very good

test/helpers/UserGenerator.js
O-Susanoo: well i followed viktor pattern please check with him to see why he did had this stuff maybe he has some use cases that may require such case

src/components/alerts/Alert.jsx
O-Susanoo: it's already the default , if i do the dry stuff then i will break the implicit return and it's such a beauty i didn't want to break it

src/redux/actions/actions.js
O-Susanoo: i guess ADD_LOGGED_IN_USER_ALERT and DELETE_LOGGED_IN_USER_ALERT ? if it seem good to you i can change it to this

DanielSeehausen: the names are good how they are - I meant to signal less detail, to communicate that they are generic and not user specific

src/components/alerts/Alert.jsx
DanielSeehausen: you can declare a style CONSTANT outside of the method :p

src/redux/thunks/thunks.js
vdiezel: When we do optimistic rendering then you also have to implement rollbacks in case the server request fails ... so I'd be in favor of not going optimistic here, unless failing deleting an event doesn't have any important implications.

test/redux/helper/subscribe.test.js
```diff
@@ -0,0 +1,57 @@
...
+    test('it should fire new alert', () => {
+      const newAlert = alerts[0]
+      eventHandler({ value: newAlert })
+      const action = mockedDispatch.mock.calls[1][0]
+      expect(mockedDispatch).toBeCalledTimes(2)
+      expect(action).toEqual(expect.objectContaining({
+        type: ACTIONS.ADD_USER_ALERT,
+        payload: newAlert,
+      }))
+    })
```
vdiezel: Can we agrre on a couple of white lines, just for readbility, something that seperates into setup - executtion - assertion ?

test/mocks/alertsMock.js
```diff
@@ -0,0 +1,31 @@
...
+export const revokedAlert = {
+  id: '7239249e-a7c0-44e4-b79e-4a87db56d5a3',
+  deviceId: '9ef4f0b3-977b-4d36-9760-bba0b789bb0c',
+  eventType: 'alarm',
+  name: 'doorClose',
+  createdAt: 1580903583372,
+  processId: 'de13009a-532d-4238-9c55-b4efc67e4fb8',
+  displayMsg: 'alert display message 2',
+  revokedId: '7239249e-a7c0-44e4-b79e-4a87db56d5b3',
+}
```
vdiezel: We have a collection of Generators for test data - would be a good oppurtunity to create a generator for alerts, too.

src/redux/thunks/thunks.js
O-Susanoo: well from the look of it if we fail to delete then the worst case scenario is that next time user refresh it's page it will see old alerts ? i can change that i had already changed that but since Daniel requested an optimistic change i thought  you guys had already went trough it , but i don't have problems with any case

test/redux/helper/subscribe.test.js
O-Susanoo: great observation i will do that

test/mocks/alertsMock.js
O-Susanoo: you are right i was a lazy on this one but i followed the github pattern , i will change that

PR #515, opened 2020-02-10
501 move page navigation controls to header

### LINKED ISSUE #501: Move the logging page navigation + results per page footer to a header
image

as a footer, it bounces around depending on row size, and a user would need to scroll down to get to it. 

Separate it from the 'load more' and move it to a header.

vdiezel:
closes #501 

Can be reviewed commit wise!

there is no way to keep the controls in the Table when we want to move them top, so I made a seperate menu that float at the top. Added the menu to the process table view and the log view:

Bildschirmfoto von 2020-02-10 14-04-12
Bildschirmfoto von 2020-02-10 14-42-16
Bildschirmfoto von 2020-02-10 14-44-18


src/components/device/processes/ProcessTableControls.jsx
```diff
@@ -0,0 +1,92 @@
...
+            onChange={updateResultsPerPage}
+          />
+        </span>
+      </Menu.Item>
+    </Menu.Menu>
+    <Menu.Menu position='right'>
+      <Menu.Item>
+        <span>
+          Select action:
+          {' '}
```
DanielSeehausen: first time recognizing this being used. very cool.

DanielSeehausen:
images looking great too:
image

what does this end up looking like when there are more paged?
image

vdiezel:
> images looking great too:
> image
> 
> what does this end up looking like when there are more paged?
> image

https://react.semantic-ui.com/addons/pagination/

PR #523, opened 2020-02-11
WIP Remove stacktrace and replace by error message

### LINKED ISSUE #476: Make alerted stack trace readable or remove the unreadable stack trace
our stack traces aren't usable because of the bundled application
this means we can't make use of the really great tool we have spent time creating!

find a way to make the stack traces readable, or remove the noise and provide any other potentially helpful information.

[libraries like this or others similar may be of help](https://www.stacktracejs.com/)

if we can not improve this, remove the stack trace and just keep the top level information

O-Susanoo:
(now WIP)

closes #476 

Here is a link for the cra response to this issue 

[CRA Issue](https://github.com/facebook/create-react-app/issues/3753)

i have kept the error message tough because we need to be able to distinguish visually between a page that have an error and just the report page

i will also open a ticket for backend to remove the stack trace from hambda


O-Susanoo:
Sadly daniel the module you provided doesn't help also , however thanks to you , i was forced to push a little bit the envelope and this picture describe my findings

image

i found that the error we receive is a pretty heavy object but when we took string of it i was only getting the non important part

the module you provided has a pretty cool function that can extract those information from the error in a very nice way , and that's what you guys in the second line

i decided that i will keep stuff as it and the stack trace will consist of the function name and line number which is pretty good also

DanielSeehausen:
looks good - guessing htis is going tog et an update though after what you posted in slack?

O-Susanoo:
I have pushed the code , i spent the entire day trying to fix those test but for some reasons it doesn't work, individually test work just fine but in group they hang , so i have commented some of them , hopefully i will find solution in other time , but i will for now leave it like this

P.S : i tried to run those test with -i and even with that it doesn't help , i think that there is something wrong with the function being mocked in nexus , i will try later to also maybe mock the module instead of the function itself

DanielSeehausen:
as discussed, ill pick it up! thanks mo.
these tests can be a pain sometimes

DanielSeehausen:
closing in preference of #549 because this git history is bonked

PR #526, opened 2020-02-12
implements the loading indicator when fetching data for graphs

### LINKED ISSUE #507: Create a loading indicator for graphs when they try to fetch data
Currently, a graph and its controls display a 'No-Data' message when it's awaiting data to be fetched from the API. When there is simply no data, the graph displays the same thing. We should distinguish the the two state of data being loaded from the is simply no data for this upstream. best solution would be a loading indicator when the api tries to fetch data for the graph - but I don't now if that feasible because the process data loading is triggered by the mounting of the graphs but burried somewhere in the thunks.

vdiezel:
closes #507 

We now have different display for when data is being fetched or when there is simply no data existing for the stream

old (you see 'No data' for a fraction of a second, webapp to friggin fast!)
old_graph
new:
new_graph


src/utils/process.js
```diff
@@ -188,6 +188,10 @@ export function getStreamDataPoints(process, streamId) {
   return process.data.find(streamData => streamData.id === streamId).dataPoints || []
 }
 
+export function isStreamDataLoaded(process, streamId) {
+  return process.data.find(streamData => streamData.id === streamId).loaded
+}
+
```
DanielSeehausen: gblame tells me you made the Process utils class for new stuff that goes in here

in you'd rather this there, I wanted to point it out:
image

DanielSeehausen:
very nice! love the gifs

PR #532, opened 2020-02-14
Foldable Graphs!

### LINKED ISSUE #444: GRAPHS 1 UIConfig: As a user, I can collapse/fold and expand graphs
As a user, I can fold and expand graphs in the same way process builder folds and expands controllers/streams. This is the visual MVP, and is part of a larger concern regarding storing UI config.

- [x] As a first step, please provide a proposal (via an image here) of where the actual UI changes required to complete the below features will take place.
- [x] A LOT of work is going to go into these graph components, so the first step is to look through them, and see what refactoring can be done right off the bat to make sure this code is easy to change going forward and that it can be worked on in the future. Keep a special eye out for components that are too big and can be broken down into smaller components.

- [ ] provide minimal test coverage that asserts folding/unfolding works
- [ ] add button for 'Fold All' that, when clicked, folds all graphs
- [ ] add button for 'Expand All' that, when clicked, expands all graphs
- [ ] ensure single graphs can be folded/unfolded. 'Fold All' and 'Expand All' should still work. 
- [ ] implement and start tracking local (react side) state schema to keep track of stream/graph UI (*DISCUSS* )
- [ ] provide minimal test coverage that asserts graph ui/config data determines whether the graphs render as folded/unfolded

These deliverables may come out cleaner and easier if completed in the above order.

reference for where we are already doing this fold/unfold general behavior:
image

### LINKED ISSUE #539: GRAPHS 0a UIConfig add config object to the state and connect to the correct component

O-Susanoo:
Peek 2020-02-21 12-59

As part of our ongoing work to streamline how users interact with process data. We are excited to role out the first feature in the series. Data streams, as well as their encompassing groups, can now be collapsed.

As a first step, these changes will only be persisted _per session_ (hard refreshing your browser will undo any custom folding/unfolding you do).

Upcoming:
- Groups and streams are user arrangeable via click and drag
- Configurations persist under all circumstances, on all devices

\private
closes #444 
closes #539


DanielSeehausen:
ill remove the codevoc stuff until i get it working properly

DanielSeehausen:
( i updated the branch through UI since it was :clean: )

DanielSeehausen:
looks like codecov is working now, which is cool!
only request is some more test coverage. The coverage requirement is woefully small right now (50%). I would take a moment to look at what the easiest, cheapest test you can make is that covers the largest % of what you put up. Should be very easy to hit 50%!

src/components/processviewer/body/ProcessViewerBody.jsx
```diff
@@ -44,44 +41,52 @@ const ProcessViewerBody = ({
...
+                activeDevice={activeDevice}
+              />
+            </Segment>
+          )}
+          {activeTab === 'Logs' && (
+            <Segment attached='bottom'>
+              <LogView
+                activeDevice={activeDevice}
+                defaultQuery={{ processId: activeProcess.id }}
+                logs={Logs.getProcessLogs(logs, activeProcess.id)}
```
vdiezel: `logs` are now undefined, because you disconnected the component from the state (and they are not passed down from the parent component). `LogView` needs them however.

src/components/processviewer/body/dataviewer/DataGraphs.jsx
```diff
@@ -21,37 +33,71 @@ class DataGraphs extends React.Component {
...
+                  toggleStreamFold={streamId => this.toggleStreamFolded(group.name, streamId)}
+                  activeProcess={activeProcess}
+                  activeDevice={activeDevice}
+                />
               ))}
-            </Grid>
+            </Accordion>
           )}
       </Container>
     )
```
vdiezel: can we do an early return when `activeProcess.state` is `executable` instead of using the ternary? :japanese_ogre:

src/components/processviewer/body/dataviewer/Stream.jsx
```diff
@@ -99,59 +101,83 @@ export default class Stream extends React.Component {
...
-            )}
-        </Grid.Column>
-      </Grid.Row>
+      <>
+        <Accordion.Title
+          active={!folded}
+          content={stream.name}
+          onClick={toggleFold}
+        />
+        { folded ? null : (
```
vdiezel: would `folded && ( <Accordion...)` be more readable than a ternary?

src/utils/process.js
```diff
@@ -682,4 +682,10 @@ export class Process {
   static isStreamDataLoaded(process, streamId) {
     return process.data.find(streamData => streamData.id === streamId).loaded
   }
+
+  static getProcessUpStream(process, streamId) {
```
vdiezel: this function should be really `getProcessStreamById` - all streams have unique Ids, and when you use the Id to fetch the stream, you don't need to distinguish between up- and downstreams anymore.

src/components/processviewer/body/ProcessViewerBody.jsx
O-Susanoo: Good eye , thanks i guess that would have happened in one of daniel refactoring , i have changed it

test/components/processviewer/body/dataviewer/DataGraphs.test.jsx
```diff
@@ -0,0 +1,115 @@
...
+
+      const { getByText } = renderDataGraphs(mockedDevice, streams, [uIConfigs], runningProcess)
+      const toggleGroup = getByText(firstGroup)
+
+      oldConfig.configs.graphs.groups[0].folded = !oldConfig.configs.graphs.groups[0].folded
+
+      thunks.callSetDeviceUIConfig = jest.fn()
+      fireEvent.click(toggleGroup)
+      expect(thunks.callSetDeviceUIConfig).toBeCalledTimes(1)
+      expect(thunks.callSetDeviceUIConfig).toBeCalledWith(mockedDevice.id, oldConfig)
```
vdiezel: is it possible to assert that the group shows up in the UI as expected? If it can be done without effort, I'd advocate to do this.

src/components/processviewer/body/dataviewer/DataGraphs.jsx
O-Susanoo: but if you do so you will not be able to view the No data yet  string is that ok ?

src/components/processviewer/body/dataviewer/Stream.jsx
O-Susanoo: we actually talked about this with daniel but since you are a cool guy i have made this change also

test/components/processviewer/body/dataviewer/DataGraphs.test.jsx
O-Susanoo: i don't think so viktor because if you do so then you need to handle redux , because the ui is just what the configuration is and the moment you don't mock thunks you start getting weird stuff , unless you know something i don't

DanielSeehausen:
> Love this. Reads like a masterpiece of literature. I left my critique here and there.
> I have one request in `ProcessViewerBody` because I don't understand why the component was disconnected from the store, leaving the `logs` passed to `LogView` to be undefined. Am I missing something?
> 
> Once this is clear and we get the minor visual changes for better distinction between groups and streams, this is gtbm

I gotta apologize because I very well may have done this without noticing.

src/components/processviewer/body/dataviewer/DataGraphs.jsx
vdiezel: if we do an early return like this?
```javascript 
if (activeProcess.state === 'executable') return <Segment> No data yet. </Segment>
```
if this looks weird because it misses the wrapping container, you could also wrap the sehment with it there.

src/components/processviewer/body/dataviewer/DataGraphs.jsx
```diff
@@ -21,37 +33,71 @@ class DataGraphs extends React.Component {
...
+    const newConfig = UIConfig.createDeviceUIConfig(activeDevice.id, userId, processUpStreams)
+    const mergedConfig = UIConfig.merge(oldConfig, newConfig)
+
+    const configChanged = !UIConfig.graphsAreEqual(oldConfig, mergedConfig)
+    const configNew = !UIConfig.deviceUIConfigExists(uIConfigs, activeDevice.id)
 
-    const groups = getGroups(processUpStreams)
+    if (configChanged || configNew) {
+      dispatchSetDeviceUIConfig(activeDevice.id, mergedConfig)
+    }
```
DanielSeehausen: While verbose, it reads directly and easily. Love how it came out.

src/components/processviewer/body/dataviewer/Stream.jsx
DanielSeehausen: 👍 
I am + on short circuit syntax. 
It is a legacy thing that ternaries were used - react nowadays handles null and undefined in renders the same (or close enough to the same as not to matter). In the past, it wanted explicit `null`

src/utils/process.js
DanielSeehausen: 👍 second generic get method. If you do change, consider using `find` over get, because work is done, which `find` denotes!

src/components/processviewer/ProcessViewer.jsx
```diff
@@ -14,21 +10,23 @@ import {
...
 import { updateProcess, loadProcessViewer } from 'redux/actions/actions'
+import PhasesNavbarWrapper from './phasesdisplaybar/PhasesNavbarWrapper'
+import ProcessBuilderHeader from './headers/ProcessBuilderHeader'
+import DescriptionMismatchModal from './modals/DescriptionMismatchModal'
+import ProcessViewerHeader from './headers/ProcessViewerHeader'
 
-const mapStateToProps = state => ({
-  processBuilderValidationErrors: state.processBuilderValidationErrors
-})
+const mapStateToProps = state => ({ processBuilderValidationErrors: state.processBuilderValidationErrors, logs: state.logs })
```
O-Susanoo: this is the most important part everything else is just some minor eslint stuff

src/components/processviewer/ProcessViewer.jsx
```diff
@@ -144,12 +142,13 @@ class ProcessViewer extends React.Component {
...
-              phaseIds={getPhaseIds(this.props.activeProcess)}
-              activeProcess={this.props.activeProcess}
-              activeDevice={this.props.activeDevice}
+              processDownStreams={getProcessDownStreams(activeProcess)}
+              processUpStreams={getProcessUpStreams(activeProcess)}
+              phases={getPhasesArray(activeProcess)}
+              phaseIds={getPhaseIds(activeProcess)}
+              activeProcess={activeProcess}
+              activeDevice={activeDevice}
+              logs={logs}
```
O-Susanoo: here we give logs to the ProcessViewerBody

test/components/processviewer/body/dataviewer/DataGraphs.test.jsx
```diff
@@ -0,0 +1,115 @@
...
+
+      const streamNames = streams.map(stream => stream.name)
+      const groupNames = [...new Set(streams.map(stream => stream.groupName))]
+      const { getByText, queryByText } = renderDataGraphs(mockedDevice, streams, [], runningProcess)
+
+      groupNames.forEach(getByText)
+
+      streamNames.forEach(name => {
+        expect(queryByText(name)).toBeNull()
+      })
```
DanielSeehausen: very very nice. This is some react testing library down the middle, straight shooting, high value, low effort, testing.

test/components/processviewer/body/dataviewer/DataGraphs.test.jsx
```diff
@@ -0,0 +1,115 @@
...
+
+      const groupName = uIConfigs.configs.graphs.groups[0].name
+      const streamNames = streams
+        .filter(stream => stream.groupName === groupName)
+        .map(stream => stream.name)
+
+      const { queryByText } = renderDataGraphs(mockedDevice, streams, [uIConfigs], runningProcess)
+      streamNames.forEach(name => {
+        expect(queryByText(name)).toBeTruthy()
+      })
```
DanielSeehausen: you could add the same group names assertion you have in the above test

test/components/processviewer/body/dataviewer/DataGraphs.test.jsx
DanielSeehausen: this test may be misleading - it says it asserts that a group is unfolded when a user click on it, but then it asserts something different (a method in thunks is called). 

The test itself is good, but the description could use a re-work.

DanielSeehausen:
looks like tests failed. If this is a flicker (as in, sometimes a test is green, sometimes its red, even though nothing changes in the code), it might be passed by simply re-running the suite in circle CI using that button at the top right. 

If this is a flicker, it needs to be documented somewhere and a ticket made up for it, because test flickers are silent killers, like high blood pressure. 

image

PR #544, opened 2020-02-24
Collapse all graphs

### LINKED ISSUE #543: Add fold all button + change style group color

O-Susanoo:
Peek 2020-02-24 13-03

Data streams, as well as their encompassing groups, can now be collapsed all with a single click .these changes will only be persisted per session (hard refreshing your browser will undo any custom folding/unfolding you do)
 

closes #543


DanielSeehausen:
I took the UI label off
The only labels we use for PRs at the moment are changelog and patch!

src/components/processviewer/body/dataviewer/DataGraphs.jsx
```diff
@@ -10,9 +10,9 @@ const mapStateToProps = state => ({
   userId: state.user.id,
 })
 
-const mapDispatchToProps = dispatch => ({
-  dispatchSetDeviceUIConfig: (deviceId, newDeviceUiConfig) => dispatch(callSetDeviceUIConfig(deviceId, newDeviceUiConfig))
-})
+const mapDispatchToProps = dispatch => (
+  { dispatchSetDeviceUIConfig: (deviceId, newDeviceUiConfig) => dispatch(callSetDeviceUIConfig(deviceId, newDeviceUiConfig)) }
+)
```
DanielSeehausen: while a really small nitpick i'm sorry:

we tend to `({` everywhere in the code base instead of a new line
this is how implicit function returns tend to be styled as well. Unsure if your linter did this, but it looks like in 95% of the cases we are doing it the way this was changed from.

src/components/processviewer/body/dataviewer/DataGraphs.jsx
```diff
@@ -47,6 +47,12 @@ class DataGraphs extends React.Component {
     dispatchSetDeviceUIConfig(activeDevice.id, newConfig)
   }
 
+  foldall = () => {
+    const { activeDevice, uIConfigs, dispatchSetDeviceUIConfig } = this.props
+    const newConfig = UIConfig.foldAllGraphs(uIConfigs, activeDevice.id)
+
+    dispatchSetDeviceUIConfig(activeDevice.id, newConfig)
+  }
```
DanielSeehausen: very nice

src/utils/UIConfig/GraphsUIConfig.js
```diff
@@ -67,5 +67,14 @@ export default class GraphsUIConfig {
...
 
+  static foldAll(graphs) {
+    graphs.groups.forEach(group => {
+      group.folded = true
+      group.streams.forEach(stream => {
+        stream.folded = true
+      })
+    })
+    return graphs
+  }
```
DanielSeehausen: this looks great for the use case.

if you wanted to use the interfaces that are growing already and be defensive programming re: the future (but this likely won't change much after this, so not worth it), streams should only be manipulated through their interface (`StreamsUIConfig`). If this were to grow, the likelihood that streams need to be folded in other calls jumps up, and then we have duplicate ways of doing the same thing, which hurts maintainability. I suppose by the same logic we should have made a `Groups` UI, but w/e

test/components/processviewer/body/dataviewer/DataGraphs.test.jsx
```diff
@@ -13,6 +13,9 @@ import { getMockedProcessData } from '../../../../mocks/processDataMock'
 const mockStore = configureMockStore()
 
 describe('DataGraphs.jsx', () => {
+  const mockedCallSetDeviceUIConfig = jest.fn()
+  thunks.callSetDeviceUIConfig = mockedCallSetDeviceUIConfig
```
DanielSeehausen: if you wrap your mock up in a `beforeAll` (see below)

test/components/processviewer/body/dataviewer/DataGraphs.test.jsx
```diff
@@ -106,10 +113,46 @@ describe('DataGraphs.jsx', () => {
 
       oldConfig.configs.graphs.groups[0].folded = !oldConfig.configs.graphs.groups[0].folded
 
-      thunks.callSetDeviceUIConfig = jest.fn()
       fireEvent.click(toggleGroup)
-      expect(thunks.callSetDeviceUIConfig).toBeCalledTimes(1)
-      expect(thunks.callSetDeviceUIConfig).toBeCalledWith(mockedDevice.id, oldConfig)
+      expect(mockedCallSetDeviceUIConfig).toBeCalledTimes(1)
+      expect(mockedCallSetDeviceUIConfig).toBeCalledWith(mockedDevice.id, oldConfig)
```
DanielSeehausen: you can assert on the call itself:
`expect(thunks.callSetDeviceUIConfig)...`

without carrying around the extra mock!

test/components/processviewer/body/dataviewer/DataGraphs.test.jsx
```diff
@@ -106,10 +113,46 @@ describe('DataGraphs.jsx', () => {
...
+    })
+
+    test('everything should be folded when user click collapse all button', async () => {
+      const mockedDevice = createDeviceObject(getMockedDeviceData({ streams, state: 'running' }))
```
DanielSeehausen: the state of `folded: false` should probably be explicitly set - otherwise we can get false passes on this in the future :(

test/components/processviewer/body/dataviewer/DataGraphs.test.jsx
```diff
@@ -106,10 +113,46 @@ describe('DataGraphs.jsx', () => {
...
 
-      thunks.callSetDeviceUIConfig = jest.fn()
       fireEvent.click(toggleGroup)
-      expect(thunks.callSetDeviceUIConfig).toBeCalledTimes(1)
-      expect(thunks.callSetDeviceUIConfig).toBeCalledWith(mockedDevice.id, oldConfig)
+      expect(mockedCallSetDeviceUIConfig).toBeCalledTimes(1)
+      expect(mockedCallSetDeviceUIConfig).toBeCalledWith(mockedDevice.id, oldConfig)
+    })
+
+    test('everything should be folded when user click collapse all button', async () => {
```
DanielSeehausen: this test appears to be a UI test (being on the datagraphs component), and the description reads like a UI test, but then the test checks for something different :( 

ideally, test readouts can be read like documentation so lazy programmers don't need to read implementation

test/utils/UIConfig/UIConfig.test.js
```diff
@@ -210,6 +210,24 @@ describe('UIConfig', () => {
       })
     })
 
-  })
+    describe('When user click on collapse all', () => {
```
DanielSeehausen: this is a test on the data interface - no user testing going on here (this guy knows nothing of clicking or meatbag users)

test/components/processviewer/body/dataviewer/DataGraphs.test.jsx
O-Susanoo: I know but i made it like this in case in future someone will want to have the real function back

test/components/processviewer/body/dataviewer/DataGraphs.test.jsx
O-Susanoo: Yeah it's gets a little bit tricky when testing with redux you are right i have fixed the test thanks for the review

PR #553, opened 2020-02-28
Setting Stream Event Triggers in Process Builder and Device Setup

### LINKED ISSUE #527: create a view in the process builder to modify min and max values for down streams which should trigger an alert (for controllers only which have a downstream with an associated upstream)

### LINKED ISSUE #528: create a view in the device setup where user can create device defaults for min and max for alerts for streams

### LINKED ISSUE #529: implement the new min and max properties into the existing 'config' property on the device and inject them into the process deviceDescription on creation

### LINKED ISSUE #546: Fork rendering of alarms when they are stream specific
stream specific events contain only a default message that is not adjusted to the custom defined units and stream names. Therefore we send the raw data of the event that triggered the alarm and the message needs to be created on the frontend from it.

vdiezel:
*THIS FEATURE WILL ONLY BE SHIPPED ONCE WE GET THE LINKS BETWEEN DOWNSTREAMS AND UPSTREAMS. CAN BE REVIEW ALREADY THOUGH. CHANGES WILL ONLY AFFECT WHICH STREAMS ALLOW SETTINGS TRIGGERS.*

closes #546 
closes #527 
closes #528 
closes #529 

Implements the creation of event triggers on on the process and on the device (device triggers are applied automatically on the process when created)

Display of the alert button will only happen when the downstream has a linked upstream - red when a trigger is set. The trigger value of 12K will be rejected when a user attempts to start the process because it is below the target for an upper limit trigger.
triggers_process

View in the setup section of the device (all streams are selectable there, but I will change this once firmware catches up)
triggers_device


src/components/alerts/Alert.jsx
```diff
@@ -39,4 +95,4 @@ const Alert = ({ alert, deleteAlert }) => (
   </Message>
 )
 
-export default Alert
+export default connect(mapStateToProps)(Alert)
```
vdiezel: changes here to accomodate for alerts that are stream related. We decided in this case to send the streamId and the raw data to be able to create a meaningful message that use the user defined stream names and units. Also having raw data emitted in the event can be stored in the logs and parsed easier later.

src/components/device/modals/StreamEventTriggerModal/StreamEventTriggerModal.js
```diff
@@ -0,0 +1,364 @@
+import React from 'react'
```
vdiezel: The main component - I'm resusing same modal for setting triggers in the process and as default on the device (dervied from the provided `mode`)

src/components/alerts/Alert.jsx
O-Susanoo: even tough the parent already have devices but i think It's good thing you didn't forward them to this component

src/components/device/modals/StreamEventTriggerModal/ParameterMenu.jsx
```diff
@@ -0,0 +1,66 @@
...
+          <Form.Group widths='equal' key={`${targetStreamId}-${parameter.name}`}>
+            <Form.Field>
+              <label>
+                Current {parameter.label} [{displayedUnit}]
+              </label>
+              <Input
+                fluid
+                type='text'
+                readOnly
+                value={currentEventTrigger ? displayedParam : 'None'}
```
O-Susanoo: Do we want to display None or is it better to display empty string

src/components/device/modals/StreamEventTriggerModal/StreamEventTriggerModal.js
```diff
@@ -0,0 +1,377 @@
...
```
O-Susanoo: This file should be renamed to a jsx file because you are using react components

src/components/device/modals/StreamEventTriggerModal/StreamEventTriggerModal.js
```diff
@@ -0,0 +1,377 @@
...
+    [name]: value,
+    isError: false,
+    errorMessage: '',
+    isSuccess: false,
+  })
+
+  parseByType = (type, value) => {
+    switch (type) {
+      case 'number':
+        return value !== '' ? Number(value) : () => NaN
```
O-Susanoo: didn't understand this line in one case you are returning a number in the other case you are returning a function ?

src/components/device/modals/StreamEventTriggerModal/StreamEventTriggerModal.js
```diff
@@ -0,0 +1,377 @@
...
+        <Modal.Header>
+          <Icon name='edit'/>
+          Create Stream Event Trigger
+        </Modal.Header>
+        <Modal.Content>
+          <Modal.Description>
+            <Form
+              error={isError}
+              success={isSuccess}
+              onSubmit={() => this.submit()}
```
O-Susanoo: You already have this function , there is not need to create an arrow function in this line

src/components/device/modals/StreamEventTriggerModal/StreamEventTriggerModal.js
```diff
@@ -0,0 +1,377 @@
...
+                header='Something went wrong.'
+                content={errorMessage}
+                icon='warning circle'
+              />
+              <Message
+                success
+                header='Success!'
+                content={<p>{successMsg}</p>}
+                icon='warning circle'
+              />
```
O-Susanoo: I think it's weird that those messages are on the doom even there is nothing , i think it's better if they are only shown when they are needed

src/components/device/modals/StreamEventTriggerModal/ParameterMenu.jsx
vdiezel: We stick with `None` for now I guess.

src/components/device/modals/StreamEventTriggerModal/StreamEventTriggerModal.js
vdiezel: You are right, don't know why I made this a function. Fixed.

src/components/device/modals/StreamEventTriggerModal/StreamEventTriggerModal.js
vdiezel: yep!

src/components/device/setup/StreamEventTriggerTable.jsx
```diff
@@ -0,0 +1,122 @@
...
+    const configuredStreams = DeviceConfig.getConfiguredStreamsWithEventTriggers(activeDevice)
+
+    return (
+      <>
+        <Button
+          key={0}
+          primary
+          icon
+          labelPosition='left'
+          onClick={() => this.toggleStreamEventTriggerModal()}
```
O-Susanoo: no need for an arrow function here you can call directly the function

src/components/device/setup/StreamEventTriggerTable.jsx
```diff
@@ -0,0 +1,122 @@
...
+        >
+          <Icon name='edit'/>
+          Set Stream Event Trigger
+        </Button>
+        {configuredStreams.length !== 0 && <Table key={1} className='ospin-red' celled collapsing>
```
O-Susanoo: same question about the key attribute ?

src/components/device/setup/StreamEventTriggerTable.jsx
```diff
@@ -0,0 +1,122 @@
...
+                      position='right center'
+                    />
+                  </Table.Cell>
+                </Table.Row>
+              ))
+            ))}
+          </Table.Body>
+        </Table>}
+        <StreamEventTriggerModal
+          closeHandler={() => this.toggleStreamEventTriggerModal()}
```
O-Susanoo: no need for arrow function

src/components/device/setup/StreamEventTriggerTable.jsx
```diff
@@ -0,0 +1,122 @@
...
+                    {EventTrigger.description(eventTrigger.type)}
+                  </Table.Cell>
+                  <Table.Cell>
+                    {EventTrigger.displayParameters(eventTrigger)}
+                  </Table.Cell>
+                  <Table.Cell>
+                    {!stream.connected ? 'stream not connected' : 'Ok'}
+                  </Table.Cell>
+                  <Table.Cell>
+                    <DeletePopup
```
O-Susanoo: Nice we are using the DeletePopup you were right this component is really handy

src/components/device/setup/StreamEventTriggerTable.jsx
```diff
@@ -0,0 +1,122 @@
...
+    const { activeDevice, dispatchCallUpdateDevice } = this.props
+
+    const updatedEventTriggers = DeviceConfig
+      .removeEventTriggerByType(activeDevice, streamId, type)
+
+    const updatedConfig = DeviceConfig
+      .updateStreamConfig(activeDevice, streamId, { eventTriggers: updatedEventTriggers })
+
+    const updateParams = { config: updatedConfig }
+    await dispatchCallUpdateDevice(activeDevice.id, updateParams)
```
O-Susanoo: do we really need to await here ?

src/utils/EventTrigger/EventTrigger.js
```diff
@@ -0,0 +1,47 @@
...
+  static directToType = type => {
+    switch (type) {
+      case this.TYPES.UPPER_LIMIT:
+        return UpperLimitTrigger
+      case this.TYPES.LOWER_LIMIT:
+        return LowerLimitTrigger
+      default:
+        throw new Error(`Unknown event trigger type ${type} provided.`)
+    }
+  }
```
O-Susanoo: I think that all the triggers have a common api ? if it's the case i think you could have declared an abstract class and then made the LowerLimitTrigger.js and UpperLimitTrigger.js extend from this class , so that if for some reason we want to add a complexe trigger we will just need to implement the abstract class memebers

O-Susanoo:
I'm still going through all of theme i think i may be done with that tomorrow

src/components/processviewer/body/builder/ChangesFooter.jsx
```diff
@@ -53,10 +53,7 @@ const ChangesFooter = ({
 
     const updateParams = { description: updatedProcess.description }
     await dispatchDispatchUpdateProcess(updatedProcess.id, updateParams)
-
-    if (PDVC.runtimeRelevantChanges(diffs)) {
```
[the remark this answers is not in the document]
vdiezel: yes, PDVC now returns only runtime relevant changes. Changes like renaming a phase or setting event triggers are not considered 'Ad hoc' because they have no effect on the process that is running on the device.

src/components/device/setup/StreamEventTriggerTable.jsx
```diff
@@ -0,0 +1,122 @@
...
+
+    const { activeDevice, user } = this.props
+    const { showStreamTriggerModal } = this.state
+
+    const configuredStreams = DeviceConfig.getConfiguredStreamsWithEventTriggers(activeDevice)
+
+    return (
+      <>
+        <Button
+          key={0}
```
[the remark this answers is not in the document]
vdiezel: if I recall correctly it's because there are two siblings within this fragment and React complained about this. But this has been done ages ago, I can check.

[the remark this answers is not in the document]
vdiezel: You were right, it doesn't need them anymore.

src/utils/EventTrigger/EventTrigger.js
vdiezel: Yes, I did exactly that on the backend side. In this case I didn't because even though they have the same functions, they don't share the behaviour of a single one, so there was no value in creating an abstract class.

PR #574, opened 2020-03-06
add stream util with image resolver

### LINKED ISSUE #520: Dashboard icons are dynamic based on stream data
- [ ] Follow up with Jordi Or Felix Or Jan about getting these SVG files!
- [ ] dashboard icons are dynamically chosen based on the upstream type that is communicated from the device. If none is communicated, the icon defaults to what it is now

DanielSeehausen:
- the device dashboard will now have dynamic image icons based on stream type

\private
closes #520 
can be reviewed commit wise

I believe the only thing that will need to be done is the actual images being imported and replaced in the map

The upstream data objects seem to be going straight through to these components (see image), so I image they will show up with their `dataType`s on them once firmware flashes a device!

This is backwards compatible, so it could be merged and the images added later (but we might as well wait until we have at least two images!)

Here is an example of the upstream data object. Imagine it has a `dataType` on it that says `theDude`
image

image


src/utils/Stream.js
```diff
@@ -0,0 +1,42 @@
...
+  'pH controller status': defaultIcon,
+  'controller status': defaultIcon,
+  'security flag': defaultIcon,
+  'motor speed': defaultIcon,
+  'valve position': defaultIcon,
+  'pump speed': defaultIcon,
+  'blood speed': defaultIcon,
+  'P parameter': defaultIcon,
+  'I parameter': defaultIcon,
+  'D parameter': defaultIcon,
```
DanielSeehausen: these will be replaced by specific images once we get them from someone in the office...I keep bothering them about this

src/utils/Stream.js
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export default class Stream {
+
+  static getDefaultImage() { return defaultIcon }
+
+  static getImage({ dataType }) {
```
DanielSeehausen: dataType is the new key that will be on upstreams (from firmware). The complete list of possible values is in the map above.

src/components/device/dashboard/MissingStreams.jsx
```diff
@@ -2,7 +2,7 @@ import React, { useState }from 'react'
 import { Segment, Table, Button, Image, Message, Icon } from 'semantic-ui-react'
```
DanielSeehausen: this file looks like a big change, but its mostly an indentation fix

src/components/device/dashboard/MissingStreams.jsx
```diff
@@ -24,57 +24,60 @@ const MissingStreams = ({ activeDevice, missingStreams, callUpdateDevice }) => {
...
+        <Icon name='warning' />
+        Some streams from your dashboard are not connected anymore:
+        <span
+          className='toggle-text'
+          onClick={() => toggleMissingStreamsView(!showMissingStreams)}
+        >
+          {toggleText}
+        </span>
+      </Message>
+      {showMissingStreams && (
```
DanielSeehausen: ternary to short circuit

src/components/device/dashboard/MissingStreams.jsx
```diff
@@ -24,57 +24,60 @@ const MissingStreams = ({ activeDevice, missingStreams, callUpdateDevice }) => {
...
-            ))}
-          </Table.Body>
-        </Table>
-      </Segment> : null}
+            </Table.Header>
+            <Table.Body>
+              {missingStreams.map(missingStream => (
+                <Table.Row key={missingStream.streamId}>
+                  <Table.Cell>
+                    <Image size='mini' src={Stream.getDefaultImage()} />
```
DanielSeehausen: here is the actual change!

src/components/device/dashboard/MissingStreams.jsx
```diff
@@ -24,57 +24,60 @@ const MissingStreams = ({ activeDevice, missingStreams, callUpdateDevice }) => {
...
+                </Table.HeaderCell>
               </Table.Row>
-            ))}
-          </Table.Body>
-        </Table>
-      </Segment> : null}
+            </Table.Header>
+            <Table.Body>
+              {missingStreams.map(missingStream => (
+                <Table.Row key={missingStream.streamId}>
```
DanielSeehausen: using streamId instead of array index

PR #582, opened 2020-03-10
Graphs configuration persistence

### LINKED ISSUE #447: GRAPHS 4 UIConfig: Data Viewer should persist and load data from backend
[This is blocking, and needs to be implemented first](https://github.com/ospin-web-dev/REACTor/issues/445)

As a user, whenever I:
- click and drag a graph to re-order it
OR
- fold/unfold a graph
I should update the backend UIConfig to represent the change. If there is outdated stream data in the UIConfig (e.g. streams that have been deleted), they should not be destroyed when new order/data is persisted, and they should maintain their relative positions. 

Additionally, when I load the Data Viewer page, it should load from this data. If there is data that is not present (e.g. deleted streams), they should be ignored and their positions maintained.

O-Susanoo:
Peek 2020-03-10 17-34

All the modifications made to graphs are now persisted , these configurations are user specific so that each user can order/fold the graphs the way that fit there best interests

\private
closes #447


src/components/processviewer/body/dataviewer/DataGraphs.jsx
```diff
@@ -35,22 +31,22 @@ class DataGraphs extends React.Component {
...
     dispatchSetDeviceUIConfig(userId, activeDevice.id, newConfig)
   }
 
   toggleStreamFolded = (groupName, streamId) => {
-    const { activeDevice, uIConfigs, dispatchSetDeviceUIConfig, userId } = this.props
-    const newConfig = UIConfig.toggleStreamFolded(uIConfigs, activeDevice.id, groupName, streamId)
+    const { activeDevice, dispatchSetDeviceUIConfig, userId } = this.props
+    const newConfig = UIConfig.toggleStreamFolded(activeDevice.uiconfig, groupName, streamId)
 
     dispatchSetDeviceUIConfig(userId, activeDevice.id, newConfig)
```
vdiezel: Didn't eslint complain about using destructuring? 
`const { id, uIConfig } = activeDevice`? or 
`const { activeDevice: { id, uIConfig }, ... } = this.props`

src/redux/thunks/thunks.js
```diff
@@ -452,13 +465,22 @@ export function callLoadDevice(loadedDevice, userId, deviceId) {
...
       await subscribeToDeviceTopics(device)
 
       const deviceUsers = await nexus.device.getUsers(device.id)
       dispatch(setActiveDevice({ device, deviceUsers }))
 
       const processesData = await nexus.device.getProcesses({ deviceId })
+
+      const deviceUpStreams = getDeviceUpStreams(device)
+      const processesUpStreams = getProcessesUniqueUpStreams(processesData)
+      const combinedUpStreams = [...deviceUpStreams, ...processesUpStreams]
```
vdiezel: I see a danger here when you parse all the processes data for new streams: 

Imagine we are going to load only the most recent 50 processes in future and the user will have a button to load the older ones. If one of the older processes contains a stream missing on the uIConfig, it will no be added to the UIconfig unless we apply the same logic again on fetching more processes. 

I'm thinking about what would be the best entry point to check for missmatches in the streams of a process, and to me this would be when you actually open a process (so maybe in the component `ProcessViewer` when it mounts?)

There I would check for missmatches and fix them.

src/utils/UIConfig/GraphsUIConfig.js
```diff
@@ -88,4 +41,23 @@ export default class GraphsUIConfig {
...
     return graphs
   }
+
+  static addNewStreamsToConfig(groups, streams) {
+    streams.forEach(({ groupName, id }) => {
+      const streamUI = { streamId: id, folded: false }
+      const groupIndex = groups.findIndex(group => group.name === groupName)
+      if (groupIndex >= 0) {
+        const streamIndex = groups[groupIndex].streams.findIndex(stream => stream.streamId === id)
+        if (streamIndex < 0) {
```
vdiezel: A `streamIndex === -1` would be more explicit

src/utils/UIConfig/UIConfig.js
```diff
@@ -2,94 +2,48 @@ import GraphsUIConfig from 'utils/UIConfig/GraphsUIConfig'
...
-    return uIConfigs.find(config => config.deviceId === deviceId)
-  }
-
-  static findOrCreateDeviceUIConfig(uIConfigs, deviceId, userId) {
-    return this.findDeviceUIConfig(uIConfigs, deviceId) || this.createDeviceUIConfig(deviceId, userId, [])
+  static checkForNewStreams(deviceUIConfig, streams) {
+    const streamsIds = streams.map(stream => stream.id)
+    const { graphs: { groups } } = deviceUIConfig
+    const backendStreamIds = groups.map(group => group.streams).flat().map(stream => stream.streamId)
+    return streamsIds.filter(streamId => !backendStreamIds.includes(streamId)).length !== 0
```
vdiezel: `return streamIds.some(streamId => !backendStreamIds.includes(streamId))`

src/components/processviewer/body/dataviewer/DataGraphs.jsx
O-Susanoo: Nope

src/redux/thunks/thunks.js
O-Susanoo: Thank you for pointing that one out  :+1:

src/utils/UIConfig/GraphsUIConfig.js
O-Susanoo: Niiiiiiiiiiiiiiiiiiiiiiiiiiiiice :100:

src/utils/UIConfig/UIConfig.js
O-Susanoo: image

PR #586, opened 2020-03-11
565 fix device widget cursor

### LINKED ISSUE #556: Differentiate Device Logs from Process Logs
When working with Jordi and Lina
It became evident that there was some confusion around logs and which is what. 

Users may benefit from some way to differentiate process logs from device logs. Consider here where they are both on screen:

image

The lowest effort solution would be to call either both of them `Device Logs` and `Process Logs` respectively. Since we have a rework coming to the process space in general, that might be best for now, as alternatives would likely involve reworking how processes are displayed. If you have a strong alternative solution in mind, :gopher: it

### LINKED ISSUE #565: Device Widget Cursor when over Device Name
When hovering over a device name, the cursor moves to a pointer, even though the behavior is the same as the pointer finger everywhere else on the cursor

### Deliverable
- [ ] When hovering over the device name, the cursor should be the same as the rest of the widget
- [ ] When editing the name of a device, the cursor should still act as a text box interaction cursor

vdiezel:
closes #565 
closes #556 

recommended to review by commit (first commit has the main change, rest is linting)


src/components/device/DeviceWidget.jsx
```diff
@@ -156,7 +156,15 @@ const renderRunning = (
...
+      className='meta'
+      style={{
+        justifyContent: 'center',
+        alignItems: 'center',
+        height: '57px',
+        display: 'flex',
+      }}
+    >
       <span style={{ fontSize: '15pt' }}>
         {connectionState === 'online' ? 'ready for process' : 'ready to connect'}
```
ConorTighe1995: Is there an interface or Object.freeze for the connection states in the REACTor codebase?

vdiezel: no, the connectionState is derived from the device heartbeat - it an empty message publishes every 2 second. As long as we receive it, the connection state is `online`. When the heartbeat goes missing for at least 3 seonds, we consider the device to be `offline`.

PR #594, opened 2020-03-18
592 fix loading spinners in graphs

### LINKED ISSUE #302: Time frame button 'Live' on the graph header bar doesn't disappear automatically when a process is finished
bug_report_1

The Live button should dissappear once a process goes into the `finished` state, but it only dissappears when you click on another button within the graph header (image shows a `finished` process)

### LINKED ISSUE #592: When data loading request fails the loading indcators don't disappear
Loading heartbeat data failed (investigation running), but the loading indicators didn't disappear.
The should disappear and maybe be replaced by an error display.

Bildschirmfoto von 2020-03-18 08-06-55

vdiezel:
first commit closes #592 
 gif shows the real use case: the heartbeat upstream for the pressure throws an error because it seems the chunks are getting too big too quickly due to the huge amount of data. In the deployed versions this graph would be stuck endlessly in the loading indicator view
typo (ould => could ) is fixed in the last commit
error_on_load

second commit closes #302 (done here because I had to touch the same code)
Older bug which is caused by our manually set condition when a Stream compoennt is updated. Had to pull process state out of the activeProcess beforehand and pass it down to make the component update when the process goes from running to finished.  The green 'Live' button now disappears as intended.
live_button_disappears


src/components/processviewer/body/dataviewer/Group.jsx
```diff
@@ -32,15 +32,22 @@ const Group = ({
                     const { streamId } = streamUI
                     const stream = Process.getProcessStreamById(activeProcess, streamId)
                     if (!stream) return null
```
ConorTighe1995: Is this null value just for rendering a component? Normally I try avoid returning null where I can:

https://stackoverflow.com/a/1274865

vdiezel: react itself proposes rendering nothing using null: 
https://reactjs.org/docs/conditional-rendering.html

PR #595, opened 2020-03-19
348 display target lines

vdiezel:
- adds support for horizontal lines in the graphs (now used for target lines, can also be used for event trigger limits!)
- implements a checkbox (only when visible when the upstream does have a target associated) toggle the display of it (should prolly go into the uiConfig, too, at a certain point).

This is backwards compatible - can be merged and needs only tiny fixes when the firmware catches up (marked with a `TODO TARGET_LINES`

show_target_lines


DanielSeehausen:
you might already have it planned, but please make it a changelog!

vdiezel:
> Looks great!
> 
> Should we consider having a target line on by default?
> I imagine its generally always wanted, and we could make it less visually powerful with the opacity key that it takes, and/or by making it dashed or something, so that if it is on by default its easy to visually ignore.

We should simply make it part of the UIconfig and let the user decide.

PR #598, opened 2020-03-20
Dashboard Widget Icon Facelift!

### LINKED ISSUE #596: add image assets for dashboard widgets
this follows up on #520, which has already been merged, by adding the image assets themselves!

self pointed because it is really just adding the images in. All of the work was already done.

DanielSeehausen:
A _lot_ of our users reach out to us and say, "Hey OSPIN, I love the Device Dashboard, and I spend more time watching it than Netflix. Is there anything you can do to make it an even more visually descriptive experience?"

Of course!

Device Dashboards will now have dynamic images based on the type of stream they represent:

dashboard

Keep the requests coming in!

\private

closes #596 

related to #520 
---
The image above shows the current mapping. Any not shown are likely on the default image of a graph. Firmware is not currently sending all of the stream types listed above.

The stream names have been replaced with the **DataType Names**, which is the value we receive from the device/firmware.

Please take a look at the icons and determine whether they make sense or not with the given datatype names. Some of them aren't great, as I worked with the lemons that Yahweh gave us this day.
---
Can be removed commit wise!

If you do, you will notice there is a commit for test data that is then reverted. The reason I did this is so I always have a point in history I can go back to (or put back on head) where I can play with mocked test data/take a new image. It would have been more clever if I kept it always as the top commit.


src/utils/StreamIcons.js
```diff
@@ -0,0 +1,58 @@
...
+  pwm: electrophysiology,
+  analog: hand,
+  duration: analogueClockPartialFill,
+  O2: cO2o2GasExchange,
+  direction: systemWithDirection,
+  switch: hand,
+  bitset: defaultIcon,
+  min: boundGraph,
+  max: boundGraph,
+  'pH controller status': pHpO2WithSensor,
```
ConorTighe1995: How come these become strings/have an apostrophe around them?

DanielSeehausen:
> Nice, why is the switch Icon a hand tho?
> 
> https://tenor.com/view/strong-hand-take-my-hand-gif-12558750

as I understand it from firmware - the switch is actually a physical switch on the device that users manipulate. We don't have a great asset for that, so I chose hand! If anyone has a better choice let me know

vdiezel:
Looks really good. Lets take ninas offer to create icons for us that are not 100% on :dart:

DanielSeehausen:
> Looks really good. Lets take ninas offer to create icons for us that are currently not 100% on 🎯

agreed!
Ill wait to see how long that will take before merging. If it is more than a couple of days ill merge this and edit later

src/utils/StreamIcons.js
DanielSeehausen: spaces are not allowed in keys :(
unfortunately we don't control these key values - this is sent to us directly by firmware.

ConorTighe1995: Ah I see! No worries that explains it.

PR #600, opened 2020-03-20
517 toggleable streams 

### LINKED ISSUE #517: Change rocker RPM variation downstream view control type to toggle-able
Firmware will communicate an `expectedDataType` with every downstream payload, with an option being 'boolean'. All others should default to what they are now ('number')

vdiezel:
closes #517 
review commit wise.

For live preview, pull that branch and checkout Celltainer.

Can't be merged without firmware coordination.


src/components/processviewer/body/builder/BooleanStream.jsx
```diff
@@ -0,0 +1,70 @@
...
+              style={{ marginTop: '8px' }}
+            />
+          </Form.Field>
+        </Form.Group>
+      </div>
+    )
+  }
+}
+
+export default connect(mapStateToProps, mapDispatchToProps)(BooleanStream)
```
vdiezel: Bildschirmfoto von 2020-03-20 08-19-46

src/components/processviewer/body/processconfiguration/ProcessConfiguration.jsx
```diff
@@ -42,24 +83,10 @@ const ProcessConfiguration = props => {
...
-              )
-            })}
-          </Table.Body>
-        </Table>
+          .filter(stream => !stream.controller || (stream.controller && showControllerParameters))
+          .map(stream => renderStreamByExpectedValueType(stream, activeDevice, phases))}
+      </Table.Body>
+    </Table>
   )
 }
```
vdiezel: Bildschirmfoto von 2020-03-20 08-19-39

src/components/device/setup/ConnectedStreamsTable.jsx
```diff
@@ -62,16 +72,16 @@ const CalibrationTable = props => {
...
               <Table.Cell>
-                {stream.min}
+                {stream.min !== undefined ? stream.min : '-'}
               </Table.Cell>
               <Table.Cell>
-                {stream.max}
+                {stream.max !== undefined ? stream.max : '-'}
               </Table.Cell>
               <Table.Cell>
                 {stream.controller ? `${stream.controller.name} (${stream.controller.type})` : '-'}
```
vdiezel: Bildschirmfoto von 2020-03-20 08-31-11

src/components/device/modals/SetStreamDefaultModal.jsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react'
```
vdiezel: Changed the looks slighly and behaviour. For boolean streams and doesn't really make sense to have a `Submit` button, so I don't isplay it. See a comparison of `float` (our regular)  and `boolean` streams.

Bildschirmfoto von 2020-03-20 09-50-51
Bildschirmfoto von 2020-03-20 09-50-55

src/components/device/setup/StreamDefaultsTable.jsx
```diff
@@ -17,6 +17,7 @@ const mapDispatchToProps = dispatch => ({
   callDeleteStreamDefault: (streamId, deviceId) => dispatch(callDeleteStreamDefault(streamId, deviceId))
```
vdiezel: Bildschirmfoto von 2020-03-20 11-29-28

src/utils/validation.js
```diff
@@ -113,6 +113,8 @@ export function validateStreamTargets(process, device) {
```
vdiezel: validation file is sadder than my love life. I created a ticket to refactor the while deal. For now, because the toggle component doesn't really allow entering something unexpected we don't have to validate anything.

src/components/device/modals/SetStreamDefaultModal.jsx
ConorTighe1995: Where does the current default value go? might be nice to display it across from the toggle in text to take up the white space

src/utils/validation.js
ConorTighe1995: We need a ticket for you to set up tinder asap

src/components/device/modals/SetStreamDefaultModal.jsx
```diff
@@ -31,6 +31,24 @@ class SetStreamDefaultModal extends React.Component {
     isSuccess: false,
   })
 
+  handleAPIResponse = res => {
```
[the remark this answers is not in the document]
ConorTighe1995: Is handleAPIResponse an interceptor of some sort for many/all API calls? just wondering because the name is very generic so its hard to tell.

src/components/device/setup/ConnectedStreamsTable.jsx
ConorTighe1995: nice

src/components/processviewer/body/builder/BooleanStream.jsx
```diff
@@ -0,0 +1,70 @@
...
+  handleSubmit = async () => {
+    const {
+      activeProcess,
+      dispatchCallUpdateProcess,
+      phaseId,
+      stream,
+      dispatchSetStreamTarget,
+    } = this.props
+
+    const newTarget = !getStreamTarget(activeProcess, phaseId, stream.id)
```
ConorTighe1995: `!getStreamTarget(activeProcess, phaseId, stream.id)`

Is this meant to be converted to a Boolean? is new target meant to be true or false? just wondering as I would assume new would indicate a new object, as opposed to something like isAvailable or isReady

src/components/device/modals/SetStreamDefaultModal.jsx
vdiezel: Because for the toggle-able streams we don't have input fields and a submit button, the toggle always shows you already the current default - there is no need for intermediate storage.

src/components/processviewer/body/builder/BooleanStream.jsx
vdiezel: it actually inverts the current target value (which are booleans for BooleanStreams).
You are right in a sense that it `getStreamTarget` for boolean streams is a bit missleading, something like `getStreamState` would make more sense but it would be a 100% duplicate of `getStreamTarget`.

src/components/device/modals/SetStreamDefaultModal.jsx
ConorTighe1995: Im not sure how they would use this to be fair so maybe Im interpreting it wrong, is Rocker RPM a custom stream or would they know what the rpm value is without needing to see it displayed? I was talking about having it in styled text so they know the default RPM value associated with the stream not an input element, as I wouldn't know what RPM value is associated with Rocker RPM, its toggled in the picture but no values to say what will be applied as the default, but Im no biologist so maybe that would be obvious to them and unneeded.

vdiezel: the naming is confusing because it was a quick spoof, something better would be: `Activate Rocker`. So it expects only `true` or `false` as a value

ConorTighe1995: Ohh ok I get what you mean, thanks for the explanation!

PR #607, opened 2020-03-20
fixes manual zoom

### LINKED ISSUE #602: In the Process Table view, the select all button should only select the visible processes!
As a user I expect that when I filter processes by a certain property (e.g. because I want to delete those) and then click 'select all' only those processes are selecteed which are actually shown. But 
it currently actually selects all (also the hidden ones) - which might be very unpleasant for a user they delete all processes this way by accident

Bildschirmfoto von 2020-03-20 15-29-25

### LINKED ISSUE #606: Manual zoom input fields are not working, making the feature useless
Input fields are not responsive: values should be changeable and they should set the viewport on the y-axis accordingly. 
Bildschirmfoto von 2020-03-20 15-51-00

vdiezel:
- Graph Control
  - manual' zoom has been renamed to 'fix y-axis' ...
  -  and it works again!
- Process Table View
  - the 'Select All' button now ignores processes that were filtered using the filter options

\private
First commit:
closes #606 
renamed manual zoom to 'fix y-axis' because that is what it is.

seconds commit:
closes #602 (squeezed in here to prevent so many deployment for tiny changes)


src/components/processviewer/body/dataviewer/ConfigForm.jsx
```diff
@@ -20,7 +20,7 @@ const getStreamUnitOptions = stream => getUnitOptions(stream.unit).map(unit => (
 
 const zoomOptions = [
   { value: 'auto', text: 'auto' },
-  { value: 'manual', text: 'manual' },
+  { value: 'fix-y-axis', text: 'fix Y-axis' },
```
DanielSeehausen: any chance of a:
fix -> fixed

to avoid any ambiguity

DanielSeehausen: i updated

PR #609, opened 2020-03-23
604 add more querying options to the process view

### LINKED ISSUE #604: Add the owner of the process to the Process Table View
Add a column here that displays the process owner (if it can be found on the device). 
Also add the property to the filter dropdown to allow to query processes by their owner (top right corner)!

Bildschirmfoto von 2020-03-20 15-40-36

### LINKED ISSUE #605: Allow to query processes by their createdAt timestamp
This can be either done easily, just query by the Date string, or more elaborate with the Calendar Components as used in the Logs view. 
Bildschirmfoto von 2020-03-20 15-40-36

vdiezel:
closes #604 
closes #605 

- adds owner name to the table and allows querying by owner name
- adds a  toggle-able date query menu in the process view (could'nt use the Popup component here because the DatePicker we have looses its display state when the popup disappears, so I used a workaround, not as beautiful as a nice popup, but the toggled menu resembles what we have in the LogView
- prevents entering a start date above the end date when end date has been chosen first in all DatePickers

datequerymenu

Can be reviewed commitwise (I'll omit this message in the future in the hope that it will always be the case)


src/components/device/processes/Processes.jsx
```diff
@@ -64,8 +68,32 @@ class Processes extends React.Component {
...
+    }
+
+    const { deviceUsers } = this.props
+    const re = new RegExp(queryString, 'i')
+
+    const userIds = deviceUsers
+      .filter(user => re.test(user.userName))
+      .map(user => user.id)
 
+    console.log(userIds)
```
vdiezel: will be gone in last commit.

src/components/device/processes/DateQueryMenu.jsx
```diff
@@ -0,0 +1,60 @@
...
+  { text: 'created', value: 'createdAt' },
+  { text: 'started', value: 'startedAt' },
+  { text: 'finished', value: 'finishedAt' },
+])
+
+const getDisplayedStyle = showDateQueryMenu => (
+  showDateQueryMenu
+    ? { marginBottom: '16px' }
+    : { display: 'none' }
+)
```
vdiezel: setting styles is a workaround for preventing the DatePickers losing their state. They would do that if I would simply not render the component by returning null.

src/components/device/processes/Processes.jsx
```diff
@@ -134,13 +132,13 @@ class Processes extends React.Component {
...
     if (startDate === null) return true
-    return process[timeQueryProperty] > startDate.getTime()
+    return process[timeQueryProperty] && process[timeQueryProperty] > startDate.getTime()
   }
 
   filterByEndTime = process => {
     const { endDate, timeQueryProperty } = this.state
     if (endDate === null) return true
-    return process[timeQueryProperty] <= endDate.getTime()
+    return process[timeQueryProperty] && process[timeQueryProperty] <= endDate.getTime()
```
vdiezel: without this additional condition, `null <= endDate.getTime()` always returns true

src/components/device/processes/ProcessRow.jsx
```diff
@@ -51,37 +82,21 @@ const ProcessRow = props => {
...
+      <Table.Cell>{getDisplayableProcessOwner()}</Table.Cell>
       <Table.Cell>{Time.getTrimmedDateString(createdAt)}</Table.Cell>
       <Table.Cell>{finishedString}</Table.Cell>
       <Table.Cell>{duration}</Table.Cell>
       <Table.Cell collapsing>
         <Checkbox
-          checked={props.checked}
-          onChange={() => props.selectHandler()}
+          checked={checked}
+          onChange={() => selectHandler()}
```
O-Susanoo: No need for arrow function

src/components/device/processes/Processes.jsx
```diff
@@ -44,19 +54,99 @@ class Processes extends React.Component {
...
+
+  filterByAppliedQuery = process => {
+    const { filteredProcessIds } = this.state
+    return !filteredProcessIds.includes(process.id)
+  }
+
+  filterByStartTime = process => {
+    const { startDate, timeQueryProperty } = this.state
+    if (startDate === null) return true
+    return process[timeQueryProperty] && process[timeQueryProperty] > startDate.getTime()
```
O-Susanoo: ternary operator can be handy

vdiezel: `process[timeQueryProperty] ? process[timeQueryProperty] > startDate.getTime() : false` was too verbose for my test

PR #610, opened 2020-03-23
Updates reactor to no longer reference 'owners' access group

DanielSeehausen:
related to ospin-web-dev/hambda#200

updates reactor to:

- remove all references to the 'owners' group
- delete the 'owners' group in the object mapper

The deletion of the 'owners' group will be removed after the data migration is finished!


src/components/device/manageaccess/User.jsx
```diff
@@ -30,10 +30,10 @@ const User = ({
 
   const { userName, email, id } = user
 
-  const groupOptions = activeDevice.access.map((group, i) => ({
+  const groupOptions = activeDevice.access.map(group => ({
     value: group.name,
-    text: Authorizer.convertOwnerGroupName(group.name),
-    key: i,
+    text: group.name,
+    key: group.name,
```
DanielSeehausen: there are a few places where I updated this as well because I was in the block 🇸🇱

src/components/device/modals/AccessProcessModal/AccessProcessModal.jsx
```diff
@@ -183,7 +183,7 @@ class AccessProcessModal extends React.Component {
 
     return access.map(group => ({
       value: group.name,
-      text: Authorizer.convertOwnerGroupName(group.name),
```
DanielSeehausen: this method was super helpful to have when doing this PR

src/components/device/modals/TransferOwnershipModal.jsx
```diff
@@ -76,9 +76,10 @@ const TransferOwnershipModal = ({
             header='Are you sure?'
             content={
               <p>
-                Transfering the ownership will remove your additional privileges within the
-                admin group and grant them to the new device owner. The new device owner
-                will be able to control any kind of access to the device, including yours!
+                Transfering the ownership will remove your ownership
+                privileges and grant them to the new device owner. The
+                new device owner will be able to control all user
+                access to the device, including yours!
```
DanielSeehausen: are these line breaks maintained?

src/redux/helper/objectMapper.js
```diff
@@ -20,6 +20,13 @@ export const createUserLogObject = userLogData => ({
   timestamp: userLogData.createdAt,
 })
 
+
+function removeAccessOwnersGroup(access) {
+  const ownersIndex = access.findIndex(group => group.name === 'owners')
+  access.splice(ownersIndex, 1)
+  return access
+}
+
```
DanielSeehausen: this, and the references to it below, will be removed once the data migration is complete

src/utils/Authorizer.js
```diff
@@ -21,12 +21,6 @@ class Authorizer {
 
   static isOwnerOf = (resource, userId) => resource.ownerId === userId
 
-  // legacy function in preparation of renaming the owners group into admin
-
-  static convertOwnerGroupName = groupName => (
-    groupName === 'owners' ? 'admin' : groupName
-  )
-
```
DanielSeehausen: 🔫

src/redux/helper/objectMapper.js
vdiezel: `return access.filter(group => group.name === 'owners')` :japanese_ogre:

src/utils/Authorizer.js
vdiezel: image

PR #626, opened 2020-03-29
623 phase choice after adhoc revert

### LINKED ISSUE #617: Small visual fixes on the Process builder
When the process tree has a bigger height than the builder form then the plus button at the bottom has no padding to the browser edge

Bildschirmfoto von 2020-03-25 10-04-28

The input fields should not be 'disabled' but either be 'read only' or completely replaced by a normal text display!

Bildschirmfoto von 2020-03-25 10-04-37

### LINKED ISSUE #623: Issues with ad hoc changes: Undoing an added phase
Reproduce: 

1. Start a process with any number of phases
2. Add a phase -> the changes table appears with your new phase listed
3. When you undo the changes now and you had the new phase selected, the process builder will appear empty, because the selected phase now doesn't exist anymore

-> solution: When a phase is selected that doesn't exist, it should select the currently running one for a running process, or the last one in a process which is executable

### LINKED ISSUE #625: Ad hoc Issue: Weird state when doing having adhoc changes staged and doing a non adhoc change to the process
Caught by our best man Adel:

We have a weird state of the process display when doing adhoc changes: 

1. Start a process
2. Do some adhoc changes to the process, but dont apply them.
3. Change the phase name -> adhoc changes will be applied to your process ONLY, but NOT updated on the device!!

The cause is this: Renaming a phase is not considered an adhoc change, so you don't need to confirm it. So the process is updated with the new phase name, but also containing adhoc changes hidden in it.

This one will be probably tricky to solve. The easy, last resort solution would be to forbid non adhoc changes (currently only eventTrigger and phase name changes) when they have adhoc changes staged.

vdiezel:
This PR contains a series of small bug fixes and visual changes: 

1. Commit: closes #623 
2. Commit: bug fix I discovered on the fly re: Pagination of processes in the process table view which was probably introduced by me recently 
3.  Commit: closes #625 - this is a bigger issue we have to address with a bigger refactoring (we need to write a merging algorithm that merges changes that are adhoc changes with those which aren't. For now, I simply forbid making non adhoc changes (the only one so far is phase renaming) when adhoc changes are staged. 
4. and 5. Commit closes #617: Minor visual changes in the process builder


src/components/processviewer/body/builder/ChangesFooter.jsx
```diff
@@ -43,6 +53,11 @@ const ChangesFooter = ({
   const revertChanges = () => {
     const revertedDescription = PDVC.revertToSnapshot(activeProcess, processDescriptionSnapshots)
 
+    if (!Process.hasPhaseWithId(revertedDescription, displayedProcessPhase)) {
+      const fallbackPhaseId = Process.getRunningPhase(activeProcess).phaseId.toString()
+      dispatchSetDisplayedProcessPhase(fallbackPhaseId)
+    }
+
```
vdiezel: the important change of this commit

src/components/processviewer/body/builder/PhaseTabContent.jsx
```diff
@@ -48,7 +54,15 @@ class PhaseTabContent extends React.Component {
...
   submitNewPhaseName = async newPhaseName => {
-    const { activeProcess, phaseId, dispatchRenamePhase } = this.props
+    const { activeProcess, phaseId, dispatchRenamePhase, processDescriptionSnapshots } = this.props
+
+    const diffs = PDVC.getDiff(activeProcess, processDescriptionSnapshots)
+
+    if (diffs.length) {
+      FlashMessenger.warning('Please revert or apply your process changes before renaming a phase')
+      return
+    }
```
vdiezel: blocks renaming a phase when adhoc changes are staged.

src/components/processviewer/phasesdisplaybar/PhasesNavbarWrapper.jsx
```diff
@@ -125,6 +125,7 @@ class PhasesNavbarWrapper extends React.Component {
             color='blue'
             className='centered-h'
             disabled={activeProcess.state === 'finished'}
+            style={{ marginBottom: '56px' }}
```
vdiezel: Old 

Bildschirmfoto von 2020-03-25 10-04-28

New

Bildschirmfoto von 2020-03-29 22-50-17

src/components/processviewer/body/builder/Stream.jsx
```diff
@@ -182,14 +182,13 @@ class Stream extends React.Component {
...
                 onFocus={this.handleFocus.bind(this)}
-                disabled={disabledInput}
+                readOnly={disabledInput}
               />
               <Select
                 compact
                 options={unitOptions}
                 value={selectedUnit}
                 onChange={(e, data) => this.changeDisplayedUnit(e, data)}
-                disabled={disabledInput}
```
vdiezel: allows now to change the unit on inputs for phases that were finished.

src/components/device/processes/Processes.jsx
```diff
@@ -46,7 +46,7 @@ class Processes extends React.Component {
     this.setState({ activePage: data.activePage })
   )
 
-  updateResultsPerPage = resultsPerPage => {
+  updateResultsPerPage = (_, { value: resultsPerPage }) => {
```
ConorTighe1995: What is the _ for in the params?

vdiezel: event - I dont need it

src/components/processviewer/phasesdisplaybar/PhasesNavbarWrapper.jsx
O-Susanoo: I like the before after image cosmetic surgery stuff

ConorTighe1995: if these keep being appending will the phase boxes just keep extending the height? could be worth creating a scroll-able container in the future to store these.

src/utils/process.js
```diff
@@ -636,7 +636,7 @@ export class Process {
   static isValidPhaseProgression = (process, nextPhase) => {
 
     const existingPhaseInProcess = getPhaseById(process, nextPhase.phaseId)
-    const runningPhaseId = getRunningPhase(process).phaseId
+    const runningPhaseId = this.getRunningPhase(process).phaseId
```
DanielSeehausen, line 639: :laser:

src/components/processviewer/body/builder/ChangesFooter.jsx
DanielSeehausen: not a request to change, just minor note - 

`displayedProcessPhase` seems to be an id. Since phase (as a data object) is a concept we do have in the code base, ID should be appended on variables like this.

src/utils/process.js
```diff
@@ -769,4 +769,15 @@ export class Process {
...
     return false
   }
+
+  static hasPhaseWithId(description, phaseId) {
+    return phaseId in description
+  }
+
+  static getRunningPhase(process) {
+    if (process.state === PROCESS_STATES.finished) {
+      return { phaseId: Infinity }
```
DanielSeehausen, line 779: What is the intent of using `Infinity` here?

src/components/device/processes/Processes.jsx
DanielSeehausen: @ConorTighe1995 something often seen in JS is an underscore which, (if you haven't already required _ in global scope to mean the underscore library), is used to throw away an unneeded arg in a series of ordered args.

Taking `_` in the place of `e` says to the reader "I'm not used". If `e` was to be left, the reader would assume `e` is used, only to have their heart broken when reading the contents of the method!

ConorTighe1995: Cool! good to know! Ive seen underscore before and Ive seen this used in libraries under the hood but assumed those libraries were using underscore too! this is good info because I actually didn't know it was a practice for args.

src/components/processviewer/body/builder/PhaseTabContent.jsx
DanielSeehausen: <3 the use of FlashMessenger
High value APIs mang
brought to you by @vdiezel

src/components/processviewer/phasesdisplaybar/PhasesNavbarWrapper.jsx
DanielSeehausen: @ConorTighe1995 that's a good idea. Unless anyone else objects, [would you be interested in making a write up for the backlog?](github.com/ospin-web-dev/REACTor/issues/new)

src/components/device/processes/Processes.jsx
vdiezel: thanks to daniel for actually explaining the cryptic stuff I write often!

src/utils/process.js
vdiezel: I think it used to have a function which is not used anymore for the sake of it. I'll investigate this.

vdiezel: I think this is actually a bug from a refactoring, the `startTime` should be `Infinitity`. Somehow it never caused any trouble I assume.

DanielSeehausen: 👍

src/utils/process.js
```diff
@@ -770,14 +770,18 @@ export class Process {
...
     return phaseId in description
   }
 
   static getRunningPhase(process) {
     if (process.state === PROCESS_STATES.finished) {
-      return { phaseId: Infinity }
+      return this.noPhaseObject()
     }
-    return process.progression[process.progression.length - 1] || { phaseId: -1, startTime: -1 }
+    return process.progression[process.progression.length - 1] || this.noPhaseObject()
```
DanielSeehausen: 👍

src/components/processviewer/phasesdisplaybar/PhasesNavbarWrapper.jsx
ConorTighe1995: Yea I can do that!

PR #630, opened 2020-03-31
Fix UIConfig duplicate streamId bug + Tests

### LINKED ISSUE #628: UIConfig should check if stream belong to correct group of the process
When a stream exists on 2 groups in the ui config but only lives on one of theme on the process we display the stream only on the group it lives on in the process

O-Susanoo:
closes #628 

ironMan

I see that REACTor is all grown up now i wonder how many software years i'have been gone


src/utils/UIConfig/UIConfig.js
```diff
@@ -48,4 +49,17 @@ export default class UIConfig {
     GraphsUIConfig.moveStream(graphs, groupName, sourceIndex, destinationIndex)
     return config
   }
+
+  static isGroupEmptyOnActiveProcess(group, activeProcess) {
+    const emptyGroup = group.streams
+      .map(stream => Process.getProcessStreamById(activeProcess, stream.streamId))
+      .every(stream => stream === undefined || stream.groupName !== group.name)
+    return emptyGroup
```
vdiezel: your could immediately return here ` return group.streams.map....` :grimacing:

O-Susanoo: copy paste is a disease that destroy lifes each year

O-Susanoo: i will fix it

test/components/processviewer/body/dataviewer/DataGraphs.test.jsx
```diff
@@ -340,4 +340,181 @@ describe('DataGraphs.jsx', () => {
...
+        groupStreams.forEach(streamName => {
+          const { getByText: getStreamByName } = within(group)
+          getStreamByName(streamName)
+        })
+      })
+
+      //check if stream in old group is not visible
+      const heatingGroup = getByText(mockedDevice.uiconfig.graphs.groups[0].name).parentNode.parentNode
+      const { queryByText: getStreamByName } = within(heatingGroup)
+      expect(getStreamByName(streamsWithSameGroup[0].name)).toBeNull()
```
vdiezel: :ok_hand:

src/utils/UIConfig/UIConfig.js
vdiezel: you don't have, too - just trying to make you aware of it in the future

PR #638, opened 2020-04-02
Update issue templates

DanielSeehausen:
adds bug and feature issue templates for us 

the external bug report issue template receives minor edits


.github/ISSUE_TEMPLATE/feature.md
```diff
@@ -0,0 +1,34 @@
+---
+name: Feature
+about: describe a feature to be implemented
+title: ''
+labels: ''
+assignees: ''
```
ConorTighe1995, line 6: Should assignee's not be assigned when pulling the bug ticket form the backloag.

DanielSeehausen: you are right that assignees should only be assigned when the ticket is pulled from the backlog

this is metadata for the issue template used by github

if you provide users in this field, they whenever this template is used to create an issue, that user gets automatically assigned to it. As it is now, no one will be assigned, and issues created with this template go straight to the backlog

.github/ISSUE_TEMPLATE/ospin-browser-bug-report.md
```diff
@@ -18,7 +18,7 @@ Steps to reproduce the behavior:
 2. Click on '....'
 3. Scroll down to '....'
 4. See error
-_(providing device and process IDs is very helpful when applicable! You can find them in the url)_
+_(providing device and process IDs that this can be reproduced on is very helpful when applicable! You can find them in the url)_
```
vdiezel, line 21: Instead of asking for the deviceId and the processId we also simply could ask for the url. Makes it less complicated for them

DanielSeehausen: That sounds good - ill update

PR #642, opened 2020-04-02
639 visual indication for downstream bounds

### LINKED ISSUE #639: Add visual indication of downstream input field bounds
### Feature
 - when a user inputs an out of bounds value in a downstream input field, it goes red
 - a user must hover over the downstream title to get a tooltip of the bounds

### Proposed Solution
**as a user on the builder view...**
 - when I am actively editing a streams input field...
   - [ ] I always see the tooltip popup for the field in the same place it is now on the stream title
 - when I am hovered over a stream title, but not editing the input field...
   - [ ] I continue to see the min/max (this is implemented now)

### Images/Links

Current Behavior:
image

Tooltip:
image

DanielSeehausen:
closes #639 

- Downstreams will continue to display their bounds on hover
- Downstream input fields will now always display their bounds when they are active (when a user has clicked on them, or is provided values)

Apr-02-2020 13-25-07

(the red is more pronounced than in the gif, and it hasn't been changed from how it is currently behaving in prod)


src/components/processviewer/body/builder/Stream.jsx
```diff
@@ -162,14 +163,17 @@ class Stream extends React.Component {
     return (
       <div>
         <Popup
-          trigger={
+          trigger={(
             <span
               style={{ fontWeight: 'bold' }}
+              onMouseEnter={() => this.setState({ nameHovered: true })}
```
ConorTighe1995, line 169: I would have made a onNameHover function here:

`  const onToggleNameHover = () => {
      this.setState({ nameHovered: !nameHovered })
   }
`

then you could do this:

'onMouseEnter={this.onToggleNameHover}' and 'onMouseLeave={this.onToggleNameHover}'

vdiezel: yep! When you want to reference a variable in the previous state, REACT recommends to do this like this
`this.setState(prevState => ({ nameHovered: !prevState.nameHovered }))` 
you could also desctruct it from prevState immediately

DanielSeehausen: I considered the same re: the toggle-able and I believe it would certainly be cleaner code wise @ConorTighe1995 @vdiezel 

The reason I did enter/leave is because I have experience slightly buggy behavior in browsers with not always registering a mouse enter/leave event. With a toggle, that can then invert, and start doing the opposite behavior. 

Perhaps things have gotten better over time with handling that though, and its more fullproof now!

PR #647, opened 2020-04-02
Add processBuilder config support to UIConfig

### LINKED ISSUE #640: Add processBuilder support to UIConfig
Update REACTor UIConfig lib to detect and extract new downstreamGroups , the code should also be able to create processBuilder on UIConfig if it's not there 

- [x] rename all uiconfig on device to be uIConfig 

- [x] move methods that belong to graphs into graph module from UIConfig

- [x] Make code act on graphs instead of UIConfig

- [x] ADD Dashboard module to UIConfig

- [x] Make code act on Dashboard 

- [x] ADD ProcessBuilder module to UIConfig

O-Susanoo:
closes #640


src/utils/UIConfig/GraphsUIConfig.js
```diff
@@ -60,4 +60,16 @@ export default class GraphsUIConfig {
...
+    streamsGroups.forEach(groupName => {
+      const group = groups.find(grp => grp.name === groupName)
+      if (!group) {
+        groups.push({
+          name: groupName,
+          folded: false,
+        })
+      }
+    })
+  }
```
DanielSeehausen, line 74: I believe this belongs in a different module under `UIConfig` for `ProcessBuilderUIConfig` or something (Graphs is for the Graphs :p)

Alternatively, change the file to be generic and named generically, so its methods can be shared across both Graphs and ProcessBuilder config stuff

src/utils/UIConfig/UIConfig.js
```diff
@@ -10,11 +10,26 @@ export default class UIConfig {
     return streamsIds.some(streamId => !backendStreamIds.includes(streamId))
   }
 
+  static checkForNewGroups(deviceUIConfig, processDownStreams) {
```
DanielSeehausen, line 13: in considering the return value, this might be more accurately named `getNewGroups`

src/utils/UIConfig/UIConfig.js
```diff
@@ -10,11 +10,26 @@ export default class UIConfig {
...
     GraphsUIConfig.addNewStreamsToConfig(deviceUIConfig.graphs.groups, streams)
     return deviceUIConfig
   }
 
+  static addNewDownStreamGroupsToConfig(deviceUIConfig, processDownStreams) {
+    const streamGroups = getGroups(processDownStreams).map(group => group.name)
+    if (!deviceUIConfig.processBuilder) deviceUIConfig.processBuilder = { groups: [] }
+    GraphsUIConfig.addNewDownStreamGroupsToConfig(deviceUIConfig.processBuilder.groups, streamGroups)
+    return deviceUIConfig
+  }
```
DanielSeehausen, line 31: while I know it is split across the code base still, I believe we are preferencing downstream as a single word, so `processDownstreams`

src/utils/UIConfig/GraphsUIConfig.js
O-Susanoo: totally agree but i think i will go with your first suggestion ProcessBuilderUIConfig it make much more sense to me , the reason why it was here is beacuse the first draft i had processBuilder was just a member of graph so i didn't do the changes after

src/utils/UIConfig/UIConfig.js
O-Susanoo: :+1:

O-Susanoo:
Thanks for the review i will try to think about rearrangement then we can talk about it

test/utils/UIConfig/UIConfig.test.js
```diff
@@ -288,4 +288,145 @@ describe('UIConfig', () => {
...
+            {
+              name: 'pump',
+              folded: false,
+            },
+          ],
+        },
+      }
+
+      const result = UIConfig.addNewDownStreamGroupsToConfig(config, streams)
+      expect(result).toEqual(expect.objectContaining(expected))
```
ConorTighe1995, line 429: Could save duplication here by just doing `expect(result).toEqual(expect.objectContaining(config))`

src/redux/helper/objectMapper.js
```diff
@@ -62,7 +62,7 @@ export const createDeviceObject = deviceData => ({
   streamNameAliases: deviceData.streamNameAliases,
   subscriptions: {},
   readyForCommand: true,
-  uiconfig: deviceData.uiconfig,
+  uIConfig: deviceData.uIConfig || deviceData.uiconfig,
```
O-Susanoo, line 65: this is a temporary fix once backend code is up it will be like this
```
uIConfig: deviceData.uIConfig || deviceData.uiconfig
```

DanielSeehausen: nice!

src/utils/UIConfig/GraphsUIConfig.js
```diff
@@ -1,97 +1,41 @@
...
-    targetGroup.folded = !targetGroup.folded
-
+    graphs.groups = super.toggleGroupFolded(graphs.groups, groupName)
     return graphs
   }
 
   static toggleGroupStreamFolded(graphs, groupName, streamId) {
-    const { streams } = this.findGroupByName(graphs.groups, groupName)
-    StreamsUIConfig.toggleFolded(streams, streamId)
+    const groups = this.findGroupByName(graphs.groups, groupName)
```
DanielSeehausen: it looks like `super` is being used generally everywhere, but here `this` is used. 

If I understand correctly, I believe the difference that affords (`this` vs `super`) doesn't make a difference in your use case. If its all the same, sticking with `super` might be more clear (that you are using the parent's implementation), and consistent.

If I am totally off and not seeing the beans, please let me know!

src/utils/UIConfig/ProcessBuilderUIConfig.js
```diff
@@ -0,0 +1,40 @@
...
+
+export default class ProcessBuilderUIConfig extends StreamGroups {
+
+  static toggleGroupFolded(processBuilder, groupName) {
+    processBuilder.groups = super.toggleGroupFolded(processBuilder.groups, groupName)
+    return processBuilder
+  }
+
+  static toggleGroupStreamFolded(processBuilder, groupName, streamId) {
+    const groups = this.findGroupByName(processBuilder.groups, groupName)
```
DanielSeehausen, line 11: same comment re: this as above

src/utils/UIConfig/ProcessBuilderUIConfig.js
```diff
@@ -0,0 +1,40 @@
+import StreamGroups from './StreamGroups'
+
+export default class ProcessBuilderUIConfig extends StreamGroups {
+
+  static toggleGroupFolded(processBuilder, groupName) {
+    processBuilder.groups = super.toggleGroupFolded(processBuilder.groups, groupName)
+    return processBuilder
+  }
+
+  static toggleGroupStreamFolded(processBuilder, groupName, streamId) {
```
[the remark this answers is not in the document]
DanielSeehausen: even though process builder and graphs are almost perfect duplicates, I still like the duplication in this case as there is no guarantee the views remain so similar in the future (and they aren't even using the same component structure fully now)! The duplication will make it easy to diverge with the view if we do so later, so ++ from me for doing it like this and -- for codeclimate. You can mark this as an 'invalid' in code climate if you are so inclined.

O-Susanoo:
thank you @DanielSeehausen for the review i will apply your comment on the next pr

src/components/processviewer/body/dataviewer/DataGraphs.jsx
```diff
@@ -37,28 +38,41 @@ class DataGraphs extends React.Component {
...
+    uIConfig.graphs = GraphsUIConfig.toggleGroupFolded(graphs, groupName)
+    dispatchSetDeviceUIConfig(userId, activeDevice.id, uIConfig)
   }
 
   toggleStreamFolded = (groupName, streamId) => {
     const { activeDevice, dispatchSetDeviceUIConfig, userId } = this.props
-    const newConfig = UIConfig.toggleStreamFolded(activeDevice.uiconfig, groupName, streamId)
+    const {
+      uIConfig,
+      uIConfig: { graphs },
```
ConorTighe1995, line 53: Two vars the same name?

PR #649, opened 2020-04-02
466 view running process

### LINKED ISSUE #466: 'View Running Process' button displays even when a user does not have access to the given process
Update: Is this still happening or a problem? Try to recreate first before jumping in to fix!

Currently, this displays when a user hovers over a running device:
image

When clicked, they get a 'device not found' flash notification.

- [ ] As a user, if I do not have access to a running process, I should not see the specific hover effect/button to jump to the running process

original ticket : closes #470 (ake sure to close the original ticket, too, and follow up with ticket creator Jordi)

ConorTighe1995:
Only users with the correct access right can view a running process on a device with the only exception being admins.

I attempted to write a test that can check for hasAccess and check if the button rendered or not but couldn't see a way as the value is depended on the fetch after the component will render, there is probably a way to go about mocking this but I wasn't sure how. I could try export the fetchAndSetRunningData function that calls the fetch in useEffect and use jest.fn() function to mock it but I don't know if changing the implementation for a test would be justifiable here.

closes #466


DanielSeehausen:
(added patch as a reminder to me to put this in the end of sprint patch update)

DanielSeehausen:
> Only users with the correct access right can view a running process on a device with the only exception being admins.
> 
> I attempted to write a test that can check for hasAccess and check if the button rendered or not but couldn't see a way as the value is depended on the fetch after the component will render, there is probably a way to go about mocking this but I wasn't sure how. I could try export the fetchAndSetRunningData function that calls the fetch in useEffect and use jest.fn() function to mock it but I don't know if changing the implementation for a test would be justifiable here.
> 
> closes #466

@ConorTighe1995 would mocking the fetch response get you what you need in a case like this? That is a commonly used pattern, but I am unsure if it applies to your situation.

test/components/devices/DeviceList.test.jsx
```diff
@@ -26,11 +26,9 @@ describe('DeviceList.jsx', () => {
           <DeviceList
             noDevices={devices.length === 0}
             devices={devices}
-          >
-          </DeviceList>
```
DanielSeehausen, line 30: :laser: 👁

test/components/devices/DeviceWidget.test.jsx
```diff
@@ -0,0 +1,40 @@
...
+  )
+
+  describe('Rendering a device widget', () => {
+    test('Blank device widget renders and is ready for process', async () => {
+      const device = getRunningDevice()
+      const { getByText } = renderDeviceWidget(device)
+
+      getByText('ready for process')
+    })
+  })
```
DanielSeehausen, line 39: test looks awesome

not a blocking request for the PR, just a minor note about a cheap win:
because the test is set up well, it would be pretty cheap to expand the devices being imported from `mocks/deviceDataMock` to include something like `getIdleDevice`, etc., and then test for the other cases.

image

The logic for the different cases is here in DeviceMenu (currently it is in a component, and that's my oversight as I put it there, I should have put it on the interface/model with methods like `Device.isRunning(device)` 

image

ConorTighe1995: Cool! I use runningData a lot here in the tests I've added so maybe the mock folder could be used to bring it a step further, maybe the processDataMock? I could deconstruct some of the data from that.

ConorTighe1995:
> very nice, especially the tests!
> 
> Regarding the original issue for this: One part of this ticket was also to display 'You don't have access to this process' instead of 'Process not found' when users try to load a process they don't have the access to. Now by hiding the button, this can happen in only one way: A user manually inserts a processId of such a process in the url which is very unlikely, but could happen.
> 
> This problem should be solved in a different way however: When we load all processes (`devices/{deviceId}/processes`), we should load only a list representation of them and only when a user navigates to a specific process (`devices/{deviceId}/processes/{processId}` - which is the same component displayed when a user hits the 'View running process" button on the device widget), we should load the full process. The last step would be correct entry point for an access/not found check.
> 
> Because this requires a bigger refactoring, I'm going to create a new ticket for this: #651

I actually tested that but kept the error as I thought the flash message only applied when clicking the button and loading the page. But ill add that now.

PR #650, opened 2020-04-04
648 finalize the target line displays

### LINKED ISSUE #648: Finalize relatedDownstreams for target display
Firmware has implemented the `relatedDownstreams` key on some devices now. 
This allows to finalize the display of graph target lines

vdiezel:
closes #648 

Had to write my own implementation for the behaviour we wanted:

targetLine

We can't have transparancy and also a label on top of the line, but we get the advantage that the y-axis rescales, that the graph can display all targets at once in form of a step function and as a compensation for the missing label you can hover over the graph to see which target was set.

**I changed a big chunk of the logic in the later commits, so just review the final result**


src/components/processviewer/body/dataviewer/PhaseSelector.jsx
```diff
@@ -6,7 +6,7 @@ const PhaseSelector = ({ mode, selectedPhaseId, setSelectedPhase, activeProcess
   <Dropdown
     trigger={(
       <Button positive={mode === 'phase'}>
-        {getPhaseName(activeProcess, selectedPhaseId)}
+        {getPhaseName(activeProcess, selectedPhaseId) || 'Phase'}
```
vdiezel, line 9: I refactored the phase dropdown slightly so it doesn't display the unselected first phase as the default but displays the phase name only then, when a phase has been selected

src/components/processviewer/graphs/Graph.jsx
```diff
@@ -83,7 +81,6 @@ const Graph = ({
       gridThickness: 1,
     },
     toolTip: {
-      shared: true,
```
vdiezel, line 86: made it very difficult to find a point on the target line because the datapoint density is much higher on the stream data graph

src/utils/Stream.js
```diff
@@ -2,6 +2,17 @@ import StreamIcons from './StreamIcons'
 
 export default class Stream extends StreamIcons {
 
-  // currently only using methods in Streamicons. I didn't want to burden whoever implements an actual stream operation with having to abstract that, so I did it and left this here for you, future programmer who found their way here.
```
vdiezel, line 5: image

src/utils/TargetLine.js
```diff
@@ -0,0 +1,107 @@
...
+    // and if trimmed insert a new point at the cut edge
+
+    const dataPoints = this.trimLeft(vertices, startFrom)
+    if (dataPoints.length !== vertices.length) {
+      dataPoints.unshift({ x: startFrom, y: dataPoints[0].y })
+    }
+
+    return this.createTargetLine(dataPoints)
+  }
+}
```
vdiezel, line 105: Main function - start here to understand what is going on. In short
1. Find corners of the target line step function by parsing the start times of the progressed phases
2. fill the gaps in between to allow seing the tooltip when hovering over it

src/utils/TargetLine.js
```diff
@@ -0,0 +1,107 @@
...
+    // so the target values appear on hover at least on some spots
+    const totalTime = endTime - startTime
+    const interval = this.FILLER_INTERVAL
+    const numberOfPoints = Math.floor(totalTime / interval)
+    const points = Array.from({ length: numberOfPoints }, (_, i) => (
+      { x: startTime + interval * (i + 1), y: value }))
+
+    if (points.length === 0) points.push({ x: startTime + totalTime / 2, y: value })
+    return points
+  }
```
vdiezel, line 60: I will chage this to match the viewport probably instead of using a fixed interval

src/utils/TargetLine.js
```diff
@@ -0,0 +1,107 @@
...
+      showInLegend: true,
+      legendText: 'target',
+      lineColor: '#5eafed',
+      color: '#5eafed',
+      lineDashType: 'dash',
+      dataPoints,
+    }
+  }
+
+  static secondsSinceProcessStart(activeProcess, reference = Date.now()) {
```
ConorTighe1995, line 24: The reference is always the `reference = Date.now()` or have you passed it as a param so you can get the seconds from start according using a set time too? if its always the system time why not do it inside the function:

```
static secondsSinceProcessStart(activeProcess, reference = Date.now()) {
    return (reference - activeProcess.startedAt) / 1000
}
```

src/utils/TargetLine.js
```diff
@@ -0,0 +1,107 @@
...
+    return 600
+  }
+
+  static createTargetLine(dataPoints) {
+    return {
+      markerSize: 0,
+      type: "line",
+      showInLegend: true,
+      legendText: 'target',
+      lineColor: '#5eafed',
```
ConorTighe1995, line 17: To add to the discussion we had in the chat, do you think it could be worth adding a object for customization here to give the user more control of the target lines?

```
static createTargetLine(dataPoints,options) {

    return {
        ...
        lineColor: !!options.lineColor ? options.lineColor : '#5eafed',
        ...
     }
```

ConorTighe1995: That code may not work but its just to demonstrate the idea

src/components/processviewer/body/dataviewer/Stream.jsx
```diff
@@ -14,7 +16,7 @@ import { ConfigForm, NoDataConfigForm } from './ConfigForm'
 export default class Stream extends React.Component {
 
   state = {
-    selectedPhaseId: getFirstPhaseId(this.props.activeProcess),
+    selectedPhaseId: undefined,
     mode: (this.props.activeProcess.state === 'running' || this.props.activeProcess.state === 'paused') ? 'live' : 'full',
```
ConorTighe1995, line 20: I would deconstruct props here as its a really long line.

src/components/processviewer/graphs/Graph.jsx
ConorTighe1995: Is this more to do the the layout of the graph or the source its self? This could be an indicator to use different types of graphs for different metrics in the future, I dont know if other types of graphs would be better suited for this but could be worth looking into.

vdiezel: There is a step function for CanvasJS which could be used here, too, but it didn't bring any advantages. 
The tooltip option here changed the behaviour so that it would automatically jump to the closest data point based JUST on the x-Value of the mouse position. Because the real graph has way more data points, the tooltip will most likely select those points when you hover through the graph and it makes it very difficult to find a point on the target line. 

By removing this option the tooltip will also consider the mouse position on the y-Axis - so when you are closer the target line it will actually find the points there first. 

**Best way to see whatn is going on is actually just to pull this and toggle this flag ;)**

src/utils/TargetLine.js
vdiezel: I was thinking about it - but then I thought YAGNI. It can be added very easily using an options objkect with default parameters. Easiert would be using key overwriting with object destructuring, but the safer method would be your proposal!

src/utils/TargetLine.js
vdiezel: it is not always now ;) - see ll. 34

src/components/processviewer/body/dataviewer/Stream.jsx
vdiezel: I can't without using a function here. but good :eye:

src/utils/Stream.js
DanielSeehausen: lol feels good man

src/components/processviewer/body/dataviewer/Stream.jsx
ConorTighe1995: can you not put in in a const above state?

src/utils/TargetLine.js
ConorTighe1995: Sounds good! Yep would be more elegant I just wrote that out off the top off my head but looks a bit mad for a declaration haha

Maybe something we can think about for a future ticket if there is a use.

src/utils/TargetLine.js
ConorTighe1995: My bad! good stuff

src/components/processviewer/body/dataviewer/Stream.jsx
DanielSeehausen: This is a language feature that isn't official yet - though its at least in node >12

```js
class Ada {
  someInstanceVar = { name: 'lovelace' }
}
```

I don't/didn't know if it can be used with props - I'm going to try now

DanielSeehausen: I might be missing how you are intending it to be destructured, but here is what I got in Node 13 (which class fields are active in), in which it wasn't possible:

image

test/utils/TargetLine.test.js
```diff
@@ -120,14 +126,14 @@ describe('TargetLine', () => {
...
-      const { process, upstream, targets } = createSetup()
+      const { process, upstream, targets, streamData } = createSetup()
       process.state = 'finished'
       const finishedAt = process.progression[process.progression.length - 1].startTime + 100000
       process.finishedAt = finishedAt
       const lastTimeStamp = (finishedAt - process.startedAt) / 1000
       const range = { start: 0, end: lastTimeStamp }
 
-      const { dataPoints } = TargetLine.getTargetLines(upstream, process, range)
+      const { dataPoints } = TargetLine.getTargetLines(upstream, process, range, streamData)
```
ConorTighe1995, line 136: this is not really a strict rule or anything but personally when I get to 4 parameters on a function I start to ask myself does this data deserve its own object, this can increase readability by describing bundles of data and also make testing easier.

bob martin talks about it in clean code:
https://www.matheus.ro/2018/01/29/clean-code-avoid-many-arguments-functions/

src/utils/TargetLine.js
```diff
@@ -0,0 +1,105 @@
+import Stream from 'utils/Stream'
+import { Process } from 'utils/process'
+
+
+class TargetLine {
+
+  static get FILLER_INTERVAL() {
+    return 600
+  }
```
DanielSeehausen: <3

PR #661, opened 2020-04-07
440 widget space

### LINKED ISSUE #440: Optimize phase widget usage of space
As a user, my eyes are pleased with a pleasant experience when it comes to phase widget's real estate usage:

Phase widgets should make better use of their space. There is extra use space on the right here. I believe its because the top title bar in the app is pushing it open:

image

### LINKED ISSUE #634: Redesign UI for displaying phases so it does not effect the screen height of the browser window
As process phases are added the UI will continue to expand the height of the page which can lead to a unaesthetic and awkward blank space at the bottom of the UI. The addition of phases should not effect the overall page height, this could be achieved by putting the phases inside a container and making it scroll able.

image

ConorTighe1995:
This PR makes better uses of the space to the right of the phase list by extending the column in the grid containing phases.I also fixed the bug that creates an infinite list height when adding phases. I added some styling to the scroll bar of the container that stores the phases, let me know what you think.

closes #440 
closes #634


ConorTighe1995:
Is there an intention to add something else to the tree besides phases in the future? otherwise the popup just seems to add an extra click. Also is there a way to delete a phase if I create one by mistake?

src/components/processviewer/phasesdisplaybar/PhasesNavbarWrapper.jsx
```diff
@@ -20,11 +22,11 @@ const mapDispatchToProps = dispatch => ({
 
 class PhasesNavbarWrapper extends React.Component {
 
-  state = { showPopup: false }
-
-  handleOpen = () => this.setState({ showPopup: true })
+  messagesEndRef = React.createRef()
```
O-Susanoo, line 25: this one i think it would be better if it was a const

O-Susanoo, line 25: i think this line inside a constructor would be much more clearer but no need to change this

PR #662, opened 2020-04-07
550 derive event names from display name

### LINKED ISSUE #550: Event's should now defer to the event payload's displayName
Ping daniel or viktor on how to fire fake events from IoT :)

Previously, event data objects had no displayName key. 
This was recently added in Hambda. Now, all events should defer to using the `displayName` key that comes with the event payload.

Instead of getting rid of what is in place now, I would consider having that as a fall back in case `displayName` comes through as anything but a non-empty string!

ConorTighe1995:
The newly added Event.js file in utils is used to parse the display names of emitted events from hambda. Reactor receives the events, then uses a function called deriveDisplayName to ensure event names are displayed using title casing and sentence casing. If both name and displayName are empty we render Unnamed Event on the screen. If display name is empty we use the name and parse the string, applying title casing and sentence casing.

closes #550


src/utils/Event.js
```diff
@@ -0,0 +1,15 @@
+import { titleCase } from 'title-case'
+import { sentenceCase } from 'sentence-case'
+
+class Event {
+
+  static deriveDisplayName(event) {
+    if (!event.name && !event.displayName) return 'Unnamed Event'
+    let derivedName = event.displayName ? event.displayName : event.name
```
vdiezel, line 8: can be a `const`

ConorTighe1995: ill change that now

PR #669, opened 2020-04-09
636 display name wrap and bonus

### LINKED ISSUE #636: Device widgets display name wrap AND bonus bug
When device names are long enough to warrant multiple lines, they break in the manner seen in the first image below.

Separately, when a user is editing a device name, and they click into the name field while it is under edit, (see second image below), it clicks through to the device view. We need to `stopPropagation` on that click, or something similar!

## Deliverables
- Fix the text wrap to either be left justified, or some other option that is aesthetically pleasing/doesn't disrupt the relationship to the line below
- As a user, when I click into the text edit field, it should only move the cursor line for editing text (or highlighting text), and it should not click through to the device view

image

---

image

ConorTighe1995:
This should fix the label alignment and the label navigation bug, I also threw in a simple test.

closes #636


DanielSeehausen:
(auto updated branch)

ConorTighe1995:
@vdiezel

testheader

PR #675, opened 2020-04-10
660 bring them event triggers

### LINKED ISSUE #660: Activate eventTriggers on the frontend!
event triggers are currently deactivated on the frontend. 
Bring them back to life. 
This includes: 

The device/setup component should render the `Stream Event Triggers` section at the bottom and allow creating triggers

In the Process Builder, those downstreams which are linked to an upstream should expose the button to set event triggers. 

As event triggers are considered non adhoc changes (just like renaming phases) they currently conflict with adhoc changes. Which means it shouldn't be possible to change event trigger in a running process when adhoc changes are staged.

vdiezel:
We are happy to present you the next extension of our alert system: user defined sensor alerts. All device sensors which send data to the user interface can have alerts associated with them based on their value. This feature entails: 
The added support  displaying alert thresholds:

  - support for triggering alerts when the sensor value crosses a user defined lower or upper threshold
  - displaying the set thresholds in the data graph of the sensor

alert_threshold_1

  - creation of default device alerts that are automatically added to every process in the device setup page

alerts_devices

  - process specific alert configuration in the new alerts section of a process at any time

triggers_in_process

  - creation of a log entry for every triggered alert which can be viewed in the device and process logs

alerts_logs

  - displayig a flash message to the user when alerts are triggered and gathering them in the alerts overview page accessible via the sidebar

alert
 
For the time being, alerts are fired every 30 minutes until the issue has been resolved. We are planning to provide a richer customization including user defined alert intervals, rules for resolving them and much more! 

\private
closes #660 

This brings the event triggers to the UI. I changed the way we designed them in the beginning. Because we wanted them to be phase specific we pushed them into the process description. Because only downstreams are listed there, we could only create triggers for upstreams which are related to downstreams. 

I thought this is an unnecessary restriction and it would be worth to trade the phase association for that. So I created a new top level key `eventTriggers` on process (`123456` is a streamId): 

```js
eventTrigger: {
  "123456": {
    "UPPER_LIMIT": { ... },
    "LOWER_LIMIT": { ... },
  }
}
```

We can bring the phase association back by adding more keys but I think most customers are goig to use those alerts phase unspecific but rather for general alerts - at least that is what Celltainer expects of them. Lets wait for the request to make them phase specific.

The are two places where you can set those triggers: 
1. In the device setup. These triggers will be automatically added to any process created on this device. 
2. In the process you can add, modify and remove triggers. Also during a running process (they are not considered adhoc changes, so you dont have to commit them in a running process)

Setting triggers in device:

device:triggers

What you see in the process then
process_triggers

How they are triggered (not spoof here, this is real)
firing_trigger

I also created a new flash message for alerts that uses the alarm bell icon an prepended the device name so the user can tell which device threw it:
Bildschirmfoto von 2020-04-10 20-04-40

And adds a toggle to display threshold lines in the graph: 
alert_threshold


src/components/device/modals/CreateProcessModal.jsx
```diff
@@ -58,6 +58,7 @@ class CreateProcessModal extends React.Component {
       comment: processComment,
```
vdiezel, line 58: in this PR I made the decision to move the eventTriggers to a top level key on the process.

src/components/device/setup/DeviceSetup.jsx
```diff
@@ -60,6 +60,7 @@ const DeviceSetup = props => {
         <Divider/>
         <StreamEventTriggerTable
           activeDevice={props.activeDevice}
+          mode='device'
```
vdiezel, line 63: Im resuing the same component for the triggers using the `mode` flag

src/utils/DeviceConfig.js
```diff
@@ -284,6 +284,20 @@ class DeviceConfig {
     return eventTriggers.filter(trigger => trigger.type !== type)
   }
 
+  static getStreamEventTriggersForProcess = device => {
```
vdiezel, line 287: normalizes the event triggers defined on the device config into the fromat they are stored on the process

src/utils/validation.js
```diff
@@ -139,26 +137,6 @@ export function validateStreamTargets(process, device) {
...
-
-          validationErrors.push({
-            type: "StreamEventTriggerError",
-            phaseName: phase.name,
-            message: `Value in group ${stream.groupName} at stream ${stream.name} has invalid event trigger parameters.
-              Provided target of ${displayedTarget} ${replaceNoUnitCharacter(displayedUnit)},
-              but event ${EventTrigger.description(type)} of ${displayedLimit} ${replaceNoUnitCharacter(displayedUnit)}`,
-          })
-        }
-      })
```
vdiezel, line 161: with the triggers applicable to all upstreams the validation could only test triggers which do have a related target. This is only a very small part of all upstreams, so for now we delegate this responsibility to the user. Because alerts can be deleted easily and trigger only once every 30 minutes this shouldn't be an issue.

src/utils/EventTrigger/UpperLimitTrigger.js
```diff
@@ -18,7 +18,7 @@ class UpperLimitTrigger {
     ]
   }
 
-  static suitableStreams = device => Device.getDownStreamsWithLinkedUpstream(device)
+  static suitableStreams = ({ streams }) => Streams.getAllLinkedTargetStreams(streams)
 
   static invalidParameters = (trigger, target) => {
```
ConorTighe1995, line 24: invalidParameters here is almost identical to invalidParameters in LowerLimitTrigger.js, the classes help explain the functionality but its still very similar code with a opposite outcome , maybe give them distinct names to help with readability.

src/utils/Streams.js
```diff
@@ -0,0 +1,35 @@
...
+      .filter(relation => relation.relationType === 'target')
+      .map(({ streamId, streamID }) => this.getById(streams, streamId || streamID))
+      .filter(stream => stream)
+  }
+
+  static isTrackedByUpstream(streams, streamId) {
+    return streams
+      .filter(stream => stream.type === 'upstream')
+      .filter(stream => stream.relatedDownstreams && stream.relatedDownstreams.length)
+      .some(stream => stream.relatedDownstreams
```
ConorTighe1995, line 37: Maybe encapsulate some and filter into two private functions that isTrackedByUpstream uses to describe exactly whats being done with streams here to get the result.

src/redux/reducers/reducer.js
```diff
@@ -94,10 +94,9 @@ function rootReducer(state = initialState, action) {
   else if (action.type === ACTIONS.SET_STREAM_EVENT_TRIGGER){
 
     const { processes } = state
-    const { processId, phaseId, streamId, trigger } = action.payload
+    const { processId, streamId, trigger } = action.payload
 
-
-    const updateFn = process => Process.updateStreamEventTriggers(process, phaseId, streamId, trigger)
+    const updateFn = process => Process.updateStreamEventTrigger(process, streamId, trigger)
```
ConorTighe1995: What does Fn stand for

src/components/processviewer/body/dataviewer/Stream.jsx
```diff
@@ -124,11 +130,12 @@ export default class Stream extends React.Component {
     } = this.props
 
     const { lastDataPoint, displayedGraphData } = GraphTools.getDisplayedStreamData({
-      stream, streamData, selectedUnit, mode, selectedPhaseId, activeProcess, showTargetLine, zoomParams
+      stream, streamData, selectedUnit, mode, selectedPhaseId, activeProcess, showTargetLine, zoomParams, showEventTriggerLines,
```
ConorTighe1995, line 133: a lot of params for one line

src/utils/GraphTools.js
```diff
@@ -156,35 +156,64 @@ class GraphTools {
...
+        streamData,
+        displayedGraphData,
+        stream,
+        activeProcess,
+        zoomParams,
+        lineType: 'eventTrigger',
+      })
+
+      if (triggerLines && triggerLines.length) {
+        triggerLines.forEach(line => displayedGraphData.push(line))
```
ConorTighe1995, line 204: Does array mutation matter here? do we want our display graph to be immutable? 

https://lorenstewart.me/2017/01/22/javascript-array-methods-mutating-vs-non-mutating/

src/redux/reducers/reducer.js
vdiezel: function!

src/utils/EventTrigger/UpperLimitTrigger.js
vdiezel: If I would give those functions different names, based on their type (`UPPER_LIMIT`, `LOWER_LIMIT`) then the caller of the interface (mostly `EventTrigger` which delegates to the right trigger interface like `UpperLimitTrigger.js`) would need to have knowledge which function to invoke based on the trigger type. I don't want this - I use something called ducktyping which means I assume every trigger interface will have a `invalidParameters` function! This makes it scalable and robust. Have a look at the `EventTrigger` interface, maybe it becomes clear there ;)

Sandy Metz has a good part about it the Ruby book we gave you (POODR2).

src/components/processviewer/body/dataviewer/Stream.jsx
vdiezel: I agree

src/utils/GraphTools.js
vdiezel: I didn't use array destructuring here because I worried about the performance, thats why I mutate the array.

src/components/processviewer/body/builder/Group.jsx
```diff
@@ -40,23 +38,10 @@ const Group = ({
     />
   )
```
O-Susanoo, line 39: i know that this is not in pr but is it normal that this button is empty

src/redux/helper/subscribeHandlers/eventHandler.js
```diff
@@ -1,26 +1,39 @@
...
 function handleRevokedAlert(revokedId) {
   store.dispatch(removeUserAlertById({ alertId: revokedId }))
 }
 
+function getMessagePrefix(deviceId) {
+  const { devices } = store.getState()
+
+  const deviceName = devices && deviceId
+    ? Device.getById(devices, deviceId)?.name
+    : undefined
```
O-Susanoo, line 21: Nice

src/utils/EventTrigger/EventTrigger.js
```diff
@@ -21,8 +21,8 @@ export default class EventTrigger {
     }
   }
 
-  static getSuitableDeviceStreams = (type, activeDevice) => (
-    this.directToType(type).suitableStreams(activeDevice)
+  static getSuitableDeviceStreams = (type, streams) => (
+    this.directToType(type).suitableStreams(streams)
```
O-Susanoo, line 25: i don't to mean picky but i think that getSuitableDeviceStreams was better before because it received a device in argument

src/utils/GraphTools.js
```diff
@@ -156,35 +156,64 @@ class GraphTools {
...
+      case 'target':
+        return TargetLine.getTargetLines(stream, activeProcess, { start, end }, streamData)
+      case 'eventTrigger':
+        return TargetLine.getEventTriggerLines(stream, activeProcess, { start, end }, streamData)
+      default:
+        return []
+    }
+  }
+
+  static getDisplayedGraphData({ streamData, mode, selectedPhaseId, activeProcess, unitConverter, stream, showTargetLine, zoomParams, showEventTriggerLines }) {
```
O-Susanoo, line 175: i know this one is my fault , it would have been grat to have parameters on multiple lines

src/utils/GraphTools.js
O-Susanoo, line 204: i guess this one could also go like this
```
displayedGraphData.push(...triggerLines)
```

test/utils/EventTrigger.test.js
```diff
@@ -44,37 +44,29 @@ describe('EventTrigger', () => {
       test('return suitable streams', () => {
         const fakeDevice = TDG.generateDevice()
 
-        const expected = fakeDevice.streams.filter(stream => stream.type === 'downstream')
-        const res = EventTrigger.getSuitableDeviceStreams(type, fakeDevice)
+        const expected = fakeDevice.streams.filter(stream => stream.type === 'upstream')
+        const res = EventTrigger.getSuitableDeviceStreams(type, fakeDevice.streams)
```
O-Susanoo, line 48: how did went from downstream to upstream

src/components/device/modals/CreateProcessModal.jsx
DanielSeehausen: good decision!

src/components/processviewer/body/builder/Group.jsx
DanielSeehausen: that might just be because github was collapsing the view for you @O-Susanoo 
here is an image expanded
image

src/components/streameventtrigger/StreamEventTriggerModal/StreamEventTriggerModal.jsx
```diff
@@ -1,33 +1,16 @@
 import React from 'react'
```
DanielSeehausen, line 1: while we don't have many, our directories also seem to follow a casing convention:

if it is named after the main class it is meant to export: pascal case
image

otherwise, camelcase:
image

src/components/utility/FlashMessage.tsx
```diff
@@ -11,7 +11,7 @@ const mapDispatchToProps = (dispatch: Dispatch<AnyAction>, ownProps: OwnProps):
 })
 
 export type Message = {
-  type: 'warning' | 'error' | 'info' | 'success'
+  type: 'warning' | 'error' | 'info' | 'success' | 'alarm',
```
DanielSeehausen, line 14: NICE

well written code,
easy to extend!

src/redux/helper/subscribeHandlers/eventHandler.js
```diff
@@ -1,26 +1,39 @@
 import FlashMessenger from 'utils/FlashMessenger'
+import { Device } from 'utils/device'
 import store from 'redux/store'
 import { addUserAlert, removeUserAlertById } from 'redux/actions/actions'
 
 const EVENT_TYPE_TO_FLASH_TYPE = {
   notification: FlashMessenger.info,
   warning: FlashMessenger.warning,
-  alarm: FlashMessenger.error,
+  alarm: FlashMessenger.alarm,
```
DanielSeehausen, line 9: is it intentional that all events of alarm level go to an alarm bell?

src/utils/DeviceConfig.js
```diff
@@ -284,6 +284,20 @@ class DeviceConfig {
...
+    const connectedStreamsWithTriggers = this
```
DanielSeehausen, line 288: very minor thing, but preferencing calling through the interface (`DeviceConfig`) instead of `this` might one day save you a headache. I believe it should also be more functionally faithful

src/utils/EventTrigger/EventTrigger.js
```diff
@@ -21,8 +21,8 @@ export default class EventTrigger {
     }
```
DanielSeehausen, line 21: out of range of this PR, the `directToType` above jumps out as type checking. A flag that maybe we should be interfacing with the children directly, and that if there is functionality in this class that they need, they should inherit/compose it from this class

src/utils/EventTrigger/LowerLimitTrigger.js
```diff
@@ -9,26 +10,31 @@ class LowerLimitTrigger {
...
   }
 
-  static displayParameters = trigger => `limit: ${trigger.limit}`
+  static displayParameters = (trigger, streamUnit, displayedUnit) => {
 
-  static create = ({ limit }, eventType) => ({ type: this.type, limit, eventType })
+    const converter = convertFrom(streamUnit).to(displayedUnit)
+
+    return `threshold: ${converter(trigger.threshold)}`
+  }
```
DanielSeehausen, line 35: this is a duplicate of UpperLimitTrigger

src/components/processviewer/body/builder/Group.jsx
O-Susanoo: Thanks (github can get sneaky sometimes)

src/utils/EventTrigger/EventTrigger.js
```diff
@@ -33,8 +33,8 @@ export default class EventTrigger {
     this.directToType(type).description
   )
 
-  static displayParameters = eventTrigger => (
-    this.directToType(eventTrigger.type).displayParameters(eventTrigger)
+  static displayParameters = (eventTrigger, streamUnit, displayedUnit) => (
+    this.directToType(eventTrigger.type).displayParameters(eventTrigger, streamUnit, displayedUnit)
```
DanielSeehausen, line 37: having a method that returns the appropriate interface (given an `eventTrigger.type`) would make this a little more maintainable as we could then inherit/compose this class in the children, and not need to delegate: `Parent.x` -> `Child.x` for every divergent method (see review comment though)

src/utils/EventTrigger/UpperLimitTrigger.js
DanielSeehausen: g4m3r use of ducktyping

src/utils/EventTrigger/UpperLimitTrigger.js
```diff
@@ -9,26 +10,31 @@ class UpperLimitTrigger {
...
 
-  static suitableStreams = device => Device.getDownStreamsWithLinkedUpstream(device)
+  static suitableStreams = streams => Streams.getUpstreams(streams)
 
   static invalidParameters = (trigger, target) => {
-    const { limit } = trigger
-    if (target >= limit) return true
+    const { threshold } = trigger
+    if (target >= threshold) return true
     return false
```
DanielSeehausen, line 27: `return target >= threshold` might be the same thing

src/utils/GraphTools.js
```diff
@@ -156,35 +156,64 @@ class GraphTools {
...
 
-  static getDisplayedGraphData({ streamData, mode, selectedPhaseId, activeProcess, unitConverter, stream, showTargetLine, zoomParams }) {
+  static createLinesByType({ streamData, displayedGraphData, stream, activeProcess, zoomParams, lineType }) {
+    const { start, end } = this.getDisplayedRange({ zoomParams, displayedGraphData })
+
+    switch (lineType) {
+      case 'target':
+        return TargetLine.getTargetLines(stream, activeProcess, { start, end }, streamData)
+      case 'eventTrigger':
+        return TargetLine.getEventTriggerLines(stream, activeProcess, { start, end }, streamData)
```
DanielSeehausen, line 169: naming is colliding a little bit here
`TargetLine` now owns target line stuff as well as event trigger lines
would a generic line, which TargetLines and EventTriggerLines inherit/compose/reference, make sense?

src/utils/Streams.js
vdiezel: will do!

test/utils/EventTrigger.test.js
vdiezel: first implementation was only a spoof before firmware caught up.

src/redux/helper/subscribeHandlers/eventHandler.js
vdiezel: It was intentional - I didn't think of the possibility that the alarm icon needs to be different based on the alarm. I think for now it is fine to show the bell for all alarms.

src/utils/EventTrigger/UpperLimitTrigger.js
vdiezel: :facepalm:

src/components/processviewer/body/dataviewer/Stream.jsx
vdiezel: fixed!

src/utils/Streams.js
vdiezel: done!

src/utils/GraphTools.js
vdiezel: fixed!

src/utils/GraphTools.js
vdiezel: didnt know you could do that! thanks, it works!

src/utils/DeviceConfig.js
vdiezel: done!

src/utils/EventTrigger/EventTrigger.js
vdiezel: fixed! renamed to `getSuitableStreams` in EventTrigger

src/components/streameventtrigger/StreamEventTriggerModal/StreamEventTriggerModal.jsx
vdiezel: camle cased it

src/utils/GraphTools.js
vdiezel: I renamed `TargteLine` to Linegenerator for now.

src/utils/EventTrigger/UpperLimitTrigger.js
ConorTighe1995: That makes sense! I didn't realize we were using duck typing here just seemed like a copy and paste when I looked at the file at the time

src/utils/EventTrigger/EventTrigger.js
vdiezel: do I understand you here: You think of something like 
```js
EventTrigger.get(type) {} // which wil return a child inteface
```
So that it can be used externally like this: 
```js
EventTrigger.get('UPPER_LIMIT').displayParameters(streamUnit, displayedUnit) {}
```

src/utils/EventTrigger/LowerLimitTrigger.js
vdiezel: once I know exactly how to go around this, they can inherit from an `OutOfBoundaryTrigger` parent class like they do on the backend.

O-Susanoo:
Hi , just to make lifer easier for me :) can you please rebase on master and commit again because i think there are files that are related to nexus thus making the review a bit harder , rebase -i should be your guy on this one

PR #679, opened 2020-04-14
360 404 page

### LINKED ISSUE #360: Provide a 404 when a user tries to access a resource that doesnt exists
e.g. when a user changes the deviceId in the url and reloads, we get our 500 error screen showing up and the error is being reported to our client error API. We shoudl provide a 404 and not report anything instead.

add 404 ospin page just like we have OOPSIN

ConorTighe1995:
A 404 page added so when a user enters a device id that doesn't exist the appropriate error will be displayed. The 500 page is now the default error page but this can be changed as the error pages are expanded on.

**NEW AND IMPROVED:**

closes #360


src/components/device/Device.jsx
```diff
@@ -127,6 +131,14 @@ class Device extends React.Component {
 
     const { devices, match } = this.props
     const { isLoading } = this.state
+    const activedevice = DeviceUtils.getById(devices, match.params.deviceId)
+
+    if (!activedevice) {
+      this.props.history.push({
```
vdiezel, line 137: get history from the props in line 132

src/components/utility/Report.jsx
```diff
@@ -34,34 +34,46 @@ const RenderStackTrace = ({ errorString, lineNumber, functionName, fileName}) =>
...
+            <Grid.Column>
+              <svg viewBox='0 0 400 70' style={{ transform: 'scale(0.8)' }}>
+                <polyline
+                  fill='none'
+                  stroke='#92312d'
+                  strokeWidth='2'
+                  points='0,45.486 38.514,45.486 44.595,33.324 50.676,45.486 57.771,45.486 62.838,55.622 71.959,9 80.067,63.729 84.122,45.486 97.297,45.486 103.379,40.419 110.473,45.486 400,45.486'
+                />
+                <text style={{ fontSize: '9px' }} fill='#92312d' x='115' y='35'>Something went wrong! The error has been reported to our developers.</text>
+              </svg>
```
vdiezel, line 68: In case of 404 -> can we get a differenct message, some thing: 
`The resource you requested could not be found`.
We also don't need to send an error report in this case.

ConorTighe1995:
Apologies about the commit history, the first few are where the changes that matter then the others were me trying to make sure nexus didn't break in my code when I did a rebase.

DanielSeehausen:
> This is heading in the right direction.
> I have a small request reagrding the error text in case of 404 and a slightly bigger one re: the bahaviour when a 404 occurs: a 404 is not an error that should be reported (see comment), so we don't need to forward to the `/report` route. We could have a default route for this.
> 
> the correct behaviour would be this: we actually keep the url with the unkwon resource Id and display the 404 screen with a button to go back e.g. `/devices` but I don't think that is actually possible with React router because even a wrong deviceId will match the route like `/device/{worngId}` and will try to render the device component again. The only way to bypass this is actually to return the 404 screen instead of the device view within the render function of `Device.jsx` and not using history push.
> 
> The alternative is to have designated route for 404, something like `/not-found`.
> 
> I think I would prefer the first solution - and adding additionally a fall back route to the router that also renders the same 404 component when a route completely unknown to our router is attempted to load.
> 
> But I would like to hear what you think first, @DanielSeehausen and @O-Susanoo .

Agree on it not being an error!
I don't know if there is a specific correct use, but using GitHub as the north star, they keep the URL and display 404. 

image

test/components/utility/ErrorPage.test.jsx
```diff
@@ -0,0 +1,39 @@
...
+        <ErrorPage
+          history={history}
+          errorCode={errorCode}
+        />
+      </Router>
+      ,
+    ))
+
+  describe('Test error pages ', () => {
+    test('display header', async () => {
```
DanielSeehausen, line 26: this test description diverges from the test assertions!

src/components/utility/ErrorPage.jsx
```diff
@@ -0,0 +1,65 @@
...
+          <Error404 />
+        </ErrorContainer>
+      )
+    default:
+      return (
+        <ErrorContainer history={history}>
+          <Error404 />
+        </ErrorContainer>
+      )
+  }
```
DanielSeehausen, line 40: This could be described with a little less duplication where the container wraps the specific component:

```jsx
<ErrorContainer history={history}>
  deriveErrorComponentFromCode(errorCode) / ERROR_CODE_TO_COMPONENT[errorCode]
</ErrorContainer>
```

src/components/utility/ErrorPage.jsx
```diff
@@ -0,0 +1,65 @@
...
+          <Grid.Column>
+            { children }
+          </Grid.Column>
+        </Grid.Row>
+
+        <Grid.Row centered>
+          <Grid.Column verticalAlign='middle'>
+            <br />
+            <br />
+            <Button onClick={() => history.push('/')} primary size='big'>Go to devices</Button>
```
DanielSeehausen, line 57: would there be any issue explicitly setting this to `devices`? I know we match `/` on devices right now, but that might change in the future. If we set it to `/devices` here, we don't run the risk of accidentally diverging from what the button says it does later on.

test/components/utility/ErrorPage.test.jsx
```diff
@@ -0,0 +1,39 @@
...
+      const expectedResult = '/'
```
DanielSeehausen, line 27: same request above with the `/device` as the specific route

src/components/main/Main.jsx
```diff
@@ -131,7 +133,14 @@ class Main extends React.Component {
...
             <Route path='/device/:deviceId' component={Device} />
-            <Route path='/' component={Devices} />
+            <Route
+              to={{ errorCode: '404', pathname: '/404' }}
+              component={ErrorPage}
+            />
+            <Redirect
+              from='*'
+              to='/404'
+            />
```
DanielSeehausen, line 143: I am a react router noob so please bear with my question:

will these replace the pathname to 404? If so, I agree with @vdiezel that we should keep the path they entered and not replace it with 404 or anything like that

what is the difference here and the following:
```
    <Route>
      <ErrorPage errorCode='404' />
    </Route>
```

src/components/utility/ErrorPage.jsx
DanielSeehausen: or something similar. Not a requested change

src/components/main/Main.jsx
DanielSeehausen: referencing this guy:
https://reacttraining.com/react-router/web/api/Switch

test/components/utility/ErrorPage.test.jsx
ConorTighe1995: missed that, thanks ill update now

src/components/device/Device.jsx
```diff
@@ -135,7 +145,7 @@ class Device extends React.Component {
         <Grid columns={2}>
           <Grid.Row>
             <DeviceMenu
-              activeDevice={DeviceUtils.getById(devices, match.params.deviceId)}
```
DanielSeehausen, line 138: 🔫

ConorTighe1995:
> I ran this local host and i'm getting this behavior:
> 
> Apr-17-2020 12-42-58
> 
> It seems the 404 from devices is working right, but the 404 for any failing routes in general is acting a little different. Attempt the following route, for example:
> http://localhost:3000/device/a8207dab-43ab-44be-93eb-308a60df1f29/processes/ff

I actually did that intentionally as if a process doesn't exist then it will 404 but I can change it back

vdiezel:
> I ran this local host and i'm getting this behavior:
> 
> Apr-17-2020 12-42-58
> 
> It seems the 404 from devices is working right, but the 404 for any failing routes in general is acting a little different. Attempt the following route, for example:
> http://localhost:3000/device/a8207dab-43ab-44be-93eb-308a60df1f29/processes/ff

Because for all routes the router know of, e.g. `/processes` etc. we will need to provide either `exact` route matching or in case of single resources like `process/{some-unknow-ID}` we will need to manually check if the resource can be found.

DanielSeehausen:
> > I ran this local host and i'm getting this behavior:
> > Apr-17-2020 12-42-58
> > It seems the 404 from devices is working right, but the 404 for any failing routes in general is acting a little different. Attempt the following route, for example:
> > http://localhost:3000/device/a8207dab-43ab-44be-93eb-308a60df1f29/processes/ff
> 
> Because for all routes the router knows of, e.g. `/processes` etc. we will need to provide either `exact` route matching or in case of single resources like `process/{some-unknow-ID}` we will need to manually check if the resource can be found.

would it be possible to do the same default case of 404 in the sub routers like was done in the top level router?

ConorTighe1995:
> I ran this local host and i'm getting this behavior:
> 
> Apr-17-2020 12-42-58
> 
> It seems the 404 from devices is working right, but the 404 for any failing routes in general is acting a little different. Attempt the following route, for example:
> http://localhost:3000/device/a8207dab-43ab-44be-93eb-308a60df1f29/processes/ff

So I actually looked into that and decided this was to do with redirecting in the wrong place, instead of rendering a error page that responsibility is in main and the redirect now happens in Device.jsx. Im going to refactor this quickly as I think we can actually improve routing even more here.

ConorTighe1995:
I did a decent bit of refactoring and added a gif showing custom 404 pages for routes, I have a gif of the new routes too. My internet was a bit messed up so sorry about the lag in it.

src/components/utility/Report.jsx
```diff
@@ -34,34 +29,14 @@ const RenderStackTrace = ({ errorString, lineNumber, functionName, fileName}) =>
...
-    {errorObject != null ? RenderStackTrace(errorObject) : ''}
-  </Container>
-)
+const Report = props => {
+  const { location: { errorObject }, history } = props
+  return (
+    <Container>
+      <ErrorPage history={history} errorCode={500} />
+      {errorObject != null ? RenderStackTrace(errorObject) : ''}
+    </Container>
```
O-Susanoo, line 38: Nice with this refactor

src/components/device/Device.jsx
```diff
@@ -47,109 +56,118 @@ class Device extends React.Component {
       match.params.deviceId,
     )
 
-    this.setState({ isLoading: false })
+    const activeDevice = DeviceUtils.getById(this.props.devices, this.props.match.params.deviceId)
+
+    this.setState({ isLoading: false, device: activeDevice })
```
O-Susanoo, line 61: my heart is broken when i'm dumped and when i see complex objects in state  i wish a more simpler state :)

test/components/utility/ErrorPage.test.jsx
```diff
@@ -0,0 +1,49 @@
...
+      <Router history={history}>
+        <ErrorPage
+          history={history}
+          errorCode={errorCode}
+        />
+      </Router>
+      ,
+    ))
+
+  describe('Test error pages ', () => {
```
O-Susanoo, line 25: i think another test that check that we don't change path will be good

src/components/device/Device.jsx
ConorTighe1995: I understand but this replaces 4-5 http requests and helped simplify issues with rending so I thought it was worth the trade off!

vdiezel: I'm afraid we run into some risky behaviour here: When we store the device in component state, but redux will create a deep copy of the device in the global application state when we call some update actions, we have an outdated activeDevice stored in this component state - or am I wrong?
I think we shouldn't break the pattern of deriving the device always from the global state.

O-Susanoo: i join viktor on this one

ConorTighe1995:
> Looks awesome @ConorTighe1995
> This might have gotten the award of most discussion ever in any OSPIN pr + slack. Thank you for your patience.

Haha no worries there was some good feedback and well spotted issues I missed, I'll gladly accept.

award

PR #680, opened 2020-04-14
a straight A - 2/87

vdiezel:
Bildschirmfoto von 2020-04-14 20-27-46

this targets the thunks (penalty of 1 week by codeclimate):
I created a wrapper for the `try-catch` block, made a seperate file for every thunk and organzied them somewhat in folders.

It is mainly copy pasting stuff around here - the only interesting part is the wrapper. Maybe you can just go through the commits and just double check the function signatures - thats where i could have an error most likely when extracting those function!

This gets us down from 1 week to 12 hours (new issues - codeclimate is really tough on the DRY aspect...): https://codeclimate.com/github/ospin-web-dev/REACTor/pull/680#issues

Next is the reducer.
CLOSES #NOTHING


src/redux/thunks/auth/callAuthenticateUser.js
```diff
@@ -0,0 +1,17 @@
+import * as nexus from 'nexus/index.js'
```
vdiezel, line 1: thats the only thunk that didn't fit into the abstraction pattern - will have a look on this later!

src/redux/thunks/thunkWrapper.js
```diff
@@ -0,0 +1,14 @@
+import { logAndExtractError } from 'utils/error'
+
+const responseWrapper = (fn, dispatch) => async (...args) => {
```
ConorTighe1995, line 3: noiceeee

O-Susanoo, line 3: i think this function could use 2 more optional function one in catch block and one after the catch , doing this could solve the auth thunks

PR #684, opened 2020-04-15
664 a straight A - part 3/87

vdiezel:
Reduce the reducer!
Does nothing but splitting up the reducer into domains and seperate files for every function.
 
But Viktor, why don't you use `combineReducer`? Tell us! 

Using `combineReducers` also changes the way you access the application state which means it requires a bigger refactoring. Also it requires a clear seperation of domains which we currently don't have.

very little code added - line number count due to abstracting code into new files (always feel guilty about that)

here are the new ratings: 
https://codeclimate.com/github/ospin-web-dev/REACTor/pull/684#ratings


src/redux/helper/objectMapper.js
```diff
@@ -11,15 +11,6 @@ export const createUserObject = userData => ({
   events: userData.events,
 })
 
-export const createUserLogObject = userLogData => ({
```
vdiezel, line 14: thats a ghost of a dead end point!

src/redux/reducers/command/endProcess.js
```diff
@@ -0,0 +1,25 @@
...
+import FlashMessenger from 'utils/FlashMessenger'
+import { ellipseString } from 'utils/string'
+
+export default (state, action) => {
+  const { processes, flashMessages } = state
+  const { processId } = action.payload
+
+  const process = Process.getById(processes, processId)
+
+  if (!process) return state
```
ConorTighe1995, line 12: If we dont get a process using id from process we return the previous state which contain the last processes? is that what is happening here?

src/redux/reducers/reducer.js
```diff
@@ -1,135 +1,37 @@
...
-
-  if (action.type in userHandlers) {
-    return userReducer(state, action)
-  } else if (action.type in processHandlers) {
-    return processReducer(state, action)
-  } else if (action.type in uIHandlers) {
-    return uIReducer(state, action)
-  } else if (action.type in deviceHandlers) {
-    return deviceReducer(state, action)
+  switch (true) {
```
ConorTighe1995, line 19: Forgive me if confusing this because its been a while since Ive had to use redux, but shouldn't this be in a combine reducer function and not a switch?

there should be a index reducer and then we use that to create the store:
https://redux.js.org/api/combinereducers

src/redux/helper/objectMapper.js
ConorTighe1995: rip in peace

src/redux/reducers/reducer.js
vdiezel: see PR description :)

ConorTighe1995: I totally missed that haha my bad when straight the code!

DanielSeehausen:
image

image

image

src/redux/helper/objectMapper.js
DanielSeehausen: 👻 🔫

DanielSeehausen:
(updated branch)

src/redux/reducers/command/endProcess.js
vdiezel: if we can't find the process we do nothing. This is for this use case: user A and user B look at the same device in the UI. User A creates a new process, starts it end stops it, too. User B's application wasn't aware of the process (it doesn't exists in its state) but now user Bs applcation does receive of the mqtt message from the device about this process. TO prevent the appication breaking here, we simply check if the app actually knows somehting about the process, and if so, do the logic required. Otherwise just don't do anything.

vdiezel:
> (updated branch)

getting the co-auth again...I see

PR #685, opened 2020-04-15
Fix Draggable focus

### LINKED ISSUE #657: inputs in process builder do not persist when clicking on a phase widget (and other interactive UI components?)
### Current Behavior
 In the process builder, when I enter a value into an input field and then click on a phase widget at the right side, the value I inserted is not persisted in the process, because the 'onBlur' event does not fire in this case(?)
This can to lead to customers starting a process with not updated values when they navigate quickly between the phases

---

Input should be persisted and saved into the process when a user clicks another UI element after insereting a value

### Images/Screenshots/Links
onblur_input

### LINKED ISSUE #682: onBlur for input fields in the process builder don't work when clicking within the same group
### Current Behavior
The `onBlur` event for the input fields in the process builder which submits the value does not trigger when clicked inside into a spot within the same spot (probably because it is in conflict with dragging). Because our Enter button submit doesn't work either (ticket created) , a user has to find a spot where they can acutally click to finally submit!

---

- clicking within a group after entering a value into an input field should also submit the input value!

---

### Steps to Reproduce
- go to any executable process, enter a value into an input field and try to click anywhere within the group -> the focus will remain in the input field (also the min-max popup will remain open) and the value will not be submitted until clicked somewhere outside

O-Susanoo:
This fix is very very very hacky but there is no way around it at least that's what i came up with

Before
before

After

after

closes #657 
closes #682


src/components/utility/DraggableComponent.jsx
```diff
@@ -1,20 +1,35 @@
...
+  const myRef = useRef()
+  return (
+    <Draggable draggableId={draggableId} index={index}>
+      { provided => (
+        <Ref innerRef={provided.innerRef}>
+          <As
+            {...provided.draggableProps}
+            {...provided.dragHandleProps}
+            onClick={() => {
+              myRef.current.focus()
```
ConorTighe1995, line 17: Id throw this in a named function and maybe even see if you can check for the negative here.

ConorTighe1995: Line 17 and 18 that is

ConorTighe1995:
Here is a github issue in case you have not seen it:
https://github.com/atlassian/react-beautiful-dnd/issues/410

Seem like this is a issue with the library itself and intentional by the devs. They do have a mouse down solution there it seems, dunno if that is a better solution that this one as it might introduce more head ache then its worth, give it a test if you want tho.

ConorTighe1995:
I tested it out and it didn't save when I clicked devices and navigated off the page, this could be a good opportunity to use something like componentWillUnmount or useEffect. Also wonder would it be worth firing a flash message to give the user visual feedback that the value is now saved?

O-Susanoo:
> Here is a github issue in case you have not seen it:
> [atlassian/react-beautiful-dnd#410](https://github.com/atlassian/react-beautiful-dnd/issues/410)
> 
> Seem like this is a issue with the library itself and intentional by the devs. They do have a mouse down solution there it seems, dunno if that is a better solution than this one as it might introduce more head ache then its worth, give it a test if you want tho.

I didn't see this one , but it seem i'm on the same page with , just my solution is way better :) because they don't solve the div focus issue i also looked at the mouseDown it's not working or at least i couldn't make it work

O-Susanoo:
> I tested it out and it didn't save when I clicked devices and navigated off the page, this could be a good opportunity to use something like componentWillUnmount or useEffect. Also wonder would it be worth firing a flash message to give the user visual feedback that the value is now saved?

Sorry i didn't understand what you meant here but when i clicked on devices it did save in fact this is working even without this fix ? yeah a flash message could be a good idea let's what the others think about it

O-Susanoo:
i failed to give viktor before a demo of the behavior with div , but since conor brought it back here is a demo of it

Peek 2020-04-16 11-57

src/components/utility/DraggableComponent.jsx
O-Susanoo: well i have two contradicting reason

- it's so simple that it doesn't t deserve a name

- it's so complicated no one dare speak its name

ConorTighe1995:
> > I tested it out and it didn't save when I clicked devices and navigated off the page, this could be a good opportunity to use something like componentWillUnmount or useEffect. Also wonder would it be worth firing a flash message to give the user visual feedback that the value is now saved?
> 
> Sorry i didn't understand what you meant here but when i clicked on devices it did save in fact this is working even without this fix ? yeah a flash message could be a good idea let's what the others think about it

It actually seems to work now so maybe it was my internet or something something to do with my browser

src/components/utility/DraggableComponent.jsx
ConorTighe1995: Simpler functions are better and easier to test plus they are cleaner, not encapsulating makes something simple more complex, the best functions are only 2-3 lines long and easy to read/reuse.

> it's so complicated no one dare speak its name
https://tenor.com/view/hm-hmm-hmmm-black-guy-gif-12687265

O-Susanoo: hahah nice meme , you are right , but the only thing that is left is if i do this i need to create a function inside the function itself and i don't like that much but i'm almost convinced i think i will do that

O-Susanoo: done

PR #677, opened 2020-04-14
a straight A - Part 1/87

vdiezel:
First part of the ticket #664 
also closes #673 

code climate is a bit unreliable when it comes to the issue count, it opens up issues for code that I didn't touch - but in total its an improvement: 
https://codeclimate.com/github/ospin-web-dev/REACTor/pull/677#ratings

PR #687, opened 2020-04-16
620 Require email on submission

### LINKED ISSUE #620: Add front end validation on required `email` when creating user
Currently, this form does not require email

@O-Susanoo says npm package: `yup`!

## Deliverables
- update the form to not allow submit when no email is present

image

ConorTighe1995:
This was a small bug that would allow you to submit a blank email to hambda. it was a quick fix using regex.

closes #620


vdiezel:
we will probably have a bigger conflict here once I merge #677 because I refactored SignUp big  time, too (it had a D as a rating). I also had to touch tests for this!

PR #689, opened 2020-04-16
Fix Zoom on Live Data

### LINKED ISSUE #686: Zooming in in live mode doesn't work as intended
### Current Behavior
 - as a user when I zoom in in the 'Live' time frame during a running process, the biggest part of the graph disappears

---

- I should zoom into the selected area

---

### Steps to Reproduce
- start a process, pick any graph (with the time frame 'Live' selected) and try to zoom in: 
zoom_error_live

---

O-Susanoo:
It seems that when the graph is moving zooming on on x,y position is not working , i couldn't found a reason , found someone talking about having few datapoints that's why it's not showing , i don't think so , because zooming sometimes work , but when it works is weird , so the basic idea of this pr is to zoom vertically not horizontally this way even if the graphs moves it will still work properly  

closes #686 

image


vdiezel:
Can you describe what the issue was?

src/components/processviewer/body/dataviewer/Stream.jsx
```diff
@@ -98,7 +97,7 @@ export default class Stream extends React.Component {
 
     if (triggersFetch) {
       const zoomParams = await GraphTools.fetchDataInViewport({
-        event, zoomMode, activeProcess, stream, completeStreamData: streamData,
+        event, zoomMode, activeProcess, stream, completeStreamData: streamData, dataIsLive: mode === 'live',
```
ConorTighe1995, line 100: Does this trigger the linter? Should this be multiple lines

src/utils/GraphTools.js
```diff
@@ -30,9 +31,13 @@ class GraphTools {
       YviewportMaximum,
       YviewportMinimum,
     }
-
     if (mode === 'auto') viewportParams.type = 'xy'
     else if (mode === 'fixed-y-axis') viewportParams.type = 'x'
```
ConorTighe1995, line 35: What would happen if mode was neither 'auto' or 'fixed-y-axis', would this throw an error? Can viewportParams.type be null and not lead to edge cases? Should there be a else case here to handle these scenarios? These are only things im wondering, could be fine.

src/utils/GraphTools.js
```diff
@@ -101,7 +107,7 @@ class GraphTools {
     const minValue = event.axisY[0].viewportMinimum
     const maxValue = event.axisY[0].viewportMaximum
```
ConorTighe1995, line 108: Destructure max and min

src/utils/GraphTools.js
```diff
@@ -101,7 +107,7 @@ class GraphTools {
...
 
-    const zoomParams = this.getZoomParams(zoomMode, maxTime, minTime, maxValue, minValue)
+    const zoomParams = this.getZoomParams(zoomMode, maxTime, minTime, maxValue, minValue, dataIsLive)
 
     if (datapointsInViewPort < this.CONSTANTS.MINIMUM_POINTS_IN_VIEWPORT) {
```
ConorTighe1995, line 112: check for the negative and return.

src/components/processviewer/body/dataviewer/Stream.jsx
DanielSeehausen: another vote for this. especially given the complexity with the renaming and logic

src/components/processviewer/body/dataviewer/Stream.jsx
```diff
@@ -82,9 +82,16 @@ export default class Stream extends React.Component {
     this.setState(prevState => ({ includeZero: !prevState.includeZero }))
   }
 
-  reloadDataOnRangeChangeHandler = async event => {
-    const { lastPan, zoomMode } = this.state
+  getZoomParamsForLiveData = (zoomMode, event) => {
+    const minTime = event.axisX[0].viewportMinimum
```
O-Susanoo, line 86: i named like this just to emphasis on the fact that this is meant only for live data

src/utils/GraphTools.js
O-Susanoo: i don't think it can be , the mode can't be anything else because it's something we provide

src/utils/GraphTools.js
```diff
@@ -30,9 +31,13 @@ class GraphTools {
...
       YviewportMinimum,
     }
-
     if (mode === 'auto') viewportParams.type = 'xy'
     else if (mode === 'fixed-y-axis') viewportParams.type = 'x'
+
+    if (dataIsLive) {
+      viewportParams.XviewportMaximum = null
+      viewportParams.XviewportMinimum = null
+    }
```
DanielSeehausen: in the current function, there is somewhere else this code might better belong (where these assignments are already being done)

In reading the function and coming along this change at the end, there is an implication that this is here at the end for a reason. I don't see anything happening between this and the initial creation of `viewportParams` though

src/utils/GraphTools.js
DanielSeehausen: I appreciate the vote of confidence in the OSPIN web dev developers, and think it's well deserved. Saying that, I believe @ConorTighe1995 point is still valid - defensive programming is to protect against mistakes generated by us more than anything.

src/components/processviewer/body/dataviewer/Stream.jsx
```diff
@@ -96,6 +101,11 @@ export default class Stream extends React.Component {
       this.setState({ lastPan: updatedPan })
     }
 
+    if (mode === 'live') {
+      this.setState({ zoomParams: this.getZoomParamsForLiveData(zoomMode, event) })
+      return
+    }
+
```
DanielSeehausen, line 108: can this happen before the ` (event.trigger === 'pan') `?
If so, I believe it should come before line 97, as its put off until later though it relies on none of the above. 

If it does need the `(event.trigger === 'pan') ` to happen first, then makes sense why its here

DanielSeehausen: If possible, please provide a gif of the fix!

O-Susanoo: what happen is that you want to be able to pan even on live data if this one goes before the pan then it will stop any panning from happening that's why it's after i actually tested both cases that's why i know

PR #692, opened 2020-04-20
Fix connectionState

### LINKED ISSUE #619: the connectionState of devices does not behave correctly anymore
Example 1: 

When loading the devices view the frist time and a device IS online, it is displayed correctly first (red lid, colored state circles below) but in certain cases, when you go back to the devices view, all devices suddenly appear offline (trickier to reproduce)

Exmaple 2: 

However this one ca be reproduced: 
Choose a device that is offline and then fire a heartbeat from th IoT Console on the topic `{stage}/ping/{deviceId}`: The device will appear online, but it should go offline after a coupe of seconds because no more pings (heartbeats) are coming in BUT it stays in its online display mode until you refresh or go INTO the device. 

I think this bugs are related.

O-Susanoo:
Guys i think you will really love reviewing this one , the thing is because redux is such a nice tool it always become a pain if we forget to go by there pattern of correctly updating a nested property 
closes #619


O-Susanoo:
it will up this after none just to make sure no weird side effect

src/redux/reducers/device/checkDeviceHeartbeats.js
```diff
@@ -7,9 +7,10 @@ export default state => {
   const devices = state.devices.map(device => {
     const lastDeviceHeartbeat = state.heartbeats.devices[device.id]
     const timeSinceLastHeartbeat = now - lastDeviceHeartbeat
-    if (timeSinceLastHeartbeat > NETWORK_CONSTANTS.MAX_MS_TIME_BETWEEN_HEARTBEATS) {
-      device.connectionState = "offline"
+    const deviceWentDown = timeSinceLastHeartbeat > NETWORK_CONSTANTS.MAX_MS_TIME_BETWEEN_HEARTBEATS && device.connectionState !== 'offline'
```
ConorTighe1995, line 10: This is a really long line, maybe something like 
```
const isDeviceDown = () => {
  const limit = timeSinceLastHeartbeat > NETWORK_CONSTANTS.MAX_MS_TIME_BETWEEN_HEARTBEATS
  return limit && device.connectionState !== 'offline'
}
```

ConorTighe1995:
Also had a look for tests but couldn't see any, I could be missing them but might be a good place to add tests to if there are none.

src/redux/reducers/device/checkDeviceHeartbeats.js
DanielSeehausen, line 10: the solution looks great

agree with conor's sentiment re: the compound operations being a little much to read. (Think the example might need some slight changing). 
however you take it, my slow 🧠 would certainly appreciate the compound being broken up. Is in keeping with Sandi Metz rule #2, which I think is actually big brain genius:
https://thoughtbot.com/blog/sandi-metz-rules-for-developers
(granted that was for ruby, I think its fair to give another line for JS...maybe)

PR #699, opened 2020-04-22
478 fix datapoints and timestamps for phase

### LINKED ISSUE #478: When a (first?) phase has only one datapoint, xlabels for timestamps are wrong
Bildschirmfoto von 2020-01-19 13-00-52

ConorTighe1995:
This is a simple solution to remove the 0 offset in a edge case scenario where someone clicks start process then immediately clicks stop, causing strange behavior on the graph. I have also addd some time util test in this one too.

closes #478


src/utils/Time.js
```diff
@@ -25,12 +25,11 @@ export default class Time {
 
   static timeAxisLabelFormat(durationInSeconds, derivedTimeDisplayFormat) {
     const ms = durationInSeconds * 1000
-
     const days = Math.floor(durationInSeconds / secondsPerDay)
     const displayFormat = (days > 0)
       ? `[${days}d ]${derivedTimeDisplayFormat}`
       : derivedTimeDisplayFormat
-
+    if (ms < 0) return moment.utc(0).format(displayFormat)
```
O-Susanoo, line 32: this is the fix for anyone reviewing this pr

PR #708, opened 2020-04-24
552 refactor access device modal

### LINKED ISSUE #552: Device Sharing Modal Rework
- [ ] rework modal behavior
- [ ] fix text

#### Current Behavior:
  -  Attempt to share a device with a user via the `Add User` button -> see modal
  - Enter valid information and click `Submit` -> User is presented with a helpful green 'Success' message (see image)
  - Click `Submit` again -> Receive a misleading 'Something went wrong' message
  - Exit the modal by clicking off screen or on `Close`
  - Click on `Add User` again -> Get the modal with the same messages still populated within it

#### Potential Reworked Behavior
There are three things worth reworking in the above flow:
1. When a user clicks submit, there isn't any other action for them to take with the modal. A successful submission should close the modal. In order to still provide the positive affirmation, we could display a green FlashNotification (one possible solution). Alternatively, we could display the green message and disable the option to submit again. 

On error, we could keep behavior as is, and display the message from the API response in the red box (it may be more helpful than what we are displaying now). 

2. if ospin-web-dev/hambda#261 is resolved, we could display a more meaningful error message when the user has already been invited. See ticket for details. 

3. When the user exits a modal that has messages populated in it, then opens up the modal with `Add User`, they shouldn't see the old messages there (see screenshot). The modal's state should be renewed when it is opened.

Successfully submitting for the first time:
image

Submitting a second time in a row (also, if the modal is closed and reopened, this is what the user would see again):
image

----

## Fix Text
Spitballing some ideas here 

(this one is just adds 'the')
"User daniel.seehausen@ospin.de was invited to the device"
or
"User daniel.seehausen@ospin.de was invited to device: ${device.name}"
or
"An invitation to the ${device.name} has been sent to daniel.seehausen@ospin.de"
or 
whatever you like

image

ConorTighe1995:
Credits to @vdiezel  he he solved a lot of this ticket while refactoring. I added a flash message and did some refactoring to AccessDeviceModal.jsx.

Ive removed the portal as it was causing issues with tests and was not worth the hassle of fixing.

closes #552


src/components/device/modals/AccessDeviceModal.jsx
```diff
@@ -1,67 +1,70 @@
...
-const mapStateToProps = state => ({
-  user: state.user
-})
+const mapStateToProps = state => ({ user: state.user })
 
-const mapDispatchToProps = dispatch => ({
-  callGrantDeviceAccess: (deviceId, email, groupName, hostName) =>
-  dispatch(callGrantDeviceAccess(deviceId, email, groupName, hostName))
-})
+const mapDispatchToProps = dispatch => ({ callGrantDeviceAccess: (deviceId, email, groupName, hostName) => dispatch(callGrantDeviceAccess(deviceId, email, groupName, hostName)) })
```
O-Susanoo, line 10: since you were doing linting it would have been great if you changed this  callGrantDeviceAccess to dispatchCallGrantDeviceAccess

src/components/device/modals/AccessDeviceModal.jsx
```diff
@@ -70,62 +73,56 @@ const AccessDeviceModal = props => {
...
+        placeholder='Choose group'
+      />
+      <Message
+        error={saveMsg.isError}
+        success={!saveMsg.isError}
+        hidden={saveMsg.hidden}
+        icon={!saveMsg.isError ? 'check circle' : 'warning circle'}
+        header={messageHeader}
+        content={saveMsg.msg}
+      />
```
vdiezel, line 107: didn't know you could squash them like that.

vdiezel: but desctructuring would be nice in any case.

test/components/utility/FlashMessagesContainer.test.js
```diff
@@ -9,7 +9,7 @@ import * as actions from 'redux/actions/actions'
 
 const mockStore = configureMockStore()
 
-describe('FlashMessagesContainer.tsx', () => {
+describe.only('FlashMessagesContainer.tsx', () => {
```
vdiezel, line 12: That guy needs to go :)

src/components/utility/modal/formalize.jsx
```diff
@@ -11,10 +11,7 @@ function formalize(Inputs) {
   return props => {
     const {
       errorMessage,
-      errorMessageHeader = 'Something went wrong',
       isError,
-      successMessage,
-      successMessageHeader = 'Success',
```
vdiezel, line 17: This breaks the interface of the component because now we can't pass a `successMessage` or a `successMessageHeader` down.

src/components/utility/modal/formalize.jsx
```diff
@@ -27,16 +24,11 @@ function formalize(Inputs) {
       <Form error={isError} success={isSuccess} style={{ marginBottom: '15px', paddingBottom: '15px' }}>
         <Inputs {...props} />
         <Message
-          error
-          header={errorMessageHeader}
+          error={isError}
+          success={!isError}
+          icon={!isError ? 'check circle' : 'warning circle'}
+          header={!isError ? 'Success!' : 'Something went wrong.'}
           content={errorMessage}
```
vdiezel, line 31: The `content`  property  should take the correct text (either errorMessage or successMessage) depending on what is true.  Same applies for the header.

src/components/utility/modal/formalize.jsx
```diff
@@ -27,16 +24,11 @@ function formalize(Inputs) {
       <Form error={isError} success={isSuccess} style={{ marginBottom: '15px', paddingBottom: '15px' }}>
         <Inputs {...props} />
         <Message
-          error
-          header={errorMessageHeader}
+          error={isError}
+          success={!isError}
+          icon={!isError ? 'check circle' : 'warning circle'}
+          header={!isError ? 'Success!' : 'Something went wrong.'}
```
vdiezel, line 30: The initial state of the modal will have a `isError = false`, so a success message will be displayed as I don't see any hidden property here whic you will need to use if you want to reuse the same message component for `isError` or `isSuccess`

ConorTighe1995:
> The refactoring is already nice, but your changes to the `formalize` component do break the interface here, and because this HOC is already used at several places we have to come to an agreement here.
> 
> 1. The formalize HOC should allow setting a specific success message and success messega header. This allows us to communicate clearly what happened to the user.
> 2. If you do want to change formalize, then you have to check all the other places where it is currently used and validate that nothing breaks. IIt will not break codewise, but I will not display the messages as intended in some cases
> 
> I will try convince you to revert the changes in formalize:
> 
> The only difference in your approach and the current implementation is the handling of responses for message display, so successful and and not succesfull request. While you use a `hidden` and a `isError` flag, the current code uses `isSuccess` and `isError`. There are only tiny differences, but my opinion is this:
> 
> Your implementation implies that when `isError` is false, then the request was succesful, and the only reason a user doesn't see a success message is because you hide it. I think this might be confusing, because the user might not have actually fired the request yet. This also has implicit knowledge required: A request is considered succesful when `isError` and `hidden` are false, while the current implementation is explicit.
> 
> But you can convince me otherwise :) - in this case you will need to guarantee though that you refactor it also for all the other modals. If you agree though, this is my proposal:
> Keep formalize as it is and adapt the AccessDeviceModal to fit the formalize API.

I have refactored it to use the original form to simplify things, let me know what you think!

src/components/device/modals/AccessDeviceModal.jsx
```diff
@@ -1,67 +1,70 @@
...
+    const res = await props.dispatchGrantDeviceAccess(props.activeDevice.id, email, assignedGroup, props.user.userName)
 
-    if (res.success){
-      setValues({ ...modalState, isSuccess: true, isLoading: false })
-    }
-    else {
-      setValues({ ...modalState, isError: true, errorMessage: res.message, isLoading: false })
+    if (res.success) {
+      setSaveMsg({ hidden: true, isError: false, msg: '' })
+      FlashMessenger.success(`User ${email} was invited to device.`)
```
DanielSeehausen, line 61: requesting device name here

DanielSeehausen: (or 'the device`, but I think device name is probably best)

DanielSeehausen:
looks good pending resolution of what you are @vdiezel are talking about re: the HOC. only request from my end is to update the confirmation text to mention the device by name (if you don'y have the device on hand, you can use the Device interface to derive the display name from the deviceId, which should be in the component.

ConorTighe1995:
More updates have been added to further refactor and catch some errors

test/components/processviewer/body/dataviewer/DataGraphs.test.jsx
```diff
@@ -139,7 +139,7 @@ describe('DataGraphs.jsx', () => {
         .filter(stream => stream.groupName === groupName)
         .map(stream => stream.name)
 
-      const { queryByText } = renderDataGraphs(mockedDevice, streams, runningProcess)
+      const { container, queryByText } = renderDataGraphs(mockedDevice, streams, runningProcess)
```
O-Susanoo, line 142: i don't think container is needed here

src/components/device/modals/AccessDeviceModal.jsx
```diff
@@ -70,62 +83,20 @@ const AccessDeviceModal = props => {
...
-        </Modal.Description>
-      </Modal.Content>
-    </Modal>
+    <AccessDeviceModalForm
+      isError={saveMsg.isError}
+      errorMessage={saveMsg.msg}
+      successMessage={saveMsg.msg}
+      closeHandler={onToggleClose}
+      submitHandler={onSubmit}
+      submitText='Adding User...'
```
vdiezel, line 93: The submit text is the one displayed on the Button; maybe make it 'Add User' (communicating an action rather than progress 'Adding User...'

src/components/device/modals/AccessDeviceModal.jsx
```diff
@@ -1,67 +1,80 @@
...
+const mapStateToProps = state => ({ user: state.user })
 
-const mapDispatchToProps = dispatch => ({
-  callGrantDeviceAccess: (deviceId, email, groupName, hostName) =>
-  dispatch(callGrantDeviceAccess(deviceId, email, groupName, hostName))
-})
+const mapDispatchToProps = dispatch => ({ dispatchGrantDeviceAccess: (deviceId, email, groupName, hostName) => dispatch(callGrantDeviceAccess(deviceId, email, groupName, hostName)) })
 
 const AccessDeviceModal = props => {
+  const [saveMsg, setSaveMsg] = useState({ hidden: true, isError: false, msg: '' })
```
vdiezel, line 13: `hidden` o not used anymore. `isError` false or true toggle the display in the `formalized` component

PR #709, opened 2020-04-26
Adds collapsing and hiding the Sidebar.

### LINKED ISSUE #442: Collapse Ospin Hamburger Menu when the viewport is below a certain width
Please first make a new ticket for a 1 or 2 point spike into this to see if it is viable to do and not some crazy amount of effort. Following, define this ticket further for a plan of action OR close it without any points on it as labels. 

We need some more real-estate [in some cases](https://github.com/ospin-web-dev/REACTor/issues/441) (and will increasingly), and now is a good time to add an easy win for it:

when the viewport is below a certain width (experiment to find the appropriate minimum), the hamburger menu should minimize to just show icons:

current:
image

should collapse to just display the width required to comfortable show the icons. something like: 
image

The Ospin logo should resize and be at the top still!

See here: https://app.datadoghq.com

vdiezel:
closes #442 
Mimics somewhat the behaviour of the datadog sidebar. 
Sidebar can now be either manually collapsed: 

collapse_1

or hidden (works also in collapsed mode):

hide_1
hide_2

also automatically collapses it below 1200 px width. Can't be expanded below this viewport width, but hidden. 

rescale_1

When you hover the blue button when the sidebar is hidden, the sidebar shows up, but disappears if you don't pin it when leaving  area with the mouse (gif makes it looks very jerky): 

hover_1

---
Update: I made the collapsed sidebar slightly slimmer (72px -> 56px), didn't want to reshoot the gifs. 

---
Update II: Removed hiding the sidebar; made the collpase icon a button with background color on hover which is alway visible now (onHover is bad for tablet/mobile devices):

collpase_v2


DanielSeehausen:
This looks awesome! I have some UI thoughts, which I hope don't detract from how great this is looking. 

Would you be able to center the icons and the logo? They seem to be in different positions depending on when the sidebar is forced to collapse because of viewport width, vs. when the user elects to collapse it.

image

The grey `<` and `x` are also right on the border with the top of the viewport. they might benefit from being moved a couple pixels down. They are also really close together and really small. Datadog solved this by making them slightly larger and doing the background box.

On an unrelated note, while I like the pin/hide feature a lot, I perceive (from a user's perspective), that i might be adding UI complexity for minimal functionality gain. I can't imagine users will ever really make use of it. With it, they have an extra two buttons, the blue icon cramped in a corner hovering over existing functionality (the search bar), and diverging behavior re: whether the sidebar overlays or not. I see they implemented that on datadog, and I have the same impression using it there (extra noise, never used). Take a look at their interface and imagine there was only a `-` or a `+` sign provided instead.

Do we want to consider tooltips when collapsed? I don't have a strong preference either way, but users might need reinforcing which icon means what. Datadog has those sub menus which solve that issue, which we don't have or yet really need. (YET!)

Do we want this to be a part of UIConfig? I love the collapsing in general, and if someone opts to collapse then maybe we should reward their moxie and always give them a collapsed menu?

DanielSeehausen:
unrelated, check this gamers stuff out:
https://jonsuh.com/hamburgers/

DanielSeehausen:
since expanding/collapsing pushes the whole viewport to the right/left, it might be trivial to add a css animation on to the class to give it a smoother transition

ConorTighe1995:
The pin and X feature is a good idea but I think you should replace the arrow and X with a blue drawer icon. In the future if we start trying to move to a more tablet/mobile fixed floating buttons might make life a little harder.

Also forgive me if I am missing something, but in the 3rd give when you re-pin the menu devices fall behind it?

src/components/main/OspinSidebar.jsx
```diff
@@ -73,44 +97,143 @@ class OspinSidebar extends React.Component {
...
+      <>
+        {this.renderShowSidebarIcon()}
+        <Container>
+          <Sidebar
+            as={Menu}
+            visible
+            vertical
+            className={this.getSidebarStyleClasses()}
+            onMouseEnter={() => this.toggleCollapseIconDisplay(true)}
+            onMouseLeave={() => this.clearHoverFlags()}
```
O-Susanoo, line 221: no need for inline function here

vdiezel:
Thanks for the review! I'll remove the hiding feature. 

> since expanding/collapsing pushes the whole viewport to the right/left, it might be trivial to add a css animation on to the class to give it a smoother transition (maybe that is already present and the gif framerate is too slow)

I tried but the icon and the text in sidebar menu items are collpasing into two rows when having a transition which looks ugly

vdiezel:
>Do we want to consider tooltips when collapsed? I don't have a strong preference either way, but users might need reinforcing which icon means what. Datadog has those sub menus which solve that issue, which we don't have or yet really need. (YET!)

I tried that to and the semamntic ui popup doesn't work on the menu items in a sidebar due to a name conflict of classes (popup positioning doesn't work properly, as the class names 'left' and 'right' also used on the sidebar, caused unexpected beahviour...

vdiezel:
> The pin and X feature is a good idea but I think you should replace the arrow and X with a blue drawer icon. In the future if we start trying to move to a more tablet/mobile friendly design fixed floating buttons might make life a little harder. Keeping the drawer always on screen and then only hiding it on small tablets and phones will take away a lot of hassle.
> 
> Also forgive me if I am missing something, but in the 3rd give when you re-pin the menu devices fall behind it?

things fall only behind when the sidebar is not pinned and appears only on hovering. Once pinned, things are being pushed to the left.

ConorTighe1995:
> > The pin and X feature is a good idea but I think you should replace the arrow and X with a blue drawer icon. In the future if we start trying to move to a more tablet/mobile friendly design fixed floating buttons might make life a little harder. Keeping the drawer always on screen and then only hiding it on small tablets and phones will take away a lot of hassle.
> > Also forgive me if I am missing something, but in the 3rd give when you re-pin the menu devices fall behind it?
> 
> things fall only behind when the sidebar is not pinned and appears only on hovering. Once pinned, things are being pushed to the left.

Cool but I think that just adds to the point of it making adaptive design a bit more difficult, if we were to keep it we should add a animation to hiding it when you move the mouse away which can be done with a HOC and CSS but is a bit more work.

PR #711, opened 2020-04-27
597 validation rework

### LINKED ISSUE #597: Refactor our validation utility and validation of process missmatches
Our validation utility is in terrible shape. 

It's main task is to validate device state and process description to find any invalid data, like undefined stream targets, targets out of allowed limits etc. The code works, but it is terribly written by me at the beginning of  my career. Refactor it Before attacking this main task:

It also contains a validation to check if the process description is in sync with the current device streams, which means it contains exactly the same streams in it `process.deviceDescription` as currently defined on `device.streams`). When it finds a missmatch, it pops up a `DescriptionMismatchModal` to tell the customer about the mismatch and proposes further action. This is a bit annyoing, it would be better to show this modal on trying to start the process.

The validation should be written in a way that makes it easy to transport this code also to the backedn, as we need to implement the validation there too!

vdiezel:
closes #597 

splits up the validation utility file into several new Interfaces for process, device and input. 
Commit history is a bit sloppy, sry.

Regarding this part of the ticket: 
> When it finds a missmatch, it pops up a DescriptionMismatchModal to tell the customer about the mismatch and proposes further action. This is a bit annyoing, it would be better to show this modal on trying to start the process.

It is actually the intended behaviour to show it when the process is opened because if we would show it on process start, a user might spent a lot of time configuring a process jsut to find out the process description does not match the device anyway once they want to start it...so I didn't do changes here.


src/components/processviewer/headers/ProcessBuilderHeader.jsx
```diff
@@ -16,17 +17,16 @@ const mapDispatchToProps = dispatch => ({
 
 class ProcessBuilderHeader extends React.Component {
 
-  state = {
-    showExecuteModal: false,
-    executeProcessImage: 'assets/playButtonBlue.png',
```
vdiezel, line 21: not used!

src/utils/validation/ProcessValidator.js
```diff
@@ -0,0 +1,135 @@
+import Streams from 'utils/Streams'
+import DeviceConfig from 'utils/DeviceConfig'
+import { replaceNoUnitCharacter, convertManyFrom } from 'utils/units'
+import DescriptionParser from 'utils/process/DescriptionParser'
+
+export default class ProcessValidator {
+
+  static validateDescriptionMismatch(process, device){
+    const missMatches = ProcessValidator.getMismatchingStreams(process, device),
+      validationErrors = []
```
DanielSeehausen: this took me a moment to recognize what I was looking at.
might follow convention if you line up the start of `validationErrors` with `missMatches`

src/utils/validation/ProcessValidator.js
```diff
@@ -0,0 +1,135 @@
...
+
+    for (const stream of missMatches.missingInDevice){
+      validationErrors.push({
+        type: "MissingDeviceStreamError",
+        message: `Your process description includes a not connected device stream:
+                  ${stream.name}`
+      })
+    }
```
DanielSeehausen: I believe you can initialize validation errors here with a map and avoid the sandwich code

src/utils/validation/ProcessValidator.js
```diff
@@ -0,0 +1,135 @@
...
+
+    for (const stream of missMatches.missingInProcess){
+      validationErrors.push({
+        type: "MissingProcessStreamError",
+        message: `Your process description is missing a connected device stream:
+                  ${DeviceConfig.getDisplayedStreamName(device, stream.id)}`
+      })
+    }
```
DanielSeehausen: consider making both of these functions of their own

src/utils/validation/ProcessValidator.js
```diff
@@ -0,0 +1,135 @@
...
+      if (!deviceStreamIds.includes(stream.id)){
+        missingStreamsInDevice.push({...stream})
+      }
+    }
+
+    for (const stream of deviceStreams){
+      if (!processStreamIds.includes(stream.id)){
+        missingStreamsInProcess.push({...stream})
+        }
+      }
```
DanielSeehausen: these are the same method, and can be abstracted

DanielSeehausen: some funky formatting here - check that linter is running

src/utils/validation/ProcessValidator.js
```diff
@@ -0,0 +1,135 @@
...
+    const descriptionMismatchErrors = ProcessValidator.validateDescriptionMismatch(process, device)
+
+    return [...finishingMethodErrors, ...streamTargetErrors, ...descriptionMismatchErrors]
+  }
+
+  static validateFinishingMethod(process) {
+    const validationErrors = [];
+
+    for (const key of Object.keys(process.description)) {
+      const phase = process.description[key]
```
DanielSeehausen: this is a sign we probably need a helper method (or need to use an existing one) to get a processes phases.

src/utils/validation/ProcessValidator.js
```diff
@@ -0,0 +1,135 @@
...
+            message: `Value in group ${stream.groupName} at stream ${stream.name} is out of limits.
+              Provided ${displayedTarget} ${replaceNoUnitCharacter(displayedUnit, '')},
+              limits are [${displayedMin} ${replaceNoUnitCharacter(displayedUnit, '')},
+              ${displayedMax} ${replaceNoUnitCharacter(displayedUnit, '')}]`,
+          })
+        }
+      })
+    })
+    return validationErrors
+  }
```
DanielSeehausen: consider breaking this method up

test/utils/validation/ProcessValidator.test.js
```diff
@@ -0,0 +1,123 @@
...
+
+describe('ProcessValidator', () => {
+  describe('validateProcessDescription', () => {
+
+    const phaseName = 'Corona Phase'
+
+    const createSetup = (deviceStreams, processStreams) => {
+      const device = TDG.generateDevice({ streams: deviceStreams })
+
+      const description = { ...PG.generateProcessPhase(0,
```
DanielSeehausen, line 14: consider newlining the ... for readability. there may be too many things being declared in line here. consider declaring the object first.

src/utils/Streams.js
```diff
@@ -4,6 +4,10 @@ export default class Streams {
     return streams.find(stream => stream.id === streamId)
   }
 
+  static has(streams, streamId) {
```
DanielSeehausen, line 7: while they say the same thing, I very commonly see `contains` to describe this method. Not requesting it to change, just a general convention thing.

on a separate note, you can delegate to `getById` with a `!!`. While we get nothing from it in this case, as things are, the benefit is this centralizes the ways in which we can look up streams in a group of streams.

src/utils/validation/ProcessValidator.js
```diff
@@ -5,56 +5,54 @@ import DescriptionParser from 'utils/process/DescriptionParser'
...
-      })
-    }
-
-    for (const stream of missMatches.missingInProcess){
-      validationErrors.push({
-        type: "MissingProcessStreamError",
-        message: `Your process description is missing a connected device stream:
-                  ${DeviceConfig.getDisplayedStreamName(device, stream.id)}`
-      })
+  static VALIDATION_ERRORS = new Set([
```
DanielSeehausen, line 8: nice set

src/utils/validation/ProcessValidator.js
```diff
@@ -5,56 +5,54 @@ import DescriptionParser from 'utils/process/DescriptionParser'
...
+    "MissingDeviceStreamError",
+    "MissingProcessStreamError",
+    "TransitionError",
+    "StreamTargetError",
+  ])
+
+  static createError(type, message, phaseName = undefined) {
+    if (!ProcessValidator.VALIDATION_ERRORS.has(type)) {
+      console.warn(`Unknows validation error type: ${type}`)
```
DanielSeehausen, line 17: `Unknows` -> `Unknown`

src/utils/validation/ProcessValidator.js
```diff
@@ -5,56 +5,54 @@ import DescriptionParser from 'utils/process/DescriptionParser'
...
+      )
+    ))
+
+    const missingProcessStreamErrors = missingInProcess.map(stream => (
+      ProcessValidator.createError(
+        "MissingProcessStreamError",
+        `Your process description is missing a connected device stream:
+          ${DeviceConfig.getDisplayedStreamName(device, stream.id)}`,
+      )
+    ))
```
DanielSeehausen: consider breaking these out into their own methods.

test/utils/validation/InputValidator.test.js
```diff
@@ -0,0 +1,36 @@
...
+        expect(res).toBe(true)
+      })
+
+      test('returns false for a invalid comment', () => {
+        const invalidComment = 'x'.repeat(commentLength + 1)
+        const res = InputValidator.isValidProcessComment(invalidComment)
+        expect(res).toBe(false)
+      })
+    })
+  })
```
DanielSeehausen, line 35: nice!

src/utils/validation/InputValidator.js
```diff
@@ -0,0 +1,15 @@
...
+  }
+
+  static isValidProcessName = string => (
+    string.length <= InputValidator.process.nameLength
+  )
+
+  static isValidProcessComment = string => (
+    string.length <= InputValidator.process.commentLength
+  )
+}
```
DanielSeehausen, line 15: A few things come to mind re: where this is placed/how it is named. I believe this is breaching a boundary between the view and the modal, in a general sense. 

Everything in `InputValidator` is specific to a process (as it is now). Would this be better suited on the `Process` interface/domain? Consider if the process interface that already exists ever needed to reference how long a process name can be. It might not make sense that it would derive that from something that is for validating user input in the view. 

Even if this just became `utils/validation/ProcessValidator.js` that might be more intuitive.

src/utils/validation/ProcessValidator.js
```diff
@@ -0,0 +1,156 @@
...
+
+    const missingProcessStreamErrors = missingInProcess.map(stream => (
+      ProcessValidator.createError(
+        'MissingProcessStreamError',
+        `Your process description is missing a connected device stream:
+          ${DeviceConfig.getDisplayedStreamName(device, stream.id)}`,
+      )
+    ))
```
DanielSeehausen, line 39: consider abstracting to their own methods

src/utils/validation/ProcessValidator.js
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  static getMismatchingStreams(process, device) {
+    const { deviceDescription } = process
+    const { streams } = device
+
+    const missingInDevice = deviceDescription
+      .filter(processStream => !Streams.has(streams, processStream.id))
+    const missingInProcess = streams
+      .filter(deviceStream => !Streams.has(deviceDescription, deviceStream.id))
```
DanielSeehausen, line 51: these are the same method - consider abstracting and naming

src/utils/validation/ProcessValidator.js
```diff
@@ -0,0 +1,156 @@
...
+    Number.isNaN(target) || typeof target !== 'number'
+  )
+
+  static targetOutOfStreamLimits = (target, stream) => (
+    target > stream.max || target < stream.min
+  )
+
+  static isNotBoolean = target => (
+    typeof target !== 'boolean'
+  )
```
DanielSeehausen, line 87: this might not belong on the `ProcessValidator`

src/utils/validation/ProcessValidator.js
```diff
@@ -0,0 +1,156 @@
...
+        'StreamTargetError',
+        `Invalid target value ${target} in group ${stream.groupName} at stream ${stream.name}. Expected true or false.`,
+        phase.name,
+      )
+    }
+  }
+
+  static validateOneOf() {
+    // TODO
+  }
```
DanielSeehausen, line 126: should this be deleted?

DanielSeehausen: nevermind I see this is for what Mo is working on - nice

src/utils/validation/ProcessValidator.js
```diff
@@ -0,0 +1,156 @@
...
+
+      streamIds.forEach(streamId => {
+        const stream = Streams.getById(process.deviceDescription, streamId)
+        const target = DescriptionParser.getStreamTarget(process, id, stream.id)
+
+        const error = ProcessValidator.validateByStreamType(device, process, phase, target, stream)
+        if (error) validationErrors.push(error)
+      })
+    })
+    return validationErrors
```
DanielSeehausen, line 154: consider abstracting into named methods
to avoid the sandwich code, this might be a 
```js
reduce x {
  const phaseErrors = x.filter(...)
  return [ ...acc, ...phaseErrors ] (or concat)
}
```
of if you don't like reduce, if they are named methods, you can then `.join` the result off of a `map`

test/helpers/CommonGenerator.js
```diff
@@ -17,17 +17,22 @@ class CommonGenerator {
...
-    default: this.generateNumber(),
-    max: this.generateNumber(),
-    min: this.generateNumber(),
-    unit: '°C',
-    name: faker.random.word(),
-    groupName: faker.random.word(),
-    ...data,
-  })
+  static generateDownStream = data => {
+    const defaultValue = this.generateNumber()
```
DanielSeehausen, line 21: everywhere else `this.generateNumber` was used inline in the object creation. Any reason this one is different?

test/utils/validation/DeviceValidator.test.js
```diff
@@ -0,0 +1,29 @@
...
+
+    test('returns an error when the device is offline', () => {
+      const device = TDG.generateDevice({}, { connectionState: 'online', state: 'running' })
+      const res = DeviceValidator.validateState(device)
+      expect(res[0].message).toBe('Device is running another process')
+      expect(res[0].type).toBe('DeviceStateError')
+      expect(res.length).toBe(1)
+    })
+  })
+})
```
DanielSeehausen, line 29: nice!

src/utils/validation/ProcessValidator.js
DanielSeehausen: (this might be outdated)

test/helpers/CommonGenerator.js
vdiezel: I need to create it in advance now so I can make `min` and `max` value actually be smaller resp. bigger than the default value. previously it has been totally random, resulting in flickering tests when I validated stream targets out of limits

src/utils/validation/ProcessValidator.js
ConorTighe1995, line 17: Should this be a flash message? Outside of developing there is not a huge benefit to keeping warnings in the code so maybe there is a better way to handle this then pushing a console.warn to master.

src/utils/Streams.js
vdiezel: makes sense!

vdiezel: fixed!

src/utils/validation/ProcessValidator.js
```diff
@@ -82,46 +82,53 @@ export default class ProcessValidator {
...
+          Provided ${displayedTarget} ${replaceNoUnitCharacter(displayedUnit, '')},
+          limits are [${displayedMin} ${replaceNoUnitCharacter(displayedUnit, '')},
+          ${displayedMax} ${replaceNoUnitCharacter(displayedUnit, '')}]`,
+        phase.name,
+      )
+    }
+  }
 
-    const phaseIds = Object.keys(process.description)
+  static validateByStreamType(device, process, phase, target, stream) {
```
ConorTighe1995, line 128: A lot of of params, should be a single object then deconstruct the type.

The switch that looks at types, would it make sense to have a ENUM of expected types?

ConorTighe1995: Changing it to a object could involve a big refactor so If that is the case dont worry about it too much.

src/utils/validation/ProcessValidator.js
vdiezel: done!

src/components/processviewer/ProcessViewer.jsx
```diff
@@ -106,25 +106,31 @@ class ProcessViewer extends React.Component {
...
     this.setState({ amendingProcess: true })
 
-    const missMatchStreams = getMismatchingStreams(this.props.activeProcess, this.props.activeDevice)
+    const missMatchStreams = ProcessValidator.getMismatchingStreams(activeProcess, activeDevice)
 
-    const amendedProcess = ProcessRectifier.amendDescriptionMismatch(this.props.activeProcess, this.props.activeDevice, missMatchStreams)
+    const amendedProcess = ProcessRectifier
+      .amendDescriptionMismatch(activeProcess, activeDevice, missMatchStreams)
 
     const params = {
```
ConorTighe1995, line 117: Better name would be helpful

test/utils/validation/ProcessValidator.test.js
```diff
@@ -5,7 +5,7 @@ import PG from '../../helpers/ProcessGenerator'
 
 
 describe('ProcessValidator', () => {
-  describe('validateProcessDescription', () => {
+  describe('validateDescription', () => {
     const phaseName = 'Corona Phase'
```
ConorTighe1995, line 9: rock

src/utils/validation/ProcessValidator.js
vdiezel: done!

src/utils/validation/ProcessValidator.js
vdiezel: a costumer is not aware of stream types, so they would wonder about the message. But you are right on poiting this out because if the type is unknown than we or firmware screwed up and with just a warning this could go by silently. I will throw an error instead here!

src/utils/validation/ProcessValidator.js
vdiezel: fixed the params issue. Re: the enum of types for streams: Yes, we should have something for the `expectedValueType` of streams, probably in the `Streams.js` utils. Ill add it.

src/components/processviewer/ProcessViewer.jsx
vdiezel: because I didn't touch it here, I will postpone finding a better name here :)

ConorTighe1995: Ok fair enough but in the future we should avoid calling objects things like object or params, lowers cohesion significantly and readability.

src/utils/validation/Validator.js
```diff
@@ -0,0 +1,9 @@
+export default class Validator {
+  static isBoolean = value => typeof value === 'boolean'
+
+  static isNumber = value => !Number.isNaN(value) && typeof value === 'number'
+
+  static withinRange = (value, [ min, max ]) => value <= max && value >= min
+
+  static lessThanOrEqual = (value, limit) => value <= limit
```
O-Susanoo, line 8: nice

src/utils/validation/InputValidator.js
vdiezel: fixed.

src/utils/validation/ProcessValidator.js
vdiezel: fixed.

PR #712, opened 2020-04-27
renders all data when a user zooms in the live mode

### LINKED ISSUE #690: Graph not visible on live data after a while
Referene this pr #689 

### Current Behavior
 - when user zooms  into live data, if they wait long enough, data will start to disappear from the the viewport from left to right. This is because we render only the latest 120 data points in the live view.
---

 - when a user zooms in on live data, the data in viewport should persist. This can be solved by chagning the behaviour of the zoom in the `live` mode to not restrict the amount of displayed datapoints or render all datapoints from the first one visible to the user in the zoomed viewport!
 
---

### Steps to Reproduce
 - zoom on live data graph (preferable at the area with the earliest data points) and wait
---
graph

vdiezel:
closes #690 

Now, when the user is in the live mode and has zoomed in, we render all available data. This doesn't only prevent the graph from disappearing but also allows panning to the left. 

OLD: 

live_zoom_old

NEW: 

live_zoom_new


src/utils/DataManager.js
```diff
@@ -98,12 +98,15 @@ class DataManager {
...
+  static sliceStreamData(process, data, mode, selectedPhase, CONSTANTS, zoomParams) {
     if (data.length === 0) return []
 
     switch (mode) {
-      case 'live':
-        return data.slice(-CONSTANTS.POINTS_IN_LIVE_VIEW)
+      case 'live': {
+        const { XviewportMinimum } = zoomParams
+        if (XviewportMinimum !== null) return [ ...data ]
+        return data.slice(-CONSTANTS.POINTS_IN_LIVE_VIEW5)
```
O-Susanoo, line 108: Niiiiiiiiiiice clean and clear :+1:

PR #713, opened 2020-04-27
710 707 password ux

### LINKED ISSUE #707: visual bug in events
fix dis:
image

### LINKED ISSUE #710: Password Reset UX
### Current Behavior
 - as a user, when i try to reset my password, and I have an invalid password string provided, I get a non-helpful message

---

 - [x] as a user, when I reset my password and get a user error from incognito, I get a helpful response shown on the screen that will enable me to fix my behavior, be it the password or anything else (check cognito responses and make sure we are displaying them when appropriate)
- [x] as a user, there are two input fields that I need to match my password in when resetting my password
- [x] as a user, chrome does not autofill the password field when I am doing reset password! 

image

DanielSeehausen:
Closes #707 
Closes #710 

I tried to keep the changes to a minimum for both tickets. First commits are for the user password reset. Last commit is for the line break in 707. 

# #710

Now the user can't submit without matching passwords
image

When cognito fails it will not display the cognito message (this will display issues with password validation also, which I can't recreate at the moment because I have blown the cognito api limit)
image

# #707 
Events with actionable stream links:
image

Events:
image


src/components/authentication/ResetPassword.jsx
```diff
@@ -83,7 +90,12 @@ class ResetPassword extends React.Component {
     }
   }
 
-  handleUpdate = (event, {name, value}) => this.setState({[name]: value, isSuccess: false})
+  handleUpdate = (event, { name, value }) => (
+    this.setState({
+      [name]: value,
+      isSuccess: false,
```
vdiezel, line 96: do you also have to reset the `isError` here, so exsiting error messages disappear?

DanielSeehausen: @vdiezel I don't believe so -- this fires whenever a user is doing something like changing their username. I think removing an error would be a little too easy in that case. This method didn't actually change outside of indentation!

PR #721, opened 2020-04-28
implement oneofStream

### LINKED ISSUE #705: Multi Select Downstream values

This feature is to implement a view for the users in the builder that will handle when a stream expects a 'oneOf' data type. This will be used when the user should select one option from a series of options.

Following is an example of a downstream payload. This is to enable the `oneOf` option!

```
    streams: [
        // ...other streams
        {
            id: "123469",
            type: "downstream",
            name: "Heating Fan Front",
            groupName: "Heating System",
            unit: "%",
            max: 100,
            min: 0,
            default: 70,
            expectedValueType: <string, one of: Set(['boolean', 'float', 'integer', 'oneOf'])>,
            expectedValueOptions: [ <string, Set('value1', 'value2', ... 'valueN')> ] (optional)
        },
    ]
```
---

### Deliverables

As a user on the process builder...
when I have a multiple choice option for some downstream value,
I can click on a dropdownable field and select from a series of options. These options should persist, like other downstream values
If there is only one option (i.e. the `expectedValueOptions` array looks like: ['leftBag'], the field will not 'drop down', and instead just display that only option value filled in. 

Check out the existing architecture for how we handle different `expectedDataTypes` and what components are rendered. @vdiezel might have some good information on this. Since this will only expand, if its not already organized in an extendable manner, take the time to refactor how its done so appropriate components are matched up with `expectedDataTypes`. A higher order component might be in order. 

Please provide test coverage that covers at least that a dropdown is populated when a stream has a `oneOf` expected data type. All `oneOf` expected types require `expectedValueOptions`. If that is not present, we need a fallback option. Perhaps providing a `console.warn` and  not displaying the stream at all would be a good option? Perhaps making it even louder and doing a `window.alert` to alert that incompatible streams are being sent up?
---

### Additional context
Confirm with Denis/Tim once this is complete (is merged in)
---

### LINKED ISSUE #718: reviewing a chart
All information you provide is helpful to us! Thank you for taking the time.

**Describe the bug**
When expanding collapsed graphs under certain circumstances, the viewport moves around and leads to a poor user experience.

**To Reproduce**
1. Go to 'celltainer device'
2. Go to a completed process that has data
3. Collapse all graphs
4. Open a group
5. Open the top stream graph in the group
6. Scroll your screen down and then open the bottom stream graph in the group
(the viewport will jerk upwards)

(if you can't recreate, ping daniel)

**Expected behavior**
When opening charts, the viewport shouldn't move it all

**Environment**
all environments!

O-Susanoo:
OneOfStream

some works had also needed to be done on all area impacted by this change such as

image

image

image

closes #705 
closes #718 

I should mention that i didn't take too much time to fix all the issues with this view because from the look of it it's really really deep

image

but i have just fixed only very obvious such the toggleFold function and redux


src/components/processviewer/body/builder/Group.jsx
```diff
@@ -76,7 +87,7 @@ const Group = ({
       case 'boolean':
         return renderBooleanStream(stream, idx)
       default:
-        return renderStream(stream, idx)
+        return renderFreeValueStream(stream, idx)
```
ConorTighe1995, line 101: `idx`? is this `groupId`?

src/components/processviewer/body/builder/StreamTypes/FreeValueStream.jsx
```diff
@@ -0,0 +1,209 @@
...
+  async changeDisplayedUnit(event, { value: newUnit }) {
+    this.setState({
+      selectedUnit: newUnit,
+      takingInput: false,
+    })
+
+    // set remotely the chosen unit as a default if user has correct access
+    const { activeDevice, dispatchCallUpdateDevice, stream, user } = this.props
+
+    if (!Authorizer.user(user).hasEditRights(activeDevice)) {
```
ConorTighe1995: This looks like  something that should happen at the start of the function before making changes to the state, maybe even out into a functions and called at that start of save too.

src/components/processviewer/body/builder/StreamTypes/FreeValueStream.jsx
```diff
@@ -0,0 +1,209 @@
...
+
+    const processTarget = DescriptionParser.getStreamTarget(activeProcess, phaseId, id)
+    const displayedTarget = this.getDisplayedTargetValue(processTarget)
+
+    const [ displayedMin, displayedMax ] = convertManyFrom(unit).to(selectedUnit)([min, max])
+
+    const limits = `min: ${displayedMin} ${selectedUnit} | max: ${displayedMax} ${selectedUnit}`
+
+    const unitOptions = getUnitOptions(stream.unit)
+      .map(unit => ({ text: unit, value: unit }))
```
ConorTighe1995: Should this duplicate assignment happen here? Could be wrong looks like this should be `unit.text` and `unit.value`

src/components/processviewer/body/builder/StreamTypes/FreeValueStream.jsx
```diff
@@ -0,0 +1,209 @@
...
+
+    const disabledInput = !phaseIsModifiable(activeProcess, phaseId)
+
+    return (
+      <div>
+        <Popup
+          trigger={(
+            <span
+              style={{ fontWeight: 'bold' }}
+              onMouseEnter={() => this.setState({ nameHovered: true })}
```
ConorTighe1995: Opportunity for a onHoverToggle function here that simply does this.setState({ nameHovered: !nameHovered })}

src/components/processviewer/body/builder/Group.jsx
O-Susanoo: just an index value

src/components/processviewer/body/builder/StreamTypes/FreeValueStream.jsx
O-Susanoo: thank you conor for the remarks but FreeValueStream.jsx is just me renaming the file Stream.jsx , i already talked about this with viktor and it seem we are going to do this later

O-Susanoo:
i think even tough the code is done it will not go until hardware create a real oneOfStream most of stuff  i worked on are fake, so i'm sure once they have the stream plugged stuff could came up but feel free to add your comments if you have any i wished if i had time to add more tests but i think that can be good enough for now , if you guys think this may need more coverage then i will try to add more

src/components/device/modals/SetStreamDefaultModal.jsx
```diff
@@ -140,7 +196,9 @@ class SetStreamDefaultModal extends React.Component {
       <>
         <Form.Field>
           <label>
-            Current Default [{displayedUnit}]
+            Current Default [
+            {displayedUnit}
+            ]
```
DanielSeehausen, line 201: minor lint thing, also below on 212

src/components/processviewer/body/builder/StreamTypes/OneOfStream.jsx
```diff
@@ -0,0 +1,178 @@
...
+    } = this.props
+    if (!stream.expectedValueOptions.includes(value)) return
+    // update process in applications state
+    dispatchSetStreamTarget(activeProcess.id, phaseId, stream.id, value)
+
+    // update remote process ( in case it hasn't started yet)
+    if (activeProcess.state === PROCESS_STATES.executable) {
+      await Process.save(activeProcess, dispatchCallUpdateProcess)
+    }
+  }
```
DanielSeehausen, line 72: should functionality like this be shared across the different components? The only thing that is specific to oneOfStreams is the check on line 64.

src/components/processviewer/body/builder/StreamTypes/OneOfStream.jsx
```diff
@@ -0,0 +1,178 @@
...
+    if (!Authorizer.user(user).hasEditRights(activeDevice)) {
+      return
+    }
+
+    const updatedConfig = DeviceConfig
+      .updateStreamConfig(activeDevice, stream.id, { defaultUnit: newUnit })
+
+    const updateParams = { config: updatedConfig }
+    await dispatchCallUpdateDevice(activeDevice.id, updateParams)
+  }
```
DanielSeehausen, line 54: I am less familiar with this cod, so forgive me if there are obvious answers to this question:

is this specific to OneOfStreams component, or something all of these streams need to do?

src/components/processviewer/body/builder/StreamTypes/OneOfStream.jsx
```diff
@@ -0,0 +1,178 @@
...
+    const streamValueIsConvertible = hasUnit(stream.unit)
+    const options = Stream.getStreamOptions(stream.expectedValueOptions, stream.unit, selectedUnit, streamValueIsConvertible)
+    const displayedTarget = DescriptionParser.getStreamTarget(activeProcess, phaseId, id)
+    const disabledInput = !phaseIsModifiable(activeProcess, phaseId)
+
+    const unitOptions = getUnitOptions(stream.unit)
+      .map(unit => ({ text: unit, value: unit }))
+    const uniqueUnit = !!(streamValueIsConvertible && unitOptions.length === 1)
+    const uniqueValue = stream.expectedValueOptions.length === 1
+
```
DanielSeehausen, line 148: a lot of these declarations are made just to be passed as an argument. consider deriving these in the methods that use them instead.

src/components/processviewer/body/builder/StreamTypes/OneOfStream.jsx
```diff
@@ -0,0 +1,178 @@
...
+              openValueDropDown,
+              disabledInput,
+              options,
+              uniqueUnit,
+              openUnitDropDown,
+              unitOptions,
+              selectedUnit,
+            })
+            : this.renderUnitlessStream({
+              displayedTarget, uniqueValue, openValueDropDown, options,
```
DanielSeehausen, line 170: this is a lot of methods to pass
if there isn't a compelling reason that they need to be passed, these might be better derived in the methods themselves.

src/components/processviewer/body/builder/StreamTypes/OneOfStream.jsx
```diff
@@ -0,0 +1,178 @@
...
+        value={displayedTarget}
+        open={uniqueValue ? false : openValueDropDown}
+        readOnly={disabledInput}
+        options={options}
+        closeOnChange
+        onClick={() => this.setState({ openValueDropDown: !openValueDropDown })}
+        onChange={this.save}
+        onClose={() => this.setState({ openValueDropDown: false })}
+        style={{ paddingRight: '0', marginRight: '-2px', borderRadius: '.28571429rem 0 0 .28571429rem' }}
+      />
```
DanielSeehausen, line 96: this looks like the same component as `renderUnitlessStream` with the exception of styling

O-Susanoo:
Hey @DanielSeehausen thank you for the review , i will try to fix some but not all , in fact i already had this discussion with @vdiezel about it , because as yourself i noticed a lot of duplicate code , and we (me and @vdiezel ) agreed to make anther pr where we will actually focus on doing the refactor for all the streams

src/components/processviewer/body/builder/StreamTypes/OneOfStream.jsx
O-Susanoo: it seem to be something shared among all of the other downstream

src/components/processviewer/body/builder/StreamTypes/OneOfStream.jsx
O-Susanoo: absolutely right i will make this fix

src/components/processviewer/body/builder/StreamTypes/OneOfStream.jsx
O-Susanoo: you are right i will also do that

src/components/processviewer/body/builder/StreamTypes/OneOfStream.jsx
O-Susanoo: yes i think i could make a function for this also

src/components/processviewer/body/builder/StreamTypes/OneOfStream.jsx
```diff
@@ -0,0 +1,178 @@
...
+      <Select
+        compact
+        open={uniqueUnit ? false : openUnitDropDown}
+        options={unitOptions}
+        value={selectedUnit}
+        closeOnChange
+        onClick={() => this.setState({ openUnitDropDown: !openUnitDropDown })}
+        onClose={() => this.setState({ openUnitDropDown: false })}
+        onChange={(e, data) => this.changeDisplayedUnit(e, data)}
+        style={{ borderRadius: '0 .28571429rem .28571429rem 0' }}
```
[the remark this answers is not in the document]
O-Susanoo: well it depend on streams but i think it should

src/components/processviewer/body/builder/StreamTypes/FreeValueStream.jsx
vdiezel: I think the idea here is to allow an unauthorized user to change units locally but not save them as defaults for the device

src/components/processviewer/body/builder/StreamTypes/FreeValueStream.jsx
vdiezel: unit is a string that is mapped into an object for the semantic ui react dropdown API that expects an array of object with the keys `text` and `value`

src/utils/units.js
```diff
@@ -181,3 +181,6 @@ export const applyUnitConversionOnData = (data, converter) => {
     dataSlice.dataPoints = applyUnitConversionOnDataPoints(dataSlice.dataPoints, converter)
   })
 }
+
+export const hasUnit = unit => unitToUnitType(unit) !== undefined
+&& unitToUnitType(unit) !== UNIT_TYPES.UNITLESS
```
vdiezel, line 186: This workaround shouldn't be necessary - the unit conversion should be stable enough to not break and just return the input value as is when no unit for a conversion is found but I think we talked about this and I will probably remove this in the process of the implementation for RPM calibration by casting units to lower case for conversion (which was the breaking bug here)

src/components/device/modals/SetStreamDefaultModal.jsx
```diff
@@ -140,7 +196,8 @@ class SetStreamDefaultModal extends React.Component {
       <>
         <Form.Field>
           <label>
-            Current Default [{displayedUnit}]
+            Current Default
+            {` [${displayedUnit}]`}
```
vdiezel, line 200: that looks excessive :japanese_ogre: ...guess lint made you do this.

src/components/device/setup/StreamDefaultsTable.jsx
```diff
@@ -2,65 +2,54 @@ import React from 'react'
...
-          {displayedName} [{displayedUnit}]
-        </Table.Cell>
-        <Table.Cell>
-          {displayedCustomTarget}
-        </Table.Cell>
-      </>
-    )
+  renderFloatStream = ({ targetDefault, unit, displayedUnit, displayedName }) => {
+    const streamIsConvertible = hasUnit(unit)
+    const formattedName = streamIsConvertible ? `${displayedName} [${displayedUnit}]` : displayedName
```
vdiezel, line 45: If the only reason why you did this ternaries was to prevent the dash to appear for streams without unit, there is a functions that does that for you: `replaceUnitCharacter(displayedUnit, '')` Then you can ommit all the ternaries around and your code

src/components/processviewer/body/builder/PhaseTabContent.jsx
```diff
@@ -55,6 +64,11 @@ class PhaseTabContent extends React.Component {
     await this.save()
   }
 
+  toggleFold = groupName => {
```
vdiezel, line 67: the group name never gets passed into the function, see line 168

src/components/processviewer/body/builder/Unit.jsx
```diff
@@ -0,0 +1,40 @@
...
+        value={selectedUnit}
+        closeOnChange
+        onClick={() => this.setState({ openUnitDropDown: !openUnitDropDown })}
+        onClose={() => this.setState({ openUnitDropDown: false })}
+        onChange={(e, data) => this.update(e, data)}
+        style={{ borderRadius: '0 .28571429rem .28571429rem 0' }}
+      />
+    )
+  }
+}
```
vdiezel, line 40: :fire:

src/components/processviewer/body/processconfiguration/ProcessConfiguration.jsx
```diff
@@ -2,37 +2,37 @@ import React, { useState } from 'react'
...
+import { convertFrom, hasUnit } from 'utils/units'
 
 const renderFloatStream = (stream, activeDevice, phases) => {
   const streamUnit = DeviceConfig.getDisplayedStreamUnit(activeDevice, stream.id)
+  const streamIsConvertible = hasUnit(stream.unit)
   return (
     <Table.Row key={stream.id}>
       <Table.Cell>
-        {stream.name} [{streamUnit}]
+        {streamIsConvertible ? `${stream.name} [${streamUnit}]` : stream.name}
```
vdiezel, line 13: same appliies here regarding `replaceNoUnitCharacter` helper function

test/components/processviewer/builder/Group.test.jsx
```diff
@@ -0,0 +1,139 @@
...
+    test('should render a freevalue stream', async () => {
+      const { getByText } = renderStreamGroup(executableProcess)
+      const oneOfString = 'Sensor 4'
+      const container = getByText(oneOfString).parentNode
+      const componentType = container[Object.keys(container)[0]].return.elementType
+      expect(componentType.name).toBe('FreeValueStream')
+    })
+
+  })
+})
```
vdiezel, line 139: very nice

test/components/processviewer/builder/StreamTypes/BooleanStream.test.jsx
```diff
@@ -0,0 +1,49 @@
...
+    id: '1',
+    type: 'downstream',
+    unit: '-',
+    name: 'Sensor 1',
+    groupName: 'Pump',
+    expectedValueType: 'boolean',
+  },
+]
+
+describe('OneStream.jsx', () => {
```
vdiezel, line 16: this is the boolean stream :)

test/components/processviewer/builder/StreamTypes/BooleanStream.test.jsx
```diff
@@ -0,0 +1,49 @@
...
+      disabledInput={false}
+      updateStreamTarget={updateStreamTarget}
+    />,
+  )
+
+  describe('when rendering a stream', () => {
+    test('should render a checked checkbox', async () => {
+      const stream = streams[0]
+      const { getByText } = renderOneStream(stream, true, jest.fn())
+      const checkBox = getByText(stream.name).parentNode.childNodes[1].childNodes[0].childNodes[0]
```
vdiezel, line 31: this approach would makes the tests very flaky because it depends on hierarchy order which might change pretty often. Is there no other way? I would even prefer test-Ids

DanielSeehausen:
updated branch for ya

src/components/device/modals/SetStreamDefaultModal.jsx
O-Susanoo: yes lint can be so naggy sometime :)

test/components/processviewer/builder/StreamTypes/BooleanStream.test.jsx
O-Susanoo: well i agree with you but i always try to avoid test-ids i feel like they pollute the code for no good reason , but if you feel strongly about it i can change it

src/components/device/modals/SetStreamDefaultModal.jsx
vdiezel: what is also possible is using either `&nbsp;` or `{' '}` to insert spaces

src/components/processviewer/body/builder/StreamTypes/FreeValueStream.jsx
ConorTighe1995: that makes sense

PR #724, opened 2020-04-30
Feature: Mark Processes as Favorites

### LINKED ISSUE #612: PIN 2: Update Widgets
#### Deliverables
  - [x] Update process icons (4 in total) color and icon appropriate for status
    - Green executable color should match green on device online
    - Finished process color required (use current grey of icon)
    - Process Paused Symbol required (double line)
  - [x] Remove status text from process widgets (see image)
  - [x] move edit to be inline with the title (like device widget)
  - [x] remove line separating the revealable circle icons and rest of widget
  - [x] display all icons justified left, except for delete, which is justified right
  - [x] delete icon becomes same trash icon as the delete phase trash icon (see image)

image

image

### LINKED ISSUE #613: PIN 3: UI for setting process as pinned
blocked by: ospin-web-dev/hambda#318
blocked by: ospin-web-dev/REACTor#612

#### User Story

As a user, when I hover over a process widget, I see an icon that I recognize as 'pinning'. When I click this icon, I see the process automatically move directly after the 'Build Process' icon, OR if there are already pinned process(es) present, it moves to its respective place in chronological order among the pinned processes, which are always the first group of process widgets after the 'Build Process'.

#### Deliverables
  - [x] update the process object mapper to account for a `pinned` key boolean. if none is found, defaults to false
  - [x] add grey star outline icon to set a process as pinned (find appropriate star icon from semantic) should be on the bottom left of the widget (where the executable/finished circle is now)
  - [ ] when a user clicks this, the process object gets its `pinned` boolean toggled, and the icon receives a change in color when pinned vs. the normal icon grey when not pinned. If you want to get fancy, show it transition to that position for the user. 
  - [x] update the processes display to always display pinned processes first (directly after the "Build Process" in the tile view, and as the top rows in the table view). This should be the role of the display or view, not the store. @vdiezel mentioned `Processes.jsx` might delegate a list to both the table and tile view, so the sorting should likely happen there! This is high priority for testing, at least given a collection of processes with some pinned, they are displaying in the appropriate order (and click testing on pinning if ambitious)
  - [x] Pinned processes should always display their star icon, whether pinned or not. 
  - [x] this should fire a call to

DanielSeehausen:
As some processes end up being targeted as 'high value, repeatable processes', our users found it important to have a way in which they could bookmark/pin/save these processes for quick access later. 

This is particularly useful for a process that you have taken the time to get in the exact configuration you like, that you anticipate needing to clone and run often in the future. 

We are excited to introduce favorite/pinnable processes!

On the process widget, you will find a star icon next to the other process utility icons. This star can be toggled to pin/unpin a process to the top of your process list. Pinned processes will always be first/at the top in your processes list.

image

...and to piggyback a small UI update, we have made several small changes to process widgets in general. Most notably, process tiles have been provided with additional imagery and colors to help distinguish their state (running vs executable vs finished vs paused):

image

\private

closes #612 
closes #613 

The commit history before April 30 is pretty hairy so it might be easiest to either only review the April 30th commits onwards or check out file diff only. Sorry for that. Had a tricky merge that I didn't want to mess with because this PR was open for far too long.


src/components/device/processes/ProcessTileIconBar.css
```diff
@@ -0,0 +1,21 @@
...
+
+
+
+
+
+
+
+
+
+
```
O-Susanoo, line 21: that's a lot of empty lines

src/redux/thunks/process/callUpdateProcessOptimistic.js
```diff
@@ -0,0 +1,13 @@
...
+
+const callUpdateProcess = (dispatch, processId, params) => {
+  nexus.process.update(processId, params)
+  dispatch(updateProcess({ processId, params }))
+  if (params.description) {
+    dispatch(updateProcessDescriptionSnapshot({ processId, params }))
+  }
+}
+
+export default thunkWrapper(callUpdateProcess)
```
vdiezel, line 13: just for the being persistent, can you rename this function into `callUpdateProcessOptimistic`. Will make the lookuip easier when you dont know the file name

src/utils/process.js
```diff
@@ -231,4 +231,17 @@ export class Process {
     }
     return PhaseProgression.getLatest(process) || PhaseProgression.noPhaseObject()
   }
+
+  static sortPinnedFirst(processes) {
+    const samePinnedState = (a, b) => (a.pinned === b.pinned)
+    const byCreatedAt = (a, b) => (a.createdAt < b.createdAt ? 1 : -1)
```
vdiezel, line 237: I think here also `a.createdAt - b.createdAt` would do the job

src/components/device/processes/ProcessTileIconBar.css
DanielSeehausen: lol oops! good 👁️ @O-Susanoo

PR #725, opened 2020-04-30
464 Add description of user roles on hover.

### LINKED ISSUE #464: Provide Descriptions of User Roles
Provide a user accessible description of what the different roles do.

Just spitballing here:
Perhaps a button that opens a modal on the access page (or hover + tooltip that doesn't open a modal)? Anchored right all the way across from the `Add User` button?
image

Alternatively, github has a nice interface here we could potentially mimic:
image

There may be a react component package to provide a rich dropdown. 
Use your best judgement!

The following questions should either be implicitly, or explicitly, answered via the descriptions:

- Is there a role that supersedes all others?
- Who can take the ownership role away from other users?
- What does the little crown in the access table mean?

To go in after #429 has been merged. 
Might be simpler to do after #465 has been completed, so we don't need to update the descriptions
Addresses part of #462

ConorTighe1995:
Descriptions have now been added for user permissions. We hope this helps you when selecting what access rights to assign to people on devices and processes!

\private
I updated the access drop down to display a description of the user roles on hover of the item. The dropdown is now its own reusable component for our forms. Ive added the descriptions in a gif below, its pretty long because its shows of the text and both changes.

closes #464


src/components/device/modals/AccessDeviceModal.jsx
```diff
@@ -27,12 +26,13 @@ const AccessDeviceModal = props => {
     })
   }
 
-  const handleGroupChange = (event, data) => {
-    const assignedGroupData = props.activeDevice.access.find(group => group.name === data.value)
+  const handleGroupChange = value => {
+    console.log(value)
```
O-Susanoo, line 30: noice , but i think you forgot this console.log here

src/components/device/modals/AccessDeviceModalForm.jsx
```diff
@@ -1,27 +1,59 @@
...
-      options={groupOptions}
-      onChange={handleGroupChange}
-      placeholder='Choose group'
-    />
-  </>
-)
+const descriptions = ['Has full control of the device and its process, an admin can delete, edit devices and processes on the device',
+  'Can stop and start devices alonf with create processes but cannot delete devices or processes',
+  'Can only view the processes on a device.']
+
```
O-Susanoo, line 8: i think if you made this an object with keys being role and value being description would be much better because you will not have to count on order to display data and it will give the reader a better understanding of who is who

src/components/device/modals/AccessDeviceModal.jsx
ConorTighe1995: yep your right ill axe that now

src/components/device/modals/AccessDeviceModalForm.jsx
ConorTighe1995: Cool ill try that out now

src/components/device/modals/AccessDeviceModalForm.jsx
```diff
@@ -1,27 +1,65 @@
...
+  groupOptions,
+  handleUserUpdate,
+  handleGroupChange,
+  selected,
+}) => {
+
+  const descriptions = {}
+  descriptions.admins = 'Has full control of the device and its process, an admin can delete, edit devices and processes on the device'
+  descriptions.operators = 'Can stop and start devices alonf with create processes but cannot delete devices or processes'
+  descriptions.viewers = 'Can only view the processes on a device.'
```
O-Susanoo, line 15: you can actually do this

```
const descriptions = {
admins: 'Has full control of the device and its process, an admin can delete, edit devices and processes on the device',
operators:'Can stop and start devices alonf with create processes but cannot delete devices or processes',
viewers: 'Can only view the processes on a device.'
}
```

ConorTighe1995:
I need to update the text before merging, maybe add a test or two to stop future changes to roles from breaking it too.

src/components/utility/dropdown/FormAccessDropdown.jsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react'
+import { Form, Dropdown, Popup } from 'semantic-ui-react'
+
+const AccessDropdown = props => {
+  const { selected, groupOptions, onUpdate, label, placeholder } = props
+
+  const descriptions = {
+    admins: 'Has full control of the device and its process, an admin can delete, edit devices and processes on the device',
```
vdiezel, line 8: this description seems to have a copy-paste error. The first part looks good to me: 
"Has full control of the device and its process__es__"

src/components/utility/dropdown/FormAccessDropdown.jsx
```diff
@@ -0,0 +1,45 @@
...
+    operators: 'Can stop and start devices alonf with create processes but cannot delete devices or processes',
```
vdiezel, line 9: "Can create and run processes on the device" should be ok here!

src/components/utility/dropdown/FormAccessDropdown.jsx
ConorTighe1995: I forgot to push those changes yesterday, good spot.

ConorTighe1995: Forgot to push yesterday, well spotted!

src/components/device/processes/ProcessTile.jsx
```diff
@@ -188,7 +188,7 @@ const ProcessTile = props => {
     if (confirmed) {
       const result = await dispatchCallDeleteProcesses([ process.id ])
       if (result.status === 403) {
-        FlashMessenger.error(result.message)
+        FlashMessenger.error('Operation failed. No sufficient access rights to one or more selected processes.')
```
vdiezel, line 191: This flash message can stay as it is, because here you can selct only one process anyway!

Issue #458, opened 2020-01-14
ADD HAMBDA TICKET FOR MEANINGFUL RETURN MESSAGES  COMPLEXITY 3) As a user with view rights to a process, deleting it results in an failure without communication to the user. 


DanielSeehausen:
### Issues with the users experience
**As a user with only view rights to a process:**
- attempt to delete a process tile: receive no feedback as to why it failed

**As a user with only view rights to a process, and owner rights to multiple other processes:**
- attempt to delete many processes in the table view with the delete all action: operation fails without user feedback

### Deliverables
- [ ] As a user, when I attempt to delete a process and get back a 403, I should be presented with some useful information on the front end
- [ ] As a user, when I attempt to delete many processes, and I have insufficient rights for some, or all, of them (via receiving a 403), I should be presented with some useful information on the front end so I can change my selection and or behavior (_this one may require a small amount of backend work, as right now the specific process(es) that fail are not returned from the server_)

here is where you can switch to table view:
![image](https://user-images.githubusercontent.com/17087405/72333133-9c763c00-36bb-11ea-82f2-fc0067f78b9a.png)

PR #727, opened 2020-05-04
458 Unauthorized process delete attempt will trigger 403 flash message.

ConorTighe1995:
Now when someone tries to delete a flash a process form the table or tile view and does not have the correct permissions they will see the following:

I also updated the edit icon to show rename:

rename-hot-fix

closes: #458


src/components/device/processes/ProcessesTable.jsx
```diff
@@ -8,39 +8,31 @@ import FlashMessenger from 'utils/FlashMessenger'
...
 
 const mapStateToProps = ({ user, deviceUsers }) => ({ user, deviceUsers })
 
 class ProcessesTable extends React.Component {
 
+  constructor() {
+    super()
+    this.checkProcess.bind(this)
+  }
+
```
O-Susanoo, line 23: You can actually remove this and make the function checkProcess an arrow function , fun fact arrow function are by default binded to class context

ConorTighe1995: I was going to do that but seeing that all the other functions are defined as class functions I thought id stick to the pattern defined in the file, I only added this constructor to get rid of a linting error as it was being binded when passed as a prop

PR #730, opened 2020-05-04
Fix CheckForNewStreams

### LINKED ISSUE #729: uIConfig not updating when stream change groups

O-Susanoo:
closes #729 

the issue was even tough the add was correct but checking for new streams was wrong because the check was on stream ids but without taking into account the group , however i should mention something really weird , i deleted adel uiConfig and even with that somehow his view was always showing old config (prod) , i deleted it again and same story but when i logged with master (point it to prod) with his account it worked , i deleted the config one more time and asked him to logIn again but this time somehow the uiConfig was correctly generated , i don't know how this stuff happend because as a reminder no code was changed in this entire scenario


src/utils/UIConfig/StreamGroups.js
```diff
@@ -75,10 +75,11 @@ export default class StreamGroups {
...
 
   static checkForNewStreams(groups, streams) {
-    const streamsIds = streams.map(stream => stream.id)
-    const backendStreamIds = groups.map(group => group.streams).flat().map(stream => stream.streamId)
-
-    return streamsIds.some(streamId => !backendStreamIds.includes(streamId))
+    return streams.some(stream => {
+      const uiStreamGroup = groups.find(group => stream.groupName === group.name)
+      if (!uiStreamGroup) return true
+      return !uiStreamGroup.streams.find(uiStream => uiStream.streamId === stream.id)
```
vdiezel, line 81: `uiStreamGroup.streams.some(...)` :)

O-Susanoo: i hesitated to do that :)

PR #734, opened 2020-05-06
small text changes to alerts components and tables

### LINKED ISSUE #697: Alerts View
(this isn't urgent ignore that lavel I just want to remember it to discuss)
Alerts view + modal
- Make the type of alert more descriptive for the users with minor copy changes
- provide user feedback for invalid input values

#### In the table
'upper alert threshold' -> 'upper boundary' (this will need to be updated for all boundary types)
'threshold: 40' -> '40'
'actions' -> 'delete' or 'remove'
should the unit in `Temperature value [°C]` be on the parameter?

image

#### In the modal
- 'upper alert threshold' -> 'upper boundary' (same as above - this should be changed for all cases)
- 'Current upper threshold' -> 'Current upper boundary'
- 'New upper threshold' -> 'New upper boundary'
- If easy, make current upper threshold
- as a user, when I put an invalid value into the new target field, I am alerted why its invalid (currently there is no feedback) [EDIT by @vdiezel: This can be only done for upstreams that do have a related downstream target, otherwise there is no information about if an upstream alert threshold is invalid - firmware will implement valid ranges for sensors in future though]

image

---

### Additional context
ask viktor about this stuff before starting as it might have larger implication

vdiezel:
closes #697 

Does small changes to the Alert components and removes the 'actions' header from the tables in the column where the delete button lives

I did not add the validation because there is nothing to validate against for the upstreams! They don't have min/max values or targets. Some might have a linked downstream target, but that is the rare exception, so for now the validation is up to the user.


src/components/device/setup/UnitsTable.jsx
```diff
@@ -50,7 +50,7 @@ class UnitsTable extends React.Component {
           clickHandler={this.toggleShowSetUnitModal}
         />
         {configuredStreams.length !== 0 && <Table key={1} className='ospin-red' celled collapsing>
-          <TableHeader items={['Stream', 'Device Default', 'Custom', 'Info', 'Actions']} />
+          <TableHeader items={['Stream', 'Device Default', 'Custom', 'Info', '']} />
```
ConorTighe1995, line 53: Why are these now empty strings, if Actions is no longer needed then shouldn't the element be removed completely?

vdiezel: This is the text content within the table header - we want to keep an empty cell at top because we keep the whole column:
old:
Bildschirmfoto von 2020-05-08 13-06-11
new:
Bildschirmfoto von 2020-05-08 13-09-40

PR #737, opened 2020-05-06
Allowing empty display messages for alerts

vdiezel:
stricter checking for undefined (empty string counts as falsy which caused no message to appear on emptry string for optional displayMsg)


src/redux/helper/subscribeHandlers/eventHandler.js
```diff
@@ -25,7 +25,7 @@ function getMessagePrefix(deviceId) {
 
 function handleNewAlert(value) {
   const { eventType, displayMsg, deviceId } = value
-  if (!eventType || !displayMsg) {
+  if (eventType === undefined || displayMsg === undefined) {
```
ConorTighe1995, line 28: I feel theses should be 2 separate ifs with 2 separate warning to save the developer time.

DanielSeehausen, line 28: would you be able to add null? 
maybe a utility for null or undefined?
Especially because mongo will preference null, and displayMsg could reasonably be set to null somewhere
nothing fancy:

```
isNullorUndefined = (val) =>  ( val === null || val === undefined)
```

vdiezel: good point, will do!

PR #744, opened 2020-05-06
668 and more visual fixes

### LINKED ISSUE #392: Stop Flash notifications size change/variable sizes
flash notifications currently have variable size depending on the amount of text in them

this leads to notification bars suddenly changing sizes if a larger notification dissapears, while a short notification remains on screen (when the large one transitions out, the notification container likely resizes, and the smaller notification along with it.)

Ensure notification bars are always a certain width, no matter the amount of text content.

### LINKED ISSUE #668: Process Builder/Graphs Visual Difference
Process Builder group/stream color and cell schemas should match the data viewer closer:

See the difference here:

image

image

---

### Deliverables
As a user...
 - [ ] the process builder streams also appear in the separate colored box within the larger grey box, just like the graphs (but as of writing this ticket, they (the streams) should not be foldedable or re-arranged, which is correct, unlike graphs)

### LINKED ISSUE #732: Builder section looks weird when a group has only controller streams
Bildschirmfoto von 2020-05-05 17-24-05

it seem like the button is somehow also pushing the input when it's there which give this kind of view
image

vdiezel:
closes #668
closes #392 
closes #732 

Changes to builder: 

- Color scheme matches Data Viewer (#668)
- phase widget rework to be clickable everywhere and have same height as the other widgets. Hides it for finished processes
- fixes weird indentation for first input when group has controller streams (issue #732 second part)
- fixes weird display when group has only controller parameters (issue #732 first part)

Further:
- small rework of the add process tile (gets blue on hover now, font color changed to blue)
- fixes varying width for flash message (issue #392)
- BONUS: fixes undiscovered bug when deleting first phase

Bildschirmfoto von 2020-05-07 01-28-23
 
Bildschirmfoto von 2020-05-07 01-20-53

Bildschirmfoto von 2020-05-07 01-25-04


src/components/processviewer/body/builder/PhaseTabContent.jsx
```diff
@@ -85,46 +85,71 @@ class PhaseTabContent extends React.Component {
...
+            name={name}
+            changeNameHandler={this.submitNewPhaseName}
+            fontSize='1rem'
+            fluid
+          />
+          {this.renderDeleteButton()}
+        </div>
+        {this.renderChangesFooter()}
+        <Divider />
+      </>
```
DanielSeehausen, line 115: interesting - this seems to be messing up github's markdown parser!

src/components/processviewer/body/builder/PhaseTabContent.jsx
```diff
@@ -60,6 +60,31 @@ class PhaseTabContent extends React.Component {
...
 
+  renderDeleteButton = () => {
+    const {
+      activeProcess,
+      phasesLength,
+      phaseId,
+    } = this.props
+
+    const phaseNeverRan = !PhaseProgression.contains(activeProcess, phaseId)
+    const disableDeleteButton = phasesLength === 1 || (!phaseNeverRan && activeProcess.state !== 'executable')
```
ConorTighe1995, line 71: bit long, maybe capture do this:

```
const isExecutable = (!phaseNeverRan && activeProcess.state !== 'executable')
const disableDeleteButton = phasesLength === 1 || isExecutable
```

src/utils/process.js
```diff
@@ -160,7 +160,9 @@ export function movePhase(process, sourceIndex, destinationIndex){
 export function amendPhasesNextProps(process, deletedPhaseId, deletedPhaseNextKey){
   const amendedProcess = { ...process }
   const prevPhaseId = DescriptionParser.getPreviousPhaseId(amendedProcess, deletedPhaseId)
-  amendedProcess.description[prevPhaseId].next = deletedPhaseNextKey
+  if (prevPhaseId !== undefined) {
```
ConorTighe1995, line 163: this is the same yea?
`if (prevPhaseId) { ...do things }`

ConorTighe1995: Unless this can be null and still run code

src/components/processviewer/body/builder/PhaseTabContent.jsx
ConorTighe1995: Weird because I dont think ive seen fragments do that before

src/components/processviewer/body/builder/Group.jsx
```diff
@@ -16,11 +16,10 @@ const Group = ({
   const renderSettingsDropdown = () => (
     <Button
       onClick={() => toggleControllerModal(true)}
-      floated='right'
-      className='margin-r-8'
       labelPosition='left'
       icon
       type='button'
+      style={{ position: 'absolute', right: '8px', top: '8px' }}
```
ConorTighe1995, line 23: Should this be margin-right and margin left? Im just wondering because on different screen and devices top,left,right and bottom can lead to very different. I know it will be independent with position absolute but im just thinking about resizing the screen down the line and binding the button to a parent element, let me know what you think:

https://stackoverflow.com/questions/3859801/difference-between-margin-left-and-left-or-margin-right-and-right

https://stackoverflow.com/questions/20718577/why-do-absolute-elements-stack-up-on-each-other-instead-of-stacking-one-after-th

src/components/processviewer/phasesdisplaybar/PhasesNavbarWrapper.jsx
```diff
@@ -98,7 +98,9 @@ class PhasesNavbarWrapper extends React.Component {
                     style={{ maxHeight: '60vh' }}
                     {...provided.droppableProps}
                   >
-                    <Header textAlign='center' as='h3'>Process Tree</Header>
+                    <Header textAlign='center' as='h3' style={{ marginTop: '0px' }}>
+                      Process Tree
+                    </Header>
                     {phases.map((phase, i) => (
                       [
                         <div key={2 * i}>
```
ConorTighe1995, line 106: I know you didn't touch this but Is this done as a work around to the linter complaining? if so this is more just avoiding the problem than fixing it. Would a phase id not be suitable here?

src/utils/process.js
vdiezel: it can be 0!

src/components/processviewer/body/builder/Group.jsx
vdiezel: I am not sure if I understand:
margin-right and margin-top will display the button on the left side, with desired margin to the top and to the right. Position absolute positions the element relative to its parent element - isn't that what you call binding to a parent ? :)

src/components/processviewer/body/builder/PhaseTabContent.jsx
vdiezel: I made your changes and forgot to commit them before merging!

src/components/processviewer/phasesdisplaybar/PhasesNavbarWrapper.jsx
vdiezel: I made your requested changes and forgot to commit them before merging...

PR #751, opened 2020-05-11
746 add modal for rpm to flow

### LINKED ISSUE #746: Feature: RPM to flow/time unit conversion - input modal for ml/RPM
In the process builder, every stream that comes with a default unit of RPM should expose a button that opens up a modal. The modal should contain two things: 

1. a description of what input is expected (ml/rotation) and how to obtain this value
2. an input field to enter this value (ml/rotation)

The submit function should not persist anything yet!

vdiezel:
closes #746 

introduces a new modal for Pump flow rate calibration: 
pump_calibration

The button is displayed for all streams which come with the default unit of 'rpm'


src/components/processviewer/body/builder/PumpCalibrationWrapper.jsx
```diff
@@ -0,0 +1,41 @@
...
+  toggleModal = () => (
+    this.setState(({ showModal }) => ({ showModal: !showModal }))
+  )
+
+  render() {
+    const { streamUnit } = this.props
+    const { showModal } = this.state
+
+    return null
+    //if (streamUnit !== 'rpm') return null
```
ConorTighe1995, line 19: Was this left in by mistake?

src/components/processviewer/modals/PumpCalibrationModal.jsx
```diff
@@ -0,0 +1,57 @@
...
+    ratio: '',
+  }
+
+  updateRatio = (_, { value }) => (
+    this.setState({ ratio: value })
+  )
+
+  submitRatio = async () => {
+    const { ratio } = this.state
+    console.log('Submittinng', ratio)
```
ConorTighe1995, line 27: Looks like you left this log here by accident when developing.

src/components/processviewer/body/builder/PumpCalibrationWrapper.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+    return (
+      <>
+        <Button
+          primary
```
ConorTighe1995, line 24: Would you consider making this button grey? If you look at the settings button for controller in the gif it is grey too so they would match and I dont think the blue looks bad but its does stick out a lot on the page.

This pop up seems closer to a secondary interaction rather then a primary like start process or add phase.

src/components/processviewer/modals/PumpCalibrationModal.jsx
```diff
@@ -0,0 +1,57 @@
...
+
+class PumpCalibrationModal extends React.Component {
+
+  state = {
+    isError: false,
+    errorMessage: '',
+    ratio: '',
+  }
+
+  updateRatio = (_, { value }) => (
```
ConorTighe1995, line 21: Nice clean way to do this :100:

src/components/processviewer/body/builder/PumpCalibrationWrapper.jsx
vdiezel: no, its not by mistake - because I wanted this one to be mergeable without finishing the feature, I simply hid it by returning null. I just left the comment to show what the real code looks like

src/components/processviewer/modals/PumpCalibrationModal.jsx
vdiezel: This code will never be called as the component is never rendered in the first place - but you are right, I will clean up.

src/components/processviewer/body/builder/PumpCalibrationWrapper.jsx
vdiezel: you are right here

PR #758, opened 2020-05-13
Load only necessary processes data

### LINKED ISSUE #651: Load only a list representation of all processes
blocked back end ticket:
ospin-web-dev/hambda#425

We load way too much data into the UI when loading the processes of a device. Most of them will actually never be opened/interacted with. This leads to abundant network traffic and perfomance issues. The correct behaviour is this: 
on route `/device/{deviceId}/processes` load a list representation of the processes that contains the data that is actually needed by this component. 
when a user enters a process (`/device/{deviceId}/process/{processId}`) the full data of the process should be loaded. 

This should also allow to check properly for 403 or 404 when trying to render a single process.

O-Susanoo:
closes #651 

Sorry guys i didn't this ticket can go quickly out of control and end up modifying multiple files


src/components/utility/dropdown/FormAccessDropdown.jsx
```diff
@@ -15,7 +15,6 @@ const AccessDropdown = props => {
       fluid
       label={label}
       placeholder={placeholder}
-      selection
```
O-Susanoo, line 18: i have removed this but i'm adding it back , the reason why i removed it the first place is an error in the console , but i noticed that the error is not there in prod , and for those who wonder why i'm bringing it back is because the dropdown is not the same once removed

src/App.jsx
```diff
@@ -17,9 +17,9 @@ class App extends React.Component {
 
   render() {
     return (
-      this.props.authenticated === undefined ?
-        <Loader active={true} size="big"/> :
-        <Routes />
+      this.props.authenticated === undefined
+        ? <DynamicLoader />
```
O-Susanoo, line 21: this is what you want to look at , its the standard loader now

test/utils/updateFunctions.test.js
```diff
@@ -1,9 +1,8 @@
-import PDVC from '../../src/utils/PDVC'
 import { createProcessObject } from 'redux/helper/objectMapper'
```
O-Susanoo, line 1: Just Linting

test/mocks/processDataMock.js
```diff
@@ -1,120 +1,118 @@
...
-      "0": {
-        transition: 'manual',
-        name: 'Phase 1',
-        next: 1,
-        duration: 0,
-        downstreams: {
-          streamId1: { value: 37 },
-          streamId2: { value: 37 }
-        }
+export const getMockedProcessData = props => ({
```
O-Susanoo, line 1: Just Linting

src/nexus/src/process/get.js
```diff
@@ -1,8 +1,8 @@
...
-import config from '../amplify/config.js';
+import config from '../amplify/config'
 
 export default id => {
-  Amplify.configure(config);
-  return API.get('process', `${id}`);
-};
+  Amplify.configure(config)
+  return API.get('process', `${id}`)
+}
```
O-Susanoo, line 8: image

Not to have another branch on nexus

src/components/utility/DynamicLoader.jsx
```diff
@@ -0,0 +1,13 @@
...
+import { connect } from 'react-redux'
+import { Loader } from 'semantic-ui-react'
+
+const mapStateToProps = ({ collapsedSidebar }) => ({ collapsedSidebar })
+
+const DynamicLoader = ({ collapsedSidebar }) => (
+  <div style={{ position: 'fixed', top: '50%', left: `calc(50% + ${collapsedSidebar ? 28 : 90}px)` }}>
+    <Loader active size='big' />
+  </div>
+)
```
O-Susanoo, line 11: This Loader is the standard loader now :) it will dynamically adjust its size depending on whether the side bar is collapsed or not

ConorTighe1995: good stuff mo

src/components/device/Device.jsx
```diff
@@ -128,6 +132,16 @@ class Device extends React.Component {
       // render a 403 - in the current code flow this code can never be reached anyway
     }
 
+    useEffect(() => {
+      const fetchAndEnrichProcess = async () => {
```
ConorTighe1995, line 136: This code is duplicated across 2 files, maybe this should be a redux thunk dispatch?

src/redux/helper/objectMapper.js
```diff
@@ -37,6 +37,22 @@ export const createProcessObject = processData => ({
   eventTriggers: processData.eventTriggers || {},
 })
 
+export const createListProcessObject = processData => ({
```
ConorTighe1995, line 40: Id call this a Item as opposed to an Object, not a big deal tho.

src/components/device/Device.jsx
```diff
@@ -72,7 +76,7 @@ class Device extends React.Component {
   renderProcesses = () => {
     const { user, processes, deviceUsers } = this.props
     const activeDevice = this.getActiveDeviceFromUrl()
-    const activeProcesses = ProcessesUtils.getByDeviceId(processes, activeDevice.id)
+    const activeProcesses = Process.getByDeviceId(processes, activeDevice.id)
```
vdiezel, line 79: I moved the `getByDeviceId` to a new `Processes` utils because this was an operation on the processes collection (and not on a single process). Can we get it back to that state ? :)

src/redux/reducers/device/setActiveDevice.js
```diff
@@ -31,6 +32,6 @@ export default (state, action) => {
     deviceUsers,
     activeDeviceInitialized: true,
     devices: [ ...updatedDevices ],
```
vdiezel, line 34: if the spreading is not necessary, is it also not necessary on the updatedDevices?

src/components/device/Device.jsx
```diff
@@ -33,6 +34,9 @@ const mapDispatchToProps = dispatch => ({
   dispatchCallLoadDevice: (device, userId, deviceId) => (
     dispatch(callLoadDevice(device, userId, deviceId))
   ),
+  dispatchCallUpdateProcess: (processId, process) => (
```
vdiezel, line 37: please rename this from `dispatchCallUpdateProcess` to `dispatchUpdateProcess`. We have the convention that `call` indicates an API call, but here we only update the app state.

src/components/device/Device.jsx
```diff
@@ -128,6 +132,16 @@ class Device extends React.Component {
...
+    useEffect(() => {
+      const fetchAndEnrichProcess = async () => {
+        const process = await callLoadProcess(processId)
+        await dispatchCallUpdateProcess(processId, process)
+        enrichProcess(true)
+      }
+      if (!processIsEnriched) fetchAndEnrichProcess()
+    }, [processId, processIsEnriched, dispatchCallUpdateProcess])
+
+    if (!processIsEnriched) return (<DynamicLoader />)
```
vdiezel, line 144: Looks like you want to run useEffect only once here, and therefore you implemented the flag. How about this: 
https://css-tricks.com/run-useeffect-only-once/

vdiezel: And shouldn't the process viewer own this logic in the first place?

src/components/device/modals/CreateProcessModal.jsx
```diff
@@ -13,8 +13,8 @@ import ProcessValidator from 'utils/validation/ProcessValidator'
 const { VALID_PROCESS_DATA: { nameLength, commentLength } } = ProcessValidator
 
 const mapDispatchToProps = dispatch => ({
-  dispatchCallCreateProcess: (deviceId, processParams) => dispatch(callCreateProcess(deviceId, processParams)),
-  dispatchLoadProcessViewer : process => dispatch(loadProcessViewer (process))
+  dispatchCallAddProcessToProcessesList: process => dispatch(addProcessToProcessesList(process)),
```
vdiezel, line 16: here also please remove the `Call` part from the name to indicate that is not an API call.

src/components/device/modals/CreateProcessModal.jsx
```diff
@@ -88,18 +88,11 @@ class CreateProcessModal extends React.Component {
...
-      /* TODO: if device processes are desctructured at the beginning
-       * of the function, they don't have the new process yet and redirection
-       * doesn't work - find more stable solution
-       */
-
-      const { deviceProcesses } = this.props
-      const newProcess = deviceProcesses[0]
-
+      const newProcess = res.process
+      dispatchCallAddProcessToProcessesList(newProcess)
```
vdiezel, line 95: I think this is circular here: callCreateProcess thunk should not return the process but instead add it to the application state (you do it anyway here in line 95). This way you might not even need the AddProcessToProcessesList as an action anymore. You could also revert your changes to the createProcess thunk and apply the thunkWrapper again.

Or am I missing something?

src/components/utility/DynamicLoader.jsx
O-Susanoo: :+1:

src/components/device/Device.jsx
O-Susanoo: Yeah it's kind hard to remove the duplication without checking if a process is enriched or not , i had this discussion with viktor before but yeah you are absolutely right maybe next another pr should solve this

src/redux/helper/objectMapper.js
O-Susanoo: well i was just copying the same pattern as the one above it

src/components/device/Device.jsx
O-Susanoo: You got it

src/components/device/Device.jsx
O-Susanoo: done

src/redux/reducers/device/setActiveDevice.js
O-Susanoo: it was like that before i don't know if it's necessary but from the look of it it doesn't seem so

src/utils/process.js
```diff
@@ -197,6 +181,28 @@ export class Process {
...
+  static isFinished(process) {
+    return (!(Number.isNaN(process.finishedAt) || typeof process.finishedAt !== 'number'))
+  }
+
+  static isRunningOnDevice(process, device) {
+    return process.id === device.runningProcessId
+  }
+
+  static deriveProcessState(process, device) {
+    debugger
```
ConorTighe1995, line 197: Was this left in by mistake? If your using vscode and trying to debug here you can just use breakpoints and run in debugger mode.

src/components/device/Device.jsx
O-Susanoo: what they are saying in that is to use this

```
  useEffect(() => {
    // Run! Like go get some data from an API.
  }, []);
```
instead of this
```
  useEffect(() => {
    // Run! Like go get some data from an API.
  });
```
i think that's an old topic because you can't use the second one at all if you want to have argument

O-Susanoo: i don't think Processviewer.jsx should own that logic because it's already waiting for a process

src/utils/process.js
O-Susanoo: yeah this happend when daniel committed a refactor i'm aware of it , i have fixed it it will pushed soon

src/utils/process.js
```diff
@@ -197,6 +181,28 @@ export class Process {
     return (existingPhaseInProcess !== undefined && expectedNextPhaseId === nextPhase.phaseId)
```
ConorTighe1995, line 181: Maybe capture these in consts as the line is long and it would improve readability.

return phaseExists && hasNextPhase

O-Susanoo: i don't recall touching this part at all

src/utils/process/Processes.js
```diff
@@ -0,0 +1,7 @@
+class Processes {
+  static getByDeviceId(processes, deviceId) {
+    return processes.filter(process => process.deviceId === deviceId)
+  }
+}
+
+export default Processes
```
DanielSeehausen: I believe I see the thought here - to distinguish operations on the collection rather than a single process data object. That makes sense if that's what you are going for and sounds good.

That being said, unless you want to also take the time to move the other processes calls out of `Process` (which I would be game for), I think we should leave this one for now. Just moving one out and the others doesn't sound great for future devs. This is especially because we are already splitting this interface within the `Process` utils file itself, and haven't completely migrated to the `Process` interface yet.

vdiezel: This was orignally implemented by me - Mo actually removed it and I asked him to put it back. The idea was - as you suspected - to seperate operations on collections from single processes.

PR #761, opened 2020-05-14
658 start process feedback

### LINKED ISSUE #658: Give better visual feedback when waiting for a process to start
As a user, when I start a process there is no visual feedback for me to tell if my command has been fired and I'm just waiting for the process to start or if nothing happened at all. There should be an indicator to show the user that the start command has been fired and the UI simply waits for a response from the device. 

---

### Deliverables
As a user...
- when I start a process, I see some visual indicator to tell me that my start command worked and a response from the device is pending

ConorTighe1995:
Loader has been added for start process button.

closes #658


src/components/processviewer/headers/ProcessBuilderHeader.jsx
```diff
@@ -45,11 +47,16 @@ class ProcessBuilderHeader extends React.Component {
...
       deviceDescription: activeProcess.deviceDescription,
-      entryPhaseId: activeProcess.entryPhaseId
+      entryPhaseId: activeProcess.entryPhaseId,
     }
     await this.props.callUpdateProcess(activeProcess.id, updatedProcess)
   }
 
+  toggleLoader() {
+    this.setState(prevState => ({ processStarting: !prevState.processStarting }
+    ))
```
DanielSeehausen, line 57: this isn't even close to something blocking, but just in case it wasn't intentional
this looks like a possible auto-format newline? two parens may have escaped here (or maybe the object should be on a new line)?

ConorTighe1995: The linter wanted it to be formatted that way, looks a little weird but when I tried to change it the linter adds red lines.

ConorTighe1995:
> Looks good to go as it is from my end. Noticing
> Might be a good opportunity to handle the response from callStartProcess here.
> 
> ```js
>   // here is the server code
>   const command = new Command(db)
>   await command.startProcess(cmdRequestData)
> 
>   return success({ ok: 1 })
> ```
> 
> Would you or anyone else reading think its a good idea to (and I am just spitballing here), on any response except for ok === 1, still toggle the loader back off, and maybe return to the play button and either:
> 
> * display a loud error either on screen
> * or maybe better just print the whole response in a `console.warn`
> * or maybe even better yet, make use of the already provided `this.props.setProcessBuilderValidationErrors`?
> 
> That way, and the real time mqtt stuff like commands is definitely a culprit for this type of thing happening, if a client reaches out to us saying 'hey im starting the process but it doesn't seem to be working or the start button keeps showing back up' we at least have something to start with.

Makes sense but might complicate things a bit to go at it in this ticket so Ill get this merged but maybe its something we could try in a future sprint after the other stuff is done.

PR #764, opened 2020-05-15
Feature: Flow rate calibration for pumps

### LINKED ISSUE #747: Feature: RPM to flow/time unit conversion - add support to the Set Unit Modal
The "Set Unit" modal in the device setup view allows only setting default units for streams that do have
alternative units. RPM doesn't have any alternative units so far.

We need to add a special case to allow customers to set the default unit to "ml/min" for streams that have a default unit of RPM/rpm. This also requires an additional input field for this case which takes the "ml/rotation" value. 

Do not persist anything yet.

### LINKED ISSUE #749: Feature: RPM to flow/time unit conversion - persist conversion
Persist the user input in the device's unit configuration and make sure to update all unit conversion across the webapp to handle this special case of unit conversion when unit conversions are applied.

This entails: 
- FreeValueStream in the process builder
- OnOfStream in the process builder
- Set stream default modal
- Display of validation errors in the `ProcessValidator.js` utility

vdiezel:
Using rpm as a unit is often an inconvinient way of controlling pumps. We introduce a new feature that allows customers to calibrate their pumps to allow conversions between rpm and flow rate in ml/min. You can calibrate your pumps in the new device setup section 'Pump Calibration'. Once a pump has been calibrated, the flow rate unit ml/min will be unlocked for it and usable throughout the application with the exception of graphs. Graphs will keep displaying the speed of the pumps in rpm until the next firmware update. 

Bildschirmfoto von 2020-05-15 13-30-42

\private
closes #747 
closes #749 

Gives customers the option to calibrate pumps to get the option to use ml/min as units for all modules that have 'rpm' as their unit. Works like this: 

Once they have inserted the ratio for ml/rotation, that stream will allow chosing the ml/min unit from the unit dropdown. Once they chose ml/min as a unit, our default unit API actually will save it automatically as their default unit for this stream. All I had to do then was to check, if for the customers default unit (ml/min) there is a stored customConversion and if so, apply it instead of our normal unit conversion.

## Is every stream with the unit 'rpm' a pump?

No, but there is no other way for us to tell for a downstream if it is a pump or not (upstreams have the dataType 'pump speed', but downstreams don't have it. So customers could be calibrating a stream where the conversion between rpm and ml/min is nonsense, but it wouldn't break anything. We need more information from the device here.

## Is the data in the graph also convertable?

No, currently not. The upstream we have that send the fake data for the pumps will remain in rpm for the simple reason that the upstreams don't communicate to which downstream pump they belong. We have a `relatedDownstreams` field on upstreams but its not set on the pumps. I have created a request for the firmware people to fix this.

Once this is fixed, I will be able to connect those and pump upstreams will be automatically use their pump downstreams conversions.


src/utils/DeviceConfig/CustomConversions.js
```diff
@@ -38,6 +40,14 @@ class CustomConversions extends DeviceConfig {
     const customConversions = CustomConversions.getCustomConversions(device, streamId) || []
     return customConversions.filter(aConversion => aConversion.to !== unit)
   }
+
+  static convertWithRatio(value, { ratio }) {
+    return roundToTheNthDecimal(value * ratio, UI_CONSTANTS.NUMBER_INPUT_DECIMALS)
+  }
```
vdiezel, line 46: `convertWithRation` convert from the default unit to the desired customer unit (e.g. rpm -> ml/min). I apply rounding in this case to prevent displays like `1.2312412421321`. When converting the other way around in `revertWithRatio` we don't round to keep maximum precision (this value is stored).

src/components/device/modals/SetStreamDefaultModal.jsx
```diff
@@ -101,7 +107,7 @@ class SetStreamDefaultModal extends React.Component {
       ? this.getOneOfValue(streamDefault, targetStream)
       : this.getFloatValue(streamDefault, activeDevice, targetStreamId, targetStream)
 
-    if (!convertedDefaultTarget) return
+    if (convertedDefaultTarget === undefined) return
```
vdiezel, line 110: it can be the number 0!

src/utils/Stream.js
```diff
@@ -10,18 +9,13 @@ export default class Stream extends StreamIcons {
...
 
   static hasRelatedStreams(stream) {
     return stream.relatedDownstreams && stream.relatedDownstreams.length
   }
 
-  static getStreamOptions = (values, streamUnit, selectedUnit) => values.map(e => ({
-    value: e,
-    text: convertFrom(streamUnit).to(selectedUnit)(e),
-  }))
-
```
vdiezel, line 24: this is more a view layer function so I moved it out for now

src/components/processviewer/body/builder/Unit.jsx
```diff
@@ -18,8 +19,13 @@ export default class Unit extends Component {
...
-    const unitOptions = getUnitOptions(stream.unit).map(unit => ({ text: unit, value: unit }))
+    const { stream, activeDevice } = this.props
+    const standardOptions = getUnitOptions(stream.unit).map(unit => ({ text: unit, value: unit }))
+
+    const conversions = CustomConversions.getCustomConversions(activeDevice, stream.id)
+    const customOptions = conversions.map(({ to }) => ({ text: to, value: to }))
+
+    const unitOptions = [ ...standardOptions, ...customOptions ]
 
     const uniqueUnit = unitOptions.length === 1
```
ConorTighe1995, line 30: What if there are more? should this be `>= 1`

src/components/device/modals/PumpCalibrationModal/PumpCalibrationModal.jsx
```diff
@@ -0,0 +1,127 @@
...
+    const { ratio, selectedStreamId } = this.state
+
+    if (selectedStreamId === undefined) {
+      this.setError('No pump selected')
+      return
+    }
+
+    const parsedRatio = parseFloat(ratio)
+    if (isNaN(parsedRatio)) {
+      this.setError('Ration value has to be a number')
```
DanielSeehausen, line 67: should this be `ratio`?

src/components/device/modals/PumpCalibrationModal/PumpCalibrationModalForm.jsx
```diff
@@ -0,0 +1,48 @@
...
+}
+
+const renderRatioInput = (ratio, updateHandler) => (
+  <Form.Field>
+    <label>
+      ml/rotation
+    </label>
+    <Input
+      fluid
+      placeholder='enter ratio...'
```
DanielSeehausen, line 31: we go from `Select pump` with a capital to `enter ratio` with a lower case character at the start
if intentional 👍 , if unintentional, wanted to point it out

src/components/device/modals/SetStreamDefaultModal.jsx
DanielSeehausen: ncie catch - this guy gets me all the time

src/components/device/modals/SetStreamDefaultModal.jsx
```diff
@@ -145,19 +151,35 @@ class SetStreamDefaultModal extends React.Component {
...
+  getStreamValueOptions = configuredStream => {
+    const { activeDevice } = this.props
+    const { unit, expectedValueOptions, displayedUnit, id } = configuredStream
+
+    return expectedValueOptions.map(e => {
+      const customConversion = CustomConversions
+        .getCustomConversionsByTargetUnit(activeDevice, id, displayedUnit)
 
+      return {
+        value: e,
```
DanielSeehausen, line 163: maybe easiest to rename `e` to `value` so its easier to understand what is going on in the other lines it is being used?

unless `e` has some specific meaning imbedded in it here

src/components/device/setup/SectionWrapper/sectionDescriptions.js
```diff
@@ -22,6 +22,9 @@ export default {
     Every sensor on the device can be calibrated using the calibration
     menu present on the sensor's associated graph.
     This section provides an overview on current calibrations.</p>,
+  pumpCalibration: <p>
+    Calibrate pumps to get the advantage of using flow rate
```
DanielSeehausen, line 26:
```suggestion
    Calibrate pumps in order to use flow rate
```

src/components/device/setup/SectionWrapper/sectionDescriptions.js
```diff
@@ -22,6 +22,9 @@ export default {
...
+    instead of rpm values for pump control.</p>,
```
DanielSeehausen, line 27:
```suggestion
    instead of rpm values.</p>,
```

src/components/processviewer/body/builder/Unit.jsx
DanielSeehausen: this might be slightly unclear because its a boolean that is being assigned here.
while this isn't in the PR, leading boolean variables with `is` may help avoid confusion: `isUniqueUnit` 

@ConorTighe1995 if I am reading it right (just from the code here, not from understanding the full context elsewhere), this can only be true if there is only 1 value. If we got a true back for multiple values, then it would not be a 'unique unit'

src/utils/Streams.js
```diff
@@ -87,4 +87,8 @@ export default class Streams {
   static getGroupStreams(streams, groupName) {
     return streams.filter(stream => stream.groupName === groupName)
   }
+
+  static getPumpStreams(streams) {
+    return streams.filter(stream => Streams.isDownstream(stream) && stream.unit === 'rpm')
```
O-Susanoo, line 92: i feel like this is too specific is there a way around this

src/components/processviewer/body/builder/Group.jsx
```diff
@@ -101,6 +101,7 @@ const Group = ({
       updateStreamTarget={updateStreamTarget}
       getStreamDisplayedUnit={getStreamDisplayedUnit}
       changeDisplayedUnit={changeDisplayedUnit}
+      activeDevice={activeDevice}
```
O-Susanoo, line 104: was there any way around this , i tried too hard to make those components as clueless as possible , so giving them an activeDevice is pushing too much knowledge on them

src/components/device/modals/PumpCalibrationModal/PumpCalibrationModal.jsx
vdiezel: had this typo 1000 times. Will fix.

src/components/device/modals/SetStreamDefaultModal.jsx
```diff
@@ -145,19 +151,35 @@ class SetStreamDefaultModal extends React.Component {
...
-  renderOneOfStream = configuredStream => {
+  getStreamValueOptions = configuredStream => {
+    const { activeDevice } = this.props
+    const { unit, expectedValueOptions, displayedUnit, id } = configuredStream
+
+    return expectedValueOptions.map(e => {
+      const customConversion = CustomConversions
+        .getCustomConversionsByTargetUnit(activeDevice, id, displayedUnit)
 
+      return {
```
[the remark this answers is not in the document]
O-Susanoo: i guess it's complaining for a reason that's why i actually moved this to Stream Class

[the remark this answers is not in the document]
vdiezel: In this case we might simply add another utility for display concern in the future

vdiezel:
> Two questions and I am guessing the second depends on the first:
> 
> **1.** Should we wait for firmware to update pumps for this first? I think the visual of data over time in the graphs is probably 80% what makes these things real and conceivable for users. Following up on the PR comment, If they aren't displayable until firmware updates, should we wait?
> 
> **2.** Should this be a changelog? I think it may be worthy of one as it's a pretty big feature/addition!
> 
> If both of those are easily answerable, then its looking like a :shipit: from my end.
> I did not realize how huge of work this is. All of the unit stuff is doing a good job becoming some of the more complicated code in the front end. Nice work keeping it clean as it grows.

Re: 1. I think this feature has already value without the graph display. Using ml/min for inputs as already a great benefit and they WILL actually see ml/min values on the process configuration. So I want to ship this. 

Re:2. Yes, this should be a changelog!

src/utils/Streams.js
vdiezel: Not sure if I get you here: 
In my view it is actually too general to derive like that if a stream is a pump or not, because it includes some streams that have rpm as a unit but arent pumps. So we need more information from the firmware here. I mentioned this in the PR description. Currently there is no way around this but there is no real harm either yet.

src/components/device/modals/SetStreamDefaultModal.jsx
vdiezel: agree. done!

src/components/processviewer/body/builder/Group.jsx
vdiezel: Ill look into it!

src/components/processviewer/body/builder/Unit.jsx
ConorTighe1995: I get you I misread that when I initially looked at it.

PR #775, opened 2020-05-27
adds a lint-diff and a lint-diff-and-fix npm command

### LINKED ISSUE #453: make ci hook for checking added lines for eslint errors
make ci OR/AND git action hook for checking added lines for eslint errors

vdiezel:
closes #453 

adds a npm command to lint diffs and to lint and fix them automatically, too. 

I also created a git hook script for this here: 

https://github.com/vdiezel/dot-files/blob/master/git-hooks/lintDiff

Linting in the CI suite via codeclimate was not a viable option because there is no way to configure it to lint only the diff - so it lints always the whole project.


package.json
```diff
@@ -72,7 +72,9 @@
     "build": "react-scripts build",
     "test": "jest",
     "test-with-coverage": "jest -i --coverage",
-    "eject": "react-scripts eject"
+    "eject": "react-scripts eject",
+    "lint-diff": "LIST=$(git diff-index --name-only --diff-filter=d HEAD | grep .*\\.[jt]sx\\\\? | grep -v json); if [ \"$LIST\" ]; then eslint $LIST; fi",
```
ConorTighe1995, line 76: So does this show a visual of linting differences similar to something like `git add -p?`

vdiezel: yes, this is how it looks:
Bildschirmfoto von 2020-05-28 11-11-01

DanielSeehausen: very cool

vdiezel:
> this is awesome! using it now
> 
> only request is to version control this (maybe a top level `.git-hooks`)
> while its still required for the dev to copy pasta that into their `.git/hooks`, then its at least front of mind for them

I have added a new repo for this kind of code:
https://github.com/ospin-web-dev/dev-comfort-zone

PR #777, opened 2020-05-28
adds a CD step to store build artifacts in a new s3 Bucket

vdiezel:
This is in preparation of a disaster recovery that will be part of the smoke tests for deployment. We are going to store build artifacts in another bucket in a compressed way for documentation and saving storage + storing the latest build in an umcompressed way for easier recovery via the S3 bucket (no need to pull the latest working build, unzip and deploy it but simply synchronize two aws s3 buckets)


DanielSeehausen:
auto updated the branch

vdiezel:
> this looks awesome - glad this is on someone's mind
> disaster recovery is an important step for us!
> 
> someone once told me a mnemonic for the tar commands:
> 'compress zee files' (czf) said in a borderline offensive german accent
> 'extract zee files' (xzf)
> 
> perhaps you just know them by now, but I can't not say it in my head

this IS how I actually memerize it...
The german language has a nice word for a mnemonic: Eselsbrücke (https://de.wikipedia.org/wiki/Merkspruch)

vdiezel:
> auto updated the branch

Co-author inc... :P

PR #780, opened 2020-05-28
740 login with try demo

### LINKED ISSUE #740: R3: Log in with demo account and navigate to /devices on 'try demo' click
As a user I should be able to click the try demo button in the splash page and the application should POST to user/demo, fully authorize me with a new demo user, then navigate to /devices using the demo account without any errors.

BLOCKED BY: #739 
---

### Deliverables
As a user...
 - [ ] I can click the try button and fire a POST request, a loading Icon will appear in the button.
 - [ ] I can see that the url will change on success and the app will navigate to /devices.
 - [ ] I see a message communicating the error on fail.
 - [ ] I can not load demo.ospin-app.com/devices directly in incognito without getting redirected to the splash page

ConorTighe1995:
Here is the login for the try demo button to handle the response from the server and access the application using the session token.

closes #740


.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ version: 2.1
 defaults: &defaults
   working_directory: ~/REACTor
   docker:
-    - image: circleci/node:latest
+    - image: circleci/node:13.1.0
```
DanielSeehausen, line 6: `"13.8.0"` is the last one in the `13.x.x` I believe circle CI has (then it jumps to `14.x.x`)

If its all the same, could we use that one? If not or if there is some specific reason we are on `13.1` then forget this comment entirely.

src/components/processviewer/body/builder/ControllerModal.jsx
```diff
@@ -88,7 +88,6 @@ class ControllerModal extends React.Component {
         size='tiny'
         open={open}
         onClose={() => this.close()}
-        onClick={event => event.stopPropagation()}
```
vdiezel, line 91: why was this line removed? Its present on the master branch and I think this change will make inputs in the controller modal not selectable anymore

ConorTighe1995: Might have been lost in a merge or rebase, ill add it now.

PR #783, opened 2020-06-03
782 fix label offset

### LINKED ISSUE #782: Fix visual off set for virtual devices on demo
offseterror

### Current Behavior
 The virtual label causes the device state label to move out of place.

---

As a potential customer I expect the UI to be presented in a aesthetic way and for the device state to be centered.

ConorTighe1995:
Center aligns the label for virtual devices.

Before:
offseterror

After:

closes #782


DanielSeehausen:
the circle is likely going to change soon per Jan and Nina's request, so another potential reason it might find a better home outside of it 

some potential other options: as a watermark behind the connection text, or above or below the circle, or as Viktor suggested as a ribbon (which might be the easiest and quickest)

DanielSeehausen:
a way to avoid this circle ci issue may be to ensure the circle I file recognizes when it's on demo and does a modified job. then we can still have one circle ci file between them. 

all the same, fixing it once should get a commit on top that reverts it and we should be good thereafter

DanielSeehausen:
left a comment in slack re: (the ui part of) this potentially wanting to be in master as well

PR #793, opened 2020-06-04
removes processes field from the device data

### LINKED ISSUE #748: Check processes field in device object mapper is used
first - create a backend ticket to be completed **after** this ticket is completed (mark it as blocked by this ticket). The backend work should update update all references to device processes to find them through a processes (deviceId) query on an interface. 

in device object mapper it seem that we have a field called processes according to @vdiezel  this field should be deprecated , try to see if we can remove it 

And if you have time try to see if they are others fields with the same curse

--- 

this may be used here:
```
// LogViewQueryBar
   11   const userProcessesOnDevice = processes
   12     .filter(process => device.processes.includes(process.id))
   13     .map(process => ({ text: process.name, value: process.id, key: process.id }))   
```

vdiezel:
closes #748


src/utils/process.js
```diff
@@ -197,8 +197,6 @@ export class Process {
     if (Process.isFinished(process)) return PROCESS_STATES.finished
     if (!Process.hasBeenStarted(process)) return PROCESS_STATES.executable
     if (device && Process.isRunningOnDevice(process, device)) return device.state
-
-    console.error(`Process: ${process.id} can not derive its state`)
```
vdiezel, line 201: this one was logging out a lot - also in production when a single process was still loading.

src/components/log/LogViewQueryBar.jsx
```diff
@@ -9,8 +9,9 @@ const mapStateToProps = state => ({ processes: state.processes })
 const getProcessOptions = (device, processes) => {
 
   const userProcessesOnDevice = processes
-    .filter(process => device.processes.includes(process.id))
+    .filter(process => process.deviceId === device.id)
     .map(process => ({ text: process.name, value: process.id, key: process.id }))
+    .reverse()
```
vdiezel, line 14: reversed them so the latest process appears first in process dropdown for device log search

PR #804, opened 2020-06-05
663 fix dropdown state

### LINKED ISSUE #663: Zoom selection dropdown on graph sticks
### Current Behavior
 - as a user on the data viewer
  - I select zoom level, i get drop down
  - I click one of the options, it successfully changes modes, but the dropdown stays open (see image)

---

 - as a user, when I use the dropdown
 - [ ] it folds up on the first selection click
 - [ ] fix all other dropdowns if possible. If this turns out to be some huge lift then leave this part out and make a new ticket to do this

---

### Steps to Reproduce
 - first, click and get focus anywhere on page that isn't the dropwdown
 - click the dropdown
 - select an option
 - see the dropdown is sticky
 - select another option (it closes)
 - click anywhere else to get focus elsewhere to do again (if focus is on the dropdown, then it successfully goes away - that's why selecting twice works)

---

### Images/Screenshots/Links

image

ConorTighe1995:
These changes remove the error of the select dropdown staying open after a item was clicked. The issue was coming from the <Form.Select> component so I now set the toggling of the select state using a regular <Select /> component controlled by its parents state.

I decided against encapsulating it in a separate components as I don't think that would bring much value compared to just just using a regular <Select /> component with a <Label /> component, I check other parts of the application and couldnt see anywhere else were this problem was appearing.

closes #663


src/components/processviewer/body/dataviewer/Stream.jsx
```diff
@@ -41,12 +43,23 @@ export default class Stream extends React.Component {
     )
   }
 
+  openZoomDropDownHandler = () => {
+    const { openZoomDropDown } = this.state
+    this.setState({ openZoomDropDown: !openZoomDropDown })
+  }
```
vdiezel, line 49: There is a shorter way of doing this, using the `prevState` that is passed into `setState` as the first argument (I combined it with destructuring immediately)

`this.setState(({ openZoomDropDown }) => ({ openZoomDropDown: !openZoomDropDown }))`

ConorTighe1995: I seen that way before, I personally default to this way out of habit and that I think its easier to read.

PR #805, opened 2020-06-05
786 widget ui ux changes

### LINKED ISSUE #786: Widget UI/UX Changes
The following is condensed from a meeting with Jan and Nina on June 3rd:

They would like to make several small cosmetic changes to UX/UI on the widgets. They brought up some good points, and in general are happy with how we are using different colors/icons to communicate application state.

#### Process Widget

- remove the dot to the left of the state text. It is almost entirely redundant now that we have different icons and colors. It also has some changing behavior -- sometimes solid, sometimes not (see green dot below) which doesn't add much to communicating information about the process
- change process widget icon to be 'edit' instead of 'rename' (as it also changes the comment). Or, any other name that you think most communicative!
- update the finished process duration time to match the running time duration text (see bottom image). The larger text of the two (which is currently the running process duration text) was the preference. 
  - This might end up a little confusing when the finished process and running process are displaying seemingly the same thing. To keep them differentiated, one thing we can perhaps do is **darken the running process duration timer compared to how it is currently** and **slightly lighten the complete process duration time compared to how the running process duration timer is now**. Something like this might help keep the immediate recognition that it is actively timing instead of a fixed amount. If you have other ideas on how to differentiate those two timers once they have the same styling, I would say go for it and snap some screenshots and post in channel. The difficulty I see is helping the user immediate recognize 'i see that one of these timers is actively running, and the others are complete durations'. 
  - The running duration is slightly not aligned. It should be aligned on the left with the text above it

---

#### Device Widget

- (no action for now on the device widget. Jan and Nina want to investigate some ideas for another image asset that is large enough to show an LED ring inside. What was most important was that the LED ring we show on the device widget matches the actual Device. Until they investigate other options, this work is on stand by)
 
### Reference Images

image

image

image

vdiezel:
closes #786 

as always, I snuck in a couple of minor fixes. Main ones are here:

duration_small

process widget:
- removed circle icon in the bottom panel of the process widget
- removed colouring of status there, too
- removed the 'comment' from the tile and showing it on hovering the info icon in the icon bar (the part I'm the least happy with because unlike the other icons there, this icon is not intended to be clicked, but couldn't find a better solution to this
- modified duration / started / finished / created time to be normal font instead of bold
- modified date display to display year and changed the format to be less american :P (got confused everytime and I hope I wasn't the only one)
- duration display shows now running seconds when process is running
- process name is restricted to a single line - when longer, text ellipses. Full name shown on hover

Processes View: 
- removed the Add Process Tile to reduce noise, instead moved out the "Create Process" button from the table view (more real estate there now) one layer up. This way the creation of a process is not depending on the selected view anymore and we have less duplication of functionality.
- to emphasize the "Create Process" button I moved the view type (Tiles/Table) to the right
- Jan liked the "add process" tile, he might kick my ass - but I have arguments!
- show calendar -> Show Calendar to match casing of our other buttons

Before:
Bildschirmfoto von 2020-06-05 18-40-31
Bildschirmfoto von 2020-06-05 18-40-38
After: 
Bildschirmfoto von 2020-06-05 18-40-13
Bildschirmfoto von 2020-06-05 18-40-22

Editable Header: 
- removed the circle around the edit button to match the icons in the process widget

PhaseTabContent: 
- removed the circle around the "Delete" Button to match the other icons (also made them bigger to compensate for the smaller click area)

Before:
Bildschirmfoto von 2020-06-05 18-48-39
After:
Bildschirmfoto von 2020-06-05 18-48-48

DeviceMenu: 
- removed double shadows from the right side device menu
- increased the clickable area for the hamburger icon when the menu collapses 

Before:
Bildschirmfoto von 2020-06-05 18-38-10

After: 
Bildschirmfoto von 2020-06-05 18-38-16


src/components/utility/EditableHeader.jsx
```diff
@@ -93,11 +93,11 @@ class EditableHeader extends React.Component {
               </Button>
               <Button
                 className={`${classes.editButton} ${EditableHeaderButtonClassName}`}
```
ConorTighe1995, line 95: Is this meant to be one or the other? looks like `EditableHeaderButtonClassName` should have been replaced here, or if this is meant to be for toggling the styles of classes.editButton maybe it should be renamed.

src/components/device/processes/ProcessTile.jsx
```diff
@@ -49,7 +48,7 @@ const DISPLAY_MAP = {
 const continuouslyUpdateRunningDuration = (startedAt, setRunningDuration) => (
   setInterval(() => {
     setRunningDuration(Date.now() - startedAt)
-  }, 15000)
+  }, 1000)
```
ConorTighe1995, line 51: What do you think about moving this code into Time.js so we have one source for the timer across the application?

https://github.com/ospin-web-dev/REACTor/blob/master/src/utils/Time.js

DanielSeehausen:
This all looks awesome. I love the changes. I have one note and one warning re: another design opinion Jan had:

> * removed the 'comment' from the tile and showing it on hovering the info icon in the icon bar (the part I'm the least happy with because unlike the other icons there, this icon is not intended to be clicked, but couldn't find a better solution to this

I like the change - I certainly think its better than before now. Currently, the edit button changes the title and the comment. Now that we have an icon that pertains to the comment, we could consider repurposing that icon to be clickable (in addition to hoverable), which pops up a modal to edit the comment. I mention this because it might seem off that one icon shows the comment, and another icon edits the comment. What do you think?

> * modified date display to display year and changed the format to be less american :P (got confused everytime and I hope I wasn't the only one)

Surely this was me who set the date this way, but it wouldn't have been intentional - I am pro deutsch date format everywhere

> * process name is restricted to a single line - when longer, text ellipses. Full name shown on hover

I brought exactly this up in the meeting last week. I mentioned we should ellipse that off hand, and Jan seemed to believe we should not, as people are embedding important identifying information in the title. That being said, I still think ellipsing is the right choice here. If people need more general info in this view, maybe a different solution than encouraging them to put it in the title would be to expand that new comment icon on hover tooltip to include more information (like a larger card with, full name, comment, phase names, users, etc. [ just spitballing ])

All the same, Jan might have a comment on that, but I am sure we can convince him otherwise on the ellipsing being a good thing. I would consider pinging him when you take this in and showing him the assets (gifs/images) and arguments you have made here already for why it makes sense. The standardization of the process widget sizes alone is a major win in my mind. 

> * removed the Add Process Tile to reduce noise, instead moved out the "Create Process" button from the table view (more real estate there now) one layer up. This way the creation of a process is not depending on the selected view anymore and we have less duplication of functionality.
> * to emphasize the "Create Process" button I moved the view type (Tiles/Table) to the right
> * Jan liked the "add process" tile, he might kick my ass - but I have arguments!
This is nicer now I agree. The consequence for me (as a viewer) is that the upper part of the screen feels washed out in semantic buttons and rows and everything. That being said, there is stuff we could do to reduce the clutter and noise of the other stuff (e.g. we could move the show calendar callback into a 'by Date' on the dropdown, or make it just an icon), and then that point about clutter would no longer be valid.

> * removed double shadows from the right side device menu
> * increased the clickable area for the hamburger icon when the menu collapses
Doing god's work god bless 🙏

vdiezel:
@DanielSeehausen 
I agree on the thing with the comment. had the same idea - but didn't want to tackle it in this PR. Regarding the name I also agree, I just don't know what the best solution there is to keep the tiles the same size. We could reserve enough space for two lines in every tile OR make the tiles longer

src/components/device/processes/ProcessTile.jsx
vdiezel: I think this timer is the update timer only for this specific component - so for now I'll keep it here!

src/components/utility/EditableHeader.jsx
vdiezel: Don't know if I understand you here correctly: 
You mean its `editable` and `edit` ? `editable` refers to `EditableHeader` which is the parent of the button

PR #808, opened 2020-06-08
close config dropdown on blur

ConorTighe1995:
The select would get stuck open on dev when only rpm was a option so I added a on blur function to make sure it closes when focus is lost


src/components/processviewer/body/dataviewer/Stream.jsx
```diff
@@ -53,6 +53,11 @@ export default class Stream extends React.Component {
     this.setState({ openUnitDropDown: !openUnitDropDown })
   }
 
+  closeDropDown = () => {
+    this.setState({ openUnitDropDown: false, openZoomDropDown: false })
+    console.log(this.state.openUnitDropDown, this.state.openZoomDropDown)
```
DanielSeehausen, line 58: this snuck by!

ConorTighe1995: Ahh yes ill fix this well spotted :100:

PR #813, opened 2020-06-09
789 outsource nexus as npm package

### LINKED ISSUE #789: Host nexus on a package sharing platform like npm
we want to start reusing code in different repos. For starters, the nexus collection should be outsourced into a seperate repo and hosted on npm (probably), so we can reuse it in REACTor and in the newly planned Admi/Support web app. 

- [x] create an organization on npm
- [x] create a namespace (@ospin)
- [x] create a new repo for nexus
- [x] update the code in the package so it works as an npm package
- [x] document the publishing workflow for the package
- [x] replace nexus directory in REACTor with the new package

vdiezel:
moves nexus into an npm package. 
closes #789


tsconfig.json
```diff
@@ -18,10 +18,11 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
-    "jsx": "preserve"
+    "jsx": "preserve",
+    "noImplicitAny": false,
```
vdiezel, line 22: had to add this to prevent creating types of nexus

src/redux/thunks/process/callDownloadStreamData.js
```diff
@@ -1,9 +1,12 @@
...
 import thunkWrapper from 'redux/thunks/thunkWrapper'
 
-const callDownloadStreamData= async (dispatch, streamId, streamName, processId) => {
+const callDownloadStreamData = async (dispatch, streamId, streamName, processId) => {
   const { url } = await nexus.datapoints.requestStreamData({ streamId, streamName, processId })
-  await nexus.datapoints.downloadStreamData(url)
+  const a = global.document.createElement('a')
+  a.href = url
+  a.setAttribute('download', 'download')
+  a.click()
```
vdiezel, line 9: this code was orinally in nexus, but it doesn't belong there as this is pure frontend

src/components/device/widget/DeviceWidget.jsx
```diff
@@ -59,28 +58,27 @@ const handleClick = (event, history, dispatchResetActiveDevice, deviceId) => {
...
-  })
-)
+  return {
+    name,
+    hasAccess,
+    startedBy: user.userName,
+    startedAtDay: Time.getDisplayDay(startedAt),
+    startedAtTime: Time.getDisplayTime(startedAt),
+    duration: (Date.now() - startedAt),
+    startedAtSinceEpoch: startedAt,
```
vdiezel, line 73: `async/await` was easier to mock

DanielSeehausen: async await all the way! never look back! (except when lambda makes us)

src/redux/thunks/process/callDownloadProcessData.js
```diff
@@ -9,7 +9,10 @@ const callDownloadProcessData= async (dispatch, processId, upStreams) => {
       name: stream.name,
     })),
   })
-  await nexus.datapoints.downloadStreamData(url)
+  const a = global.document.createElement('a')
+  a.href = url
+  a.setAttribute('download', 'download')
+  a.click()
```
DanielSeehausen, line 15: good catch

test/components/devices/DeviceList.test.jsx
```diff
@@ -14,10 +15,19 @@ import {
   getRunningDevice,
 } from '../../mocks/deviceDataMock'
 
+nexus.process.getPublic = jest.fn(
+  () => new Promise(res => res({ startedBy: '666', name: 'Purgatory Chicken Wings', startedAt: Date.now() })),
```
DanielSeehausen, line 19: sign of the devil black phillip

image

test/components/devices/DeviceList.test.jsx
```diff
@@ -14,10 +15,19 @@ import {
...
+)
+nexus.user.getPublic = jest.fn(() => new Promise(res => res({ userName: 'Der Geist, der stets verneint' })))
```
DanielSeehausen, line 21: omgsh I actually saw mephistopheles in the met opera once this was the dude it was epic he was an absolute unit
image

test/components/devices/DeviceList.test.jsx
```diff
@@ -51,30 +61,36 @@ describe('DeviceList.jsx', () => {
 
     test('loads a single offline device', async () => {
       const device = getOfflineDevice()
-      assertDeviceGeneratesText(device, 'Offline')
+      await act(() => assertDeviceGeneratesText(device, 'Offline'))
```
DanielSeehausen, line 64: til `act`

DanielSeehausen: I believe the way act is being used here is covered by react testing library (which they reference here:
https://reactjs.org/docs/testing-recipes.html#act)

Would `findBy` provide what you are looking for?
https://testing-library.com/docs/react-testing-library/cheatsheet

test/components/devices/DeviceList.test.jsx
```diff
@@ -51,30 +61,36 @@ describe('DeviceList.jsx', () => {
...
     })
 
     test('loads all the state of a devices', async () => {
       const devices = [getRunningDevice(), getPausedDevice(), getOnlineDevice(), getOfflineDevice()]
-      const { getByText } = renderDeviceList(devices)
-      const expectedTexts = ['Running', 'Paused', 'Online', 'Offline']
+      // TODO temporary fix to prevent logging the error to wrap the call in act
+      // https://reactjs.org/docs/test-utils.html#act
+      // somehow line 89 doesn't fix it
+      console.error = jest.fn()
```
DanielSeehausen, line 87: it might be prudent to [mockRestore](https://jestjs.io/docs/en/mock-function-api#mockfnmockrestore) to stay on the save side here

DanielSeehausen: (at end of this test?)

test/components/devices/DeviceList.test.jsx
vdiezel: had to google it. Have to see that movie, now.

test/components/devices/DeviceList.test.jsx
vdiezel: I had to use `act`, because it was throwing error messages via console.error. It was still passing the test, but the error message told me on crimson red font, that I had to wrap any state updates of the component into act. The state updates where caused by the `useEffect` hooks, which aren't easy to mock.

test/components/devices/DeviceList.test.jsx
vdiezel: I clear all mocks in the afterEach function

PR #818, opened 2020-06-11
814 message bug

### LINKED ISSUE #814: Reset code message bug
Ive found a small bug when sending many reset codes. I think only one of these messages should appear at a time, both appearing is condescending and confusing to the user.

Only show success or error, have them overwrite each other.

bug-reset-message

ConorTighe1995:
Both an error and success message can no longer be displayed as a response when attempting to reset the password. Now its only one or the other.

closes #814


ConorTighe1995:
I think there is a way to do this without the render function so I changed this to WIP for now.

ConorTighe1995:
> I think there is a way to do this without the render function so I changed this to WIP for now.

I refactored it, actually required very little changes in the end just some linting changes and a state change.

src/components/authentication/ResetPassword.jsx
```diff
@@ -34,36 +34,10 @@ class ResetPassword extends React.Component {
       isError: true,
       errorMessage: message,
       pending: false,
+      isSuccess: false,
```
ConorTighe1995, line 37: Where the magic happens.

PR #820, opened 2020-06-15
Get nexus back into REACTor

vdiezel:


ConorTighe1995:
Can you let me know when this is merged so I can add a call into nexus for the device support route

PR #821, opened 2020-06-16
adds update user

vdiezel:


vdiezel:
> I take it this was forgotten about?

Daniel added this to the nexus package but I reverted the changes before it got merged, so i had to catch up

PR #824, opened 2020-06-17
807 support table

### LINKED ISSUE #807: support users get second table/section in device access
### Feature

As discussed in sprint planning:

We would like to make a second table in the device access view (see image A below for the view in question) that displays support persons. If there are no support people on a device, no support person table/header should display. 

### Deliverables

**Following is a working idea of what this could look like:**

- [ ] Only users that are in group 'support' on the device will show up in the table
- [ ] Make an additional table below the existing one, replacing the following header field:
  - [ ] Username -> Contact Name (use the `fullName` field on the user. If `!fullName` or `fullName === ''`, use the `userName` value instead
  - [ ] Remove `Group` and replace with `Number` -- use the `phoneNumber` field. If no number is present, provide a 'Number not yet provided' or some other obvious text that the user needs to update their account
- [ ] The table should have a header of some type ('Device Support'?)
- [ ] Try a semi-transparent OSPIN blue for the background of the table/header to differentiate it

**Following are misc additional changes that might be simplest to put into this ticket:**

- [ ] Consider updating the 'Add User' button to be 'Add Device User'
- [ ] See the user group level text tooltips when you click on 'Add User' and hover over the groups. This should be updated to pertain to a device, not a process, using the language 'all device processes' in place of 'process'. For the admin tooltip, please update:
> Admins can view, delete, execute, and edit all device processes. With the exception of the device owner, admins can edit the permission levels of other device users.
- [ ] The tooltips should also display when a user's group is being edited (see image B below)
- (ignore the 'admin tools' bar for now - that will go into the new app soon)

#### image A
image

#### image B
image
(the tooltip should also display here)

ConorTighe1995:
This table will list off the support users on a device, there is a button that allows you to copy the email and a button that opens the email client.

The email client is a bit dodgy on the Ubuntu email client, If you guys could test it out and see how it works on your client and let me know, if its not consistent on other clients ill try improve it or remove it depending on your feedback.

The hambda ticket for support users will be merged first, but If that goes in and this is ok I can make a change log for customers.(Need to test the ability to add support users too!)

closes #807


ConorTighe1995:
Just doing some updates and adding tests to this ticket.

ConorTighe1995:
> @@           Coverage Diff           @@
> ##           master     #824   +/-   ##
> =======================================
>   Coverage   45.77%   45.78%           
> =======================================
>   Files         294      296    +2     
>   Lines        5627     5648   +21     
>   Branches      917      918    +1     
> =======================================
> + Hits         2576     2586   +10     
> - Misses       3051     3062   +11     

Balance has been restored to the universe :yin_yang:

src/components/device/manageaccess/SupportUser.jsx
```diff
@@ -0,0 +1,39 @@
...
+    dummy.select()
+    document.execCommand('copy')
+    document.body.removeChild(dummy)
+    FlashMessenger.success(`copied ${email}`)
+  }
+
+  return (
+    <Table.Row>
+      <Table.Cell>
+        {userName}
```
vdiezel, line 21: I think we should display the fullname if there is one set on the user and if not, then we default to the user name

src/components/device/manageaccess/SupportTable.jsx
```diff
@@ -0,0 +1,37 @@
...
+
+  return (
+    <>
+      <Header as='h1'>Device Support</Header>
+      <Table style={{ borderTop: '.2em solid #2878C2' }} celled selectable padded sortable>
+        <Table.Header fullWidth>
+          <Table.Row>
+            <Table.HeaderCell>Username</Table.HeaderCell>
+            <Table.HeaderCell>Email</Table.HeaderCell>
+            <Table.HeaderCell>Number</Table.HeaderCell>
```
vdiezel, line 23: I think `Phone` or `Phone Number` would be a little bit more descriptive  - because when none is set, people might wonder what kind of number it could be (at least for people who aren't english natives)

test/helpers/UserGenerator.js
```diff
@@ -19,4 +19,19 @@ export default class UserGenerator {
...
+      userName: faker.internet.userName(),
+      fullName: faker.name.findName(),
+      organization: 'ospin',
+      email: faker.internet.email(),
+      role: 'support',
+      phoneNumber: '065645646456',
+      ...preMapData,
+    }
+    return { ...createUserObject(userData), ...postMapData }
+  }
```
vdiezel, line 36: to keep code DRY, this function isn't really necessary, ou can simply use the existing `generate` function:
```js
generate({ role: 'support' })
```

src/components/device/manageaccess/UserTable.jsx
```diff
@@ -48,13 +48,15 @@ const UserTable = ({
...
       </Table.Header>
       <Table.Body>
-        {deviceUsers.map(deviceUser => <User
-          activeDevice={activeDevice}
-          key={deviceUser.id}
-          user={deviceUser}
-          me={me}
-          modifyAccessHandler={modifyAccess}
-        />)}
+        {deviceUsers.map(deviceUser => (
```
vdiezel, line 51: The deviceUsers are still containing the support users here, but they should be filtered out

src/components/device/manageaccess/SupportUser.jsx
ConorTighe1995: I had it to set as full name and changed it for 2 reasons, 1. some of the fullName values were coming in blank and username is more consistent it seems as its required on signup. 2. Full name seems to be very personal data to just display to everyone using a device? first name would make more sense but we dont have that as far as I see in the payload so username seemed like the better option.

ConorTighe1995: People working in support are told not to give a full name over the phone, I dont think it would be any different here.

src/components/device/manageaccess/SupportTable.jsx
ConorTighe1995: I agree ill update that now

ConorTighe1995:
the new look

ConorTighe1995:
with user table fixed

src/components/device/manageaccess/SupportUser.jsx
vdiezel: Convinced!

src/styles/semanticoverrides.css
```diff
@@ -3,6 +3,10 @@ body {
   background-color: #f5f5f5;
 }
 
+.ui.inverted.primary.button{
+  box-shadow: 0 0 0 2px #2878C2 inset !important;
+}
+
```
vdiezel, line 9: When you define this class like this, it will overwrite the default settings of all inverted primary buttons
I think it will be a better practice to give a specific class name to the button where you want your changes to be applied

ConorTighe1995: Yea I know but the reason I did it this way was because className would not effect this element and when we were talking you were saying that it should be the action blue we were using so it made sense to apply that to all inverted primary buttons.

If I could do it with a className I would but I spent a good 30 mins on it and couldnt seem to get the boxShadow to change.

ConorTighe1995: Ill have another look when I have the time but I still think you might as well override all inverted primary button colors as we only use that color blue in the app for actions anyway.

PR #827, opened 2020-06-19
826 support hotfix

### LINKED ISSUE #826: Fix support user offsets
Fix the following:
- Support user in user access dropdown should not be displayed.
- You cant add a support user 
- fix copy button offset.

ConorTighe1995:
closes #826 

This takes support out of the list and out off the options for a user to be added to a device user as support.

this also stop the rendering of the copy button if no email is supplied as its now pointless and will not be in the usual position.


src/components/device/manageaccess/SupportUser.jsx
```diff
@@ -0,0 +1,49 @@
...
+  return (
+    <Table.Row>
+      <Table.Cell>
+        {userName}
+      </Table.Cell>
+      <Table.Cell>
+        {email !== null ? (
+          <>
+            {email}
+            <div style={{ position: 'absolute', bottom: '15px', marginLeft: '35vw' }}>
```
ConorTighe1995, line 27: I get around the offset issue by not rendering a button at all if no email is supplied.

src/styles/semanticoverrides.css
```diff
@@ -3,6 +3,10 @@ body {
   background-color: #f5f5f5;
 }
 
+.ui.inverted.primary.button{
+  box-shadow: 0 0 0 2px #2878C2 inset !important;
```
ConorTighe1995, line 7: This overrides all inverted primary buttons to use the same blue as the normal primary buttons.

PR #829, opened 2020-06-22
select all selects only current page; resets when changing the page

### LINKED ISSUE #828: bug: select all in table view
### Description
In the table view of processes
`select all` selects every process, not just the currently visible processes
This is very problematic if someone tries to `select all`, then run a command like `delete`!

---

### Deliverable(s)
Select all should only select those processes which are currently in view on the page. Not processes that span the page select
This is what is meant by the page select:
image

---

### Additional context
image

vdiezel:
closes #828 

Now, when a user clicks "Select All" in the process table view, it selects only the current page. Also, when a user switches the page, the selection is reset.

Peek 2020-06-22 09-55


src/components/device/processes/ProcessesTable.jsx
```diff
@@ -30,9 +30,10 @@ class ProcessesTable extends React.Component {
 
   handleSelectAll = () => {
     const { allSelected } = this.state
-    const { filteredProcesses } = this.props
+    const { filteredProcesses, activePage, resultsPerPage } = this.props
     const selection = !allSelected
       ? [...filteredProcesses
+        .slice(activePage * resultsPerPage - resultsPerPage, activePage * resultsPerPage)
```
ConorTighe1995, line 36: :1st_place_medal:

PR #831, opened 2020-06-22
adds fallback cases for unrecognizable process state

DanielSeehausen:
second time we are blowing up on this (later on it can't resolve its image), figure its better to be 'least case bad wrong' than blow up for a quick solution


DanielSeehausen:
> I gave you a quick review to get this merged and get fireware back on track, can I ask what is actually happening though? From the code it looks like we were getting into an obscure state that was not triggering any `if statement` and no `else` means we would just hang?
> 
> Would be worth creating a method with a switch statement to act as a state machine and then fire a alert/log if this happens in the future on default?

you identified it right - this is the equivalent of a 'default' in a switch
we shouldn't ever hit any condition that doesn't match there, and if we do, we probably want some way to not fail later on. If we are getting in to that bad state it would be best we realize/catch that somewhere further upstream. Since this method is just for the view layer, I wanted to ensure we can still gracefully display something to the user even if we don't know what's up with a process.

It would be better to fire a log or something, I agree...maybe we should discuss a standard logging channel we can use from the front end. I don't know if we want to send it to the actual process log, because we don't want users seeing that stuff. We can discuss at lunch!

PR #841, opened 2020-06-29
839 phase group interface ii

### LINKED ISSUE #839: create process interface to manage phase groups PART2
The interface for handling phase groups needs to be extended to we get the following functionalities: 

Moving a phase out of a group
Moving a phase into a group
Moving a phase within a group
Moving a phase from one group into another

vdiezel:
closes #839 

huge test coverage - should help you to understand what is going on.

The idea behind the drag and drop logic is to always use the real indices of the phases under the hood and apply the correct move command based on where from and where to the phases are dragged + whether the dragged phase is part of a group or not 
 
For the assigning of the phase indices and group indices in the process tree we can pair @ConorTighe1995 once you are there.  

90% of code is test cvoverage because I have zero trust in this.


test/redux/reducers/process/movePhase.test.js
```diff
@@ -0,0 +1,227 @@
+import movePhase from 'redux/reducers/process/movePhase'
```
vdiezel, line 1: this tests only check if we have chosen the right move command - the logic of the move commands in tested in the `PhaseGroupHandler`

src/components/processviewer/phasesdisplaybar/PhasesNavbarWrapper.jsx
```diff
@@ -33,8 +33,8 @@ const mapDispatchToProps = dispatch => ({
   dispatchSetPhaseGroupIterationsCount: (processId, groupName, count) => dispatch(setPhaseGroupIterationsCount(
     { processId, groupName, count },
   )),
-  dispatchMoveProcessPhase: (processId, src, dest) => dispatch(moveProcessPhase(
-    { processId, src, dest },
+  dispatchMoveProcessPhase: (processId, src, dest, srcArea, destArea, sourcePhaseInGroup) => dispatch(moveProcessPhase(
```
DanielSeehausen, line 36: that `sourcePhaseInGroup` is being passed jumps out, because that is something that can/perhaps should be derived from the other variable `src` via a phase/process/processDescription utility function. At least in this file, it is derived right before it is passed. (I may be missing a use case where it can't always be derived by the information that is already passed)

src/utils/process/PhaseGroupHandler.js
```diff
@@ -122,6 +126,167 @@ class PhaseGroupHandler {
...
+      PhaseGroupHandler.setGroup(process, newPhase.phaseId, groupName)
+    })
+  }
+
+  static movePhaseToTheStart(process, src, dest, groupName, sourcePhase) {
+    const insertionPoints = PhaseGroupParser
+      .getIterationEndPhaseIds(process, groupName).slice(0, -1)
+
+    if (dest - src !== 1) {
+      // we don't need to move anything when the phase to be added is right before the group
```
DanielSeehausen, line 144: consider naming it to imbed the comment:
`const phaseBeforeGroup = dest - src !== 1`

src/utils/process/PhaseGroupHandler.js
```diff
@@ -122,6 +126,167 @@ class PhaseGroupHandler {
...
+    PhaseGroupHandler.setGroup(process, sourcePhase.id, groupName)
+    PhaseGroupHandler.propagatePhaseIterations(process, groupName, sourcePhase.id, insertionPoints)
+  }
+
+  static movePhaseToTheEnd(process, src, dest, groupName, sourcePhase) {
+    const insertionPoints = PhaseGroupParser.getIterationEndPhaseIds(process, groupName).slice(1)
+
+    if (src - dest !== 1) {
+      // we don't need to move anything when the phase to be added is right
+      // after the group (and the group has one iterartion)
```
DanielSeehausen, line 157: same comment re: naming opaque logic in variables

src/utils/process/PhaseGroupHandler.js
```diff
@@ -122,6 +126,167 @@ class PhaseGroupHandler {
...
+    const phaseSequence = PhaseGroupParser.getPhasesSequence(process)
+    const iterationData = PhaseGroupParser.getIterationData(process, groupName)
+
+    const destPhase = phaseSequence[dest]
+    const sourcePhase = phaseSequence[src]
+    const positionOfInsertion = iterationData[0].phaseIds
+      .findIndex(aPhaseId => aPhaseId === destPhase?.id)
+
+    if (positionOfInsertion === 0) {
+      PhaseGroupHandler.movePhaseToTheStart(process, src, dest, groupName, sourcePhase)
```
DanielSeehausen, line 188: these methods are $$$$

src/utils/process/PhaseGroupHandler.js
```diff
@@ -122,6 +126,167 @@ class PhaseGroupHandler {
...
+
+    if (preGroupPhaseId) {
+      setPhaseProps(process, preGroupPhaseId, { next: prevLastGroupPhaseNext })
+    }
+
+    if (destPhase) {
+      setPhaseProps(process, lastGroupPhaseId, { next: destPhase.id })
+    } else {
+      setPhaseProps(process, lastGroupPhaseId, { next: -1 })
+    }
```
DanielSeehausen, line 280: because there is too little to nitpick in this PR, let me go extreme...

also expressed:
```js
const next = destPhase ? destPhase.id : -1
setPhaseProps(process, lastGroupPhaseId, { next })
```

test/redux/reducers/process/movePhase.test.js
```diff
@@ -0,0 +1,227 @@
...
+    movePhase(initState, action)
+
+    expect(spy).toHaveBeenCalledTimes(1)
+  })
+
+  it('calls PhaseGroupHandler.moveGroup when the dragged phase is within a group and it is dragged within the process tree', () => {
+    const processId = uuidv4()
+    const process = {
+      description: {
+        ...PG.generateProcessPhase(0, { next: 1, name: 'Phase name', groupnName: 'Group Tekkan' }),
```
DanielSeehausen, line 47:
```suggestion
        ...PG.generateProcessPhase(0, { next: 1, name: 'Phase name', groupName: 'Group Tekkan' }),
```

test/redux/reducers/process/movePhase.test.js
```diff
@@ -0,0 +1,227 @@
...
+      },
+    }
+
+    const spy = jest.spyOn(Process, 'movePhase')
+    movePhase(initState, action)
+
+    expect(spy).toHaveBeenCalledTimes(1)
+  })
+
+  it('calls PhaseGroupHandler.moveGroup when the dragged phase is within a group and it is dragged within the process tree', () => {
```
DanielSeehausen, line 43: I ran the tests to read the descriptions first to understand what was going on and this confused me a little:

is this to move a whole group around?
the description might be misleading, because it reads like a single phase is being moved out of a group in to the top level process tree, and the method being used is `moveGroup`, which seems to imply a whole group is being moved within the process tree in relation to other groups/phases around it

DanielSeehausen: (granted I might be missing the point all together)

test/redux/reducers/process/movePhase.test.js
DanielSeehausen: this typo repeats throughout the test file from copy pasta I am guessing

test/utils/process/PhaseGroupHandler.test.js
```diff
@@ -62,6 +62,27 @@ describe('createGroup', () => {
   })
```
DanielSeehausen, line 62: this file diff was so big github didn't want to show it unless clicked

DanielSeehausen: image
HUGE

test/utils/process/PhaseGroupParser.test.js
```diff
@@ -85,3 +93,10 @@ describe('getGroupIterationsCount', () => {
     expect(res).toEqual(2)
   })
 })
+
+describe('getPhaseIterations', () => {
+  it('returns the iteration number for a given phase', () => {
```
DanielSeehausen, line 98: copy pasta from above i think - needs description update

src/components/processviewer/phasesdisplaybar/PhasesNavbarWrapper.jsx
vdiezel: I decided that it is too costly to do that in the reducer: I would need to find the correct process from application state, parse out the phases, find the correct one by index and then check if it has a group set. WIth the current solution, most of this work is done anyway in the component, thats why I prefer this solution

src/utils/process/PhaseGroupHandler.js
vdiezel: done!

src/utils/process/PhaseGroupHandler.js
vdiezel: much better

src/utils/process/PhaseGroupHandler.js
vdiezel: done!

test/redux/reducers/process/movePhase.test.js
vdiezel: good catch!

test/redux/reducers/process/movePhase.test.js
vdiezel: Ill try to make this better. It is about moving a group, but the API will just digest only the first phase and then figure out itself if it has to move a whole group or just the phase (depending on the`srcArea` and `destArea` and `sourcePhaseInGroup`)

test/utils/process/PhaseGroupParser.test.js
vdiezel: :gun:

PR #842, opened 2020-06-29
WIP 833 group widgets

### LINKED ISSUE #833: Groups widget for phase tree
The following will be a widget that will represent groups of phases in the process tree, the widget will hold phase widgets inside. The ability to add new phases to the group using a similar button to the one at the bottom of the process tree will be added to the widget.

DOD:
- Group widget is drag able.
- Can drag phase widgets in and out of the group.
- Click a button and a copy of the last phase in the group will be added.

ConorTighe1995:
This the widget and group action buttons for the GUI. Since there are no groups in the backed yet I am going to merge this once the tests are finished then I will create a new ticket for connecting the interface to the GUI widgets and testing the group feature functionality.

This is the chain button for the creation of groups:
chainbutton

This is the group widgets(Ive changed the button since)
addGroup

This is the edit/set group iterations button(the processes are empty as I was testing something as I took this gif).
editgroup

closes #833


ConorTighe1995:
There is still some small changes to be done and tests to be written but once this has test coverage + gui is finished we can merge as it will be backwards compatible with pure phases and I can focus on the interfaces so we can introduce groups to production.

vdiezel:
Since this is a WIP, ill just comment for now: 

In general I like the way groups looks compared to normal phases. I have a couple questions/requests for the visuals, mainly to keep things behaving the same way as in the rest of the application and keeping the overview: 

1. I believe the button to add a phase within a group should be the same as at the buttom of the process tree - it does the same thing and I think it contributes to the clarity. 

2. I think we don't need scrolling within the groups, because it make it difficult to understand the full process sequence at one glance. Just fully exapand the groups - thats totally fine.  The add button should also be just aligned with the normal phases within the group (currently its one layer up probably due to the fact that we have scrolling)

3. For the iterations, I really like the popup but I'm afraid it is also required for the customer to immediately see iterations of every group in the process tree, so we shouldn't hide it in a popup. I'm also not clear about what `Phases` mean in the popup - is that the amount of phases in the group? If so, I think this information is not really required because they can see the phases in the group (even better when we get rid of the scrolling) and the phase count itself is of no real value. With the three icons in the group name header it also looks a bit dense - maybe have a new row below the group name with an input field that shows current iterations and allows entering a number (just like we do for the downstream in the process buider). It should save iterations on blur - right now I see a save button there but in the rest of the application we save on blur so I thiink customers expect that behaviour. 

4. I don't fully understand how the `add group` button will work - you can click on it and then you select phases you want to add to the group? When we talked about a possible solution for the creation of groups we talked about having a small 'link' button between all phases (same size as the downwards chevron between the phases) which then can be used to group two phases together - maybe there was a misunderstanding? It was, of courser, just a suggestion - any solution that feels user friendly will do.

ConorTighe1995:
> Since this is a WIP, ill just comment for now:
> 
> In general I like the way groups looks compared to normal phases. I have a couple questions/requests for the visuals, mainly to keep things behaving the same way as in the rest of the application and keeping the overview:
> 
> 1. I believe the button to add a phase within a group should be the same as at the buttom of the process tree - it does the same thing and I think it contributes to the clarity.
> 2. I think we don't need scrolling within the groups, because it make it difficult to understand the full process sequence at one glance. Just fully exapand the groups - thats totally fine.  The add button should also be just aligned with the normal phases within the group (currently its one layer up probably due to the fact that we have scrolling)
> 3. For the iterations, I really like the popup but I'm afraid it is also required for the customer to immediately see iterations of every group in the process tree, so we shouldn't hide it in a popup. I'm also not clear about what `Phases` mean in the popup - is that the amount of phases in the group? If so, I think this information is not really required because they can see the phases in the group (even better when we get rid of the scrolling) and the phase count itself is of no real value. With the three icons in the group name header it also looks a bit dense - maybe have a new row below the group name with an input field that shows current iterations and allows entering a number (just like we do for the downstream in the process buider). It should save iterations on blur - right now I see a save button there but in the rest of the application we save on blur so I thiink customers expect that behaviour.
> 4. I don't fully understand how the `add group` button will work - you can click on it and then you select phases you want to add to the group? When we talked about a possible solution for the creation of groups we talked about having a small 'link' button between all phases (same size as the downwards chevron between the phases) which then can be used to group two phases together - maybe there was a misunderstanding? It was, of courser, just a suggestion - any solution that feels user friendly will do.

1. I went with this button out as when I tried the tab button I thought it took up a lot of space and we are already tight for UI space on this element so I tried this smaller button to save some room. It didn't look great when I tried to make the phase additions button its smaller and the plus icon kept offsetting, maybe I could make it smaller version of the additions component.

2. That sounds great, but I think if we are doing that we will still need a scrollbar to limit how much the widget could grow because it might become  a bit messy on different screens and devices, we should start moving in a more UI adaptive direction. I think we talked about this expand feature before but I thought it wasn't going to be implemented just yet or in another ticket.

3. Phases was meant to be total phases, as in phases * iterations, but your right I changed the text to get around a formatting issue in the popup but information was lost in the process and its not very clear. Im going back to the drawing board with the popup anyway as the is a bug where the drag and drop with the phase tree seems to interfere with the input bar, but I do intend on adding a way to represent iterations on the widget once I can successfully set the value for it. 

4. The plan I had was to make it multi click, I was going to create a ticket for the functionality of that after the UI is agreed on, tested and merged. Maybe I misunderstood, I thought when you said link, you could click the ones you want to add or click down through the processes, maybe its me misunderstanding the process of experimentation by our users, but in the demo for example they all have specific names and I imagine for the best customer experience you could click the chain/link button, click the 3-4 you want then create a group. from that, this will be added to the bottom of the tree.

